PR #787, opened 2014-12-27
flexible the build scripts and add a freebsd build script

andre-hub:
Please add my change for build scripts (has now a freebsd script)


denji:
@Unknwon 
```sh
if [ "$CC" = "gcc" ]; CGO_ENABLED=1
if [ "$CC" = "clang" ]; CGO_ENABLED=0
```

Issue #763, opened 2014-12-18
Synchronize .ssh/authorized_keys on request 


fzerorubigd:
There is no way to synchronize authorized key in web and the .ssh/authorized_key. 
I want to make sure there is no other key in the authorized_key except the keys from gogs. the problem was I reinstall gogs in my machine and I get this error : 
```
 Gogs: you are not the owner of SSH key
```

Another thing is if this file is out of sync, then the ssh add/delete dose not work correctly. but I'm not sure about this.

Unknwon:
fixed by #818

PR #818, opened 2014-12-31
add a function to rewrite all public keys on admin request

fzerorubigd:
Rewrite all ssh keys again in .ssh/authorized_key. 

ps - sorry for extra change in language file, it was my emacs hook to remove space and empty lines in save.
refs #763


Unknwon:
It's been late, but still, thanks a lot!

PR #821, opened 2015-01-02
Fix #819 by fetching the repository from db before updating

### LINKED ISSUE #819: num_watchers not updated when adding a bare repository
- **Bug Description**: 
If you add a repository (private or public, doesn't matter), the num_watchers in the database is not updated.. So if you unwatch, the number of watchers can be negative..
See for yourself https://try.gogs.io/klausenbusk/test3 and https://try.gogs.io/klausenbusk/test4
- **Gogs Version**: `v0.5.9.1229 Beta`
- **Git Version**: `2.2.1`
- **System Type**: `Arch Linux`
- **Error Log**: Nothing
- **Other information**: Don't seems to happens, if you mark the "Initialize this repository with a README.md"..

phsmit:


Unknwon:
Thanks your PR!

PR #823, opened 2015-01-02
Fix #799 by adding a tooltip for all dates in all settings/panels

### LINKED ISSUE #799: User Manage Panel does not show precise creation date
Shows only the creation date, not the time :-1:

phsmit:


Unknwon:
Thanks!

Issue #798, opened 2014-12-29
Automatically convert SSH2 public keys to OpenSSH format


rojaro:
To use a SSH2 public key with gogs one has to convert it to the OpenSSH format first i.e. through ssh-keygen
```
ssh-keygen -i -f ssh2.pub > openssh.pub
```
It would be useful to automatically convert these keys into the OpenSSH format so a windows user doesnt have to do that.

As the conversion is pretty simple (basically just string manipulation), this could be done without using ssh-keygen within gogs. To test this, you can create a SSH2 public key from your existing OpenSSH key using
```
ssh-keygen -e -f openssh.pub > ssh2.pub
```

Unknwon:
Sounds not bad. But how does Gogs know user is giving SSH2 or SSH public key?

Unknwon:
Please follow on https://github.com/gogits/gogs/issues/1877 now :D

PR #825, opened 2015-01-02
Implement #798 Flexible ssh-key input

phsmit:
It is now possible to input ssh keys in a number of formats: openssh, SSH2 or just the base64 encoded key.


Unknwon:
Grateful!

PR #830, opened 2015-01-05
Add max-width to commit message of repo-files-table

TonyTsangHK:
Prevent repository files table from overflow by a long commit message.

Before
before

After
after


Unknwon:
Thanks!

Issue #842, opened 2015-01-07
Allow http(s) git actions using application token


compressed:
I was wondering if we can use the application token that a user has created to also permit git actions to repos the user has access to.

This is useful when it's needed to programmatically push commits to a repo.

GitHub allows this via the following:

```
git push https://${GITHUB_TOKEN}@github.com/user/repo.git
```

Edit: It would be best if the token works for the gamut of git actions: e.g. pull, push, clone, etc.

Unknwon:
Will Git converts `${GITHUB_TOKEN}` to request header automatically?

compressed:
I don't think so. We probably need to add some logic in gogs that if we don't find the username, check if the username is actually a token.

Some more info here: https://github.com/blog/1270-easier-builds-and-deployments-using-git-over-https-and-oauth

Unknwon:
I want to say that Go looks like cannot get content before `@`, and it doesn't even say user is using HTTP or HTTPS in `*http.Request`.

But, 

> check if the username is actually a token.

I think it would work. :+1:

Unknwon:
Hi @compressed , thanks for waiting. I've pushed some fixes based on changes you made on https://github.com/gogits/gogs/pull/845 . You are also free to try on https://try.gogs.io/

Unknwon:
As new mirror release `v0.5.13` is going to release, and this issue is labeled as **feedback needed**, so close it for **assumed resolved**, feel free to reopen this if you want to continue discussing. :beer:

lavvy:
Please has this feature being added ?  Please how do I learn more how to use it to pull  .tar from my private repository

Unknwon:
@lavvy this was implemented a while ago, based on the code, yes, it has added.

lavvy:
Oh cool. So I am cloning into Jenkins with this url  when it's public repo
https://try.gogs.io/lavvy/newproj

And I download tar file with this url
https://try.gogs.io/lavvy/newproj/archive/master.tar.gz

So if it now private repo, how will the urls be ?

Unknwon:
You use `curl` with basic auth through HTTP.

lavvy:
Yeah I mean assessing it from a url for instance Jenkins will access it as
an application. Then if I want to do wget command from Linux , how will the
url be

lavvy:
And then how to generate the token for a repo. I couldn't just see how

Unknwon:
[...]

PR #845, opened 2015-01-08
allow http push by token - #842

compressed:
Hi,
Here's my attempt to allow pushing by token. Not sure if there were any other places to change.

Let me know how it looks.

Thanks!

Edit - sent to gogs/master instead of gogs/dev by accident. Fixed now.

### DIFF
--- a/models/token.go
+++ b/models/token.go
@@ -62,6 +62,21 @@ func ListAccessTokens(uid int64) ([]*AccessToken, error) {
 	return tokens, nil
 }
 
+// ListAllAccessTokens returns all access tokens
+func ListAllAccessTokens() ([]*AccessToken, error) {
+	tokens := make([]*AccessToken, 0, 5)
+	err := x.Desc("id").Find(&tokens)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, t := range tokens {
+		t.HasUsed = t.Updated.After(t.Created)
+		t.HasRecentActivity = t.Updated.Add(7 * 24 * time.Hour).After(time.Now())
+	}
+	return tokens, nil
+}
+
 // DeleteAccessTokenById deletes access token by given ID.
 func DeleteAccessTokenById(id int64) error {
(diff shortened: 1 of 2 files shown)


Unknwon:
Very late reply, sorry!

But for this feature, if no other security concerns, I think change this single line would solve the problem: https://github.com/gogits/gogs/blob/master/modules/auth/auth.go#L30 

What are your thoughts?

compressed:
No worries!

@Unknwon - that would be great if we can make a single line change. I'm not sure though what that change would be.... Can you help?

It seems when I clone using an API key as the user name, the API key is within the http Basic header, which means the encoding needs to be applied. There won't be anything in the token header. Also, it seems even if we change `auth.go` to return the proper token.Uid from Basic auth, it still ends up going to `http.go` where it again goes through the `askAuth` check which will ultimately not find the `authUsername` (because it is the token not user name). Hopefully I'm missing something here!

compressed:
ping @Unknwon

Unknwon:
@compressed I'm here, many PRs are left in my vocation, I'm working on these in order, so will be soon help on this! 

Sorry about this!

Unknwon:
> It seems when I clone using an API key as the user name, the API key is within the http Basic header, which means the encoding needs to be applied. There won't be anything in the token header. Also, it seems even if we change auth.go to return the proper token.Uid from Basic auth, it still ends up going to http.go where it again goes through the askAuth check which will ultimately not find the authUsername (because it is the token not user name). Hopefully I'm missing something here!

You're absolutely right!

Just a note that list all tokens isn't a good way when token number gets very large. 

I'll merge right now, but will push some fixes soon after.

Thanks for your patient and work!

compressed:
Great, thanks a lot! Yes, I realized iterating through the tokens wasn't a good idea, but I never updated PR. Thanks for doing that!

Unknwon:
@compressed , no problem, my pleasure!

PR #847, opened 2015-01-09
Translate

gittex:
This PR contains some translation updates:
1) it adds missing translation into german (de_DE)
2) it improves two english texts a little bit


Unknwon:
Hmm, I think you realize we do translation work on https://crowdin.com/project/gogs , unless for English version(`locale_en-US.ini`).

PR #848, opened 2015-01-09
Improve look of the user email settings page

gittex:
This PR tries to improve the look and feel of the email settings page.

It aligns the design to the other settings pages, adds highlightning of the current item
and introduces sorting of the email adresses listed. The primary address will always be listed first with the
others following in lexical order.

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -219,12 +219,13 @@ change_password_success = Password is changed successfully. You can now sign in 
 
 emails = E-mail Addresses
 manage_emails = Manage e-mail addresses
-email_desc = Current e-mail addresses
+email_desc = Your primary e-mail address will be used for notifications and other operations.
 primary = Primary
-primary_email = Make primary
+primary_email = Set as primary
 delete_email = Delete
 add_new_email = Add new e-mail address
 add_email = Add e-mail
+add_email_success = Your new E-mail address was successfully added.
 
 manage_ssh_keys = Manage SSH Keys
 add_key = Add Key
--- a/models/user.go
+++ b/models/user.go
@@ -15,6 +15,7 @@ import (
 	"image/jpeg"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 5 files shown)


phsmit:
Looks much better indeed :+1:

phsmit:
About the sort, wouldn't it be better to have the initial sorting done by the database? (adding .Asc() to the query). After that we can indeed move the primary email to the top

gittex:
Doing the sort would at least make the code look a little cleaner.

But since I have no experience with the ORM, yet I did stick to what I know.
The way it is implemented now would allow for easy addition of more sphoisticated sorting orders, but if you insist I will change the patch to do the sort in the DB.

phsmit:
I'm not a Gogs maintainer, so I will not insist anything :)

gittex:
I've refactored my patch a little to touch less existing code.

conf/locale/locale_en-US.ini
```diff
@@ -219,12 +219,13 @@ change_password_success = Password is changed successfully. You can now sign in
```
Unknwon: Please rule out this file since merged in another PR.

models/user.go
```diff
@@ -629,7 +649,7 @@ func GetUserIdsByNames(names []string) []int64 {
 // Get all email addresses
 func GetEmailAddresses(uid int64) ([]*EmailAddress, error) {
 	emails := make([]*EmailAddress, 0, 5)
-	err := x.Where("uid=?", uid).Find(&emails)
+	err := x.Where("uid=?", uid).Asc("Email").Find(&emails)
```
Unknwon: In ORM you need `email` not `Email`.

public/ng/css/gogs.css
```diff
@@ -1813,6 +1813,7 @@ The register and sign-in page style
 #repo-hooks-history-panel,
```
Unknwon: Hmm... you should change `.less` file not the css ones.

Unknwon:
Thanks your PR!

I've made some comments, ping me if you run into any problem!

models/user.go
lunny: The default Mapper rule of xorm is `EmailName <-> email_name`.

Unknwon:
Close due to https://github.com/gogits/gogs/pull/895 , thanks your PR, though!

PR #865, opened 2015-01-17
Allow start.sh from any working directory

clinta:
This change to start.sh allows it to be executed from any working directory and always reference the gogs binary in the parent directory. Very helpful when calling start.sh from an init script.


Unknwon:
Thanks your PR!

Please PR to `dev` branch next time :smile:

PR #866, opened 2015-01-20
Fix in url if cache avatar is enabled

vitalvas:
Fixes a bug with the generation of addresses to download avatar.
For some reason, init() does not work.

    2015/01/19 20:31:30 [D] avatar.fetch(fetch new avatar): http:39b285bb1127ed8bbc927e9690f74e19?d=retro&r=pg&size=200
    2015/01/19 20:31:30 [T] avatar update error: Get http:39b285bb1127ed8bbc927e9690f74e19?d=retro&r=pg&size=200: http: no Host in request URL

### DIFF
--- a/modules/avatar/avatar.go
+++ b/modules/avatar/avatar.go
@@ -38,10 +38,12 @@ import (
 
 var gravatarSource string
 
-func init() {
+func UpdateGravatarSource() {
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

PR #868, opened 2015-01-21
Refactor collaborations into its own table

phsmit:
This PR introduces a new Collaboration struct which is a direct relation between a repository and a non-organization user. Introducing this table will make it much easier to refactor the access table.

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -45,7 +45,7 @@ func init() {
 		new(Issue), new(Comment), new(Attachment), new(IssueUser), new(Label), new(Milestone),
 		new(Mirror), new(Release), new(LoginSource), new(Webhook),
 		new(UpdateTask), new(HookTask), new(Team), new(OrgUser), new(TeamUser),
-		new(Notice), new(EmailAddress))
+		new(Notice), new(EmailAddress), new(Collaboration))
 }
 
 func LoadModelsConfig() {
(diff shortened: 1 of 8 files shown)


models/models.go
```diff
@@ -127,13 +127,65 @@ func SetEngine() (err error) {
...
+	return nil
+}
+
 func NewEngine() (err error) {
 	if err = SetEngine(); err != nil {
 		return err
 	}
+
+	var hasAccess, hasCollaboration, needMigration bool
+	if hasAccess, err = x.IsTableExist("access"); err != nil {
```
Unknwon: I would suppose `access` always exists(since initial Gogs version)?

phsmit: This is actually a disguised test to see if there is any tables at all. For a new installation no migration needs to be run

Unknwon: Right, but I think it costs nothing to `select` from empty `access` table, and we will have more organized code block(put `IsTableExist` and `migrateAccessToCollaboration ` together)?

phsmit: Good point. However, we can't put the two block together, as we want to test for the collaboration table before the Sync, and do the actual filling after the table has been synced.

Unknwon: Ah, yes. I'm stilling thinking on database version, so there is a function go through auto-migrate from the closest large version one by one until the current. It only requires only one person work on database struct change at one time, which I don't think is that hard?

phsmit: Ok. I'll try to make this and make a new PR. Let's see how that looks.

Unknwon: Sure, and it's completely fine to merge like this and let me do the version work.

phsmit:
Superseded by #899

PR #885, opened 2015-01-31
Add a freebsd init script.

clinta:
Adding a FreeBSDinit script will make gogs much more convenient to install for FreeBSD users, and will aid in creating a FreeBSD port for this project.


Unknwon:
Thanks a lot! Appreciate!

Issue #1, opened 2014-02-14
Register/delete users


unknwon:
- [x] Register users
- [x] Delete users

Issue #732, opened 2014-12-11
Reference to commits in issues


Mageti:
Hello !
The more I use Gogs, the more I like it.
I just found a missing feature (as of Version ```0.5.8.1210 Beta```) : seeing a commit in an issue.

Let me explain : I open an issue, that have the number '#num'.
Then I mention the issue number in a git commit message, like "this commit corrects issue #num".
When I go to the "commits" view, I see the commit message and the issue number is a link to the correct issue page :
![issue-reference](https://cloud.githubusercontent.com/assets/4189017/5394844/8f75a7a4-8142-11e4-8992-c4f786449cfa.png)

But on that issue page, there is no reference to the commit.
Like in github :
![commit-reference-in-issue](https://cloud.githubusercontent.com/assets/4189017/5394817/3b77135e-8142-11e4-8580-f4f59dd73640.png)

Subsidiary question : is it possible to close an issue with the presence of certain keywords in the commit message ? like "this commit closes #num", with the word "closes" set in configuration to close an issue if used in commit message followed by #num

Thanks in advance.

Unknwon:
It is somehow implemented a while ago...

Issue #668, opened 2014-11-26
 Could you set up links to closed tasks in commit ?


adwmicro:
Could you set up links to closed tasks in commit ?

Unknwon:
Hi @adwmicro , this should work right now on `dev` branch, you can also test on https://try.gogs.io/

Unknwon:
As new mirror release `v0.5.13` is going to release, and this issue is labeled as **feedback needed**, so close it for **assumed resolved**, feel free to reopen this if you want to continue discussing. :beer:

PR #892, opened 2015-02-02
Referencing issues from commit messages is now possible.

### LINKED ISSUE #462: Controlling issues via commit commands
Like on GitHub and other services, ex. closes #x or fixes #x or reopens #x

EtienneBruines:
Code-wise: the fixing-part of the code now doesn't reference the issue any more, because this is done by the referencing-part of the code, which is called just before.

Fixes #462, #668, #732.

### DIFF
--- a/models/action.go
+++ b/models/action.go
@@ -41,12 +41,14 @@ var (
 
 var (
 	// Same as Github. See https://help.github.com/articles/closing-issues-via-commit-messages
-	IssueKeywords    = []string{"close", "closes", "closed", "fix", "fixes", "fixed", "resolve", "resolves", "resolved"}
-	IssueKeywordsPat *regexp.Regexp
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR!

Simple question:

Would this will produce one reference and one close actions if commit message is `fix #1`?

Because based on the patterns, it seems like reference pattern will also match strings that should be only matched for `fix/fixes` etc. patterns?

EtienneBruines:
It references and closes, yes. This was already the case, and, since a "close" does not reference to an issue, this is what you'd want.

Unknwon:
Cool, I see. Will merge soon in terms of other matters.

BTW, how's your dinner going? :smile:

Unknwon:
Merged, thanks again!

PR #897, opened 2015-02-03
fix error at admin/config

makhov:
Fix error at admin/config (wrong var name in template)
```
template: admin/config:64:48: executing "admin/config" at <.DbCfg.SslMode>: SslMode is not a field of struct type interface {}
```


Unknwon:
Thanks a lot!

PR #899, opened 2015-02-04
New Collaboration pr

phsmit:
This PR introduces a new Collaboration struct which is a direct relation between a repository and a non-organization user. Introducing this table will make it much easier to refactor the access table.


phsmit:
Oops, to master. #900 supersedes.

PR #900, opened 2015-02-04
Collaboration

phsmit:
This PR introduces a new Collaboration struct which is a direct relation between a repository and a non-organization user. Introducing this table will make it much easier to refactor the access table


Unknwon:
Thanks your PR! Have classes today, will review soon.

Unknwon:
I would like to review later PRs since this one introduce database migration(if I remember right), hope you don't mind :satisfied:

phsmit:
No problem. One thing is that the next one (access refactoring) builds on this pr, so if big changes are needed that would be good to be done earlier

Unknwon:
Thanks for the tip, I'll keep in mind!

phsmit:
I made #925 now to the access branch

PR #901, opened 2015-02-04
fix lines highlighting at diff view

makhov:
Now it works:
init commit 5743f50e2c gogs go git service

### DIFF
--- a/public/ng/js/gogs.js
+++ b/public/ng/js/gogs.js
@@ -141,6 +141,7 @@ var Gogs = {};
     Gogs.renderCodeView = function () {
         function selectRange($list, $select, $from) {
             $list.removeClass('active');
+            $list.parents('tr').find('td').removeClass('selected-line');
             if ($from) {
                 var a = parseInt($select.attr('rel').substr(1));
                 var b = parseInt($from.attr('rel').substr(1));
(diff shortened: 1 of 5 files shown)


Unknwon:
Very awesome, thanks!

Is it highlighting multiple selected lines?

makhov:
Yes, it is.

Unknwon:
Merged, thanks again!

Unknwon:
It seems broken normal code view selection...

Unknwon:
Thanks your PR, and I've made some fixes(ugly but at least work right now) to `dev` branch and continue looking for your help!

The code you want to change is: https://github.com/gogits/gogs/blob/dev/public/ng/js/gogs.js#L206-L275

PS: Single line of highlighting in diff page is working, but now uses hashtag `#diff-{$i}L{$j}{$k}`.

Here is one issue we're facing:

I'm not really good at JS but I'm trying. We want to have multiple lines highlighting to be in the format of `#diff-{$i}L{$j1}{$k1}-L{$j2}{$k2}`, not sure if it's possible.

We can open a separate issue for this matter.

makhov:
Ok, I'll try to fix multiple lines highlighting today.

PR #905, opened 2015-02-05
Add option to provide configuration file on command line

phsmit:
Add a --config flag to all commands that specifies the location of the custom config file. If flag is not used, the current custom/conf/app.ini is used.


phsmit:
The failed build of travis is due to a problem on their side, not with the code. (It only fails on the tip version of go, with "go not found")

Unknwon:
> The failed build of travis is due to a problem on their side, not with the code. (It only fails on the tip version of go, with "go not found")

I know they have strange issues with Go, but I like their badge :smile:

Unknwon:
Thanks your PR, there are two things I want to say after reviewed:

1. You can use global flag https://github.com/gogits/gogs/blob/dev/gogs.go#L40 so you only need to write once. 

then the calling format will be like `./gogs -config xxx web/dump`.

2. https://github.com/gogits/gogs/pull/905/files#diff-38994e9fb06f88004313ddbefc7b2fadR32

For this line, I've thought another way to do it: 

the only reason we haven't had CMD option for custom path is because `gogs serv` isn't called by Gogs itself, so what if we let Gogs write the custom path variable into file `/path/to/gogs/custom/path.env`(name is changeable) because `gogs serv` knows where the binary is, and can find this file easily. This way, we don't need to bother SSH key file anymore.

phsmit:
About the first, that is a good one, I'll make it global.

As for the second, I think the whole advantage of having a custom conf on the command line is to separate the application from the configuration. At the moment you write something in a fixed path, seen from the binary, you will not be able to run multiple instances.

I'll continue the continue the discussion on gitter

phsmit:
Actually, my first test seems to indicate that the global flags are not honored for the commands. They are not showing in the help, nor do they seem to have effect (when put before the command). I'll investigate later more

Unknwon:
Here is an example: https://github.com/gpmgo/gopm/blob/master/lib%2Flib.go#L57

phsmit:
I have a few issues with this solution. First, global options are not shown on the individual commands help page (if you type gopm list --help you won't see the global options like --noterm). Second, you have to provide global options before the command, 'gopm list -n' is not accepted. All in all I think a command specific flag is maybe nicer?

Unknwon:
You're right... 

Let me review other parts.

Unknwon:
Just hit the rebuild on travis CI, I saw they upgrade Go build env yesterday to fix tip version.

Unknwon:
Merged, thanks again!

I'll push corresponding fixes to this parch ASAP.

Unknwon:
Hi @phsmit , I've pushed some code(https://github.com/gogits/gogs/commit/f15fa9167ade8262bed437451585d319f12d1182) to `dev` branch that allow admin to rewrite all update hook file of all repositories, which is related to your change.

Issue #884, opened 2015-01-31
Add option for hiding "Sign Up" when register is disabled


NewbieOrange:

Unknwon:
Hi, thanks your feedback!

But don't two buttons are showing at the same time? If changed "sign up" to "sign in", will we have two sign in buttons?

NewbieOrange:
Well, if there will be two buttons, remove one. Like there will be only "sign in" in the top right corner.

Unknwon:
It's easy to remove one, but would that confuse users without a notice page like current approve?

NewbieOrange:
No I think, maybe add a option?

Unknwon:
Cool, no problem.

wanghm:
We need this feature, too.

JoeHazzers:
I thought it seemed strange to have the "Sign Up" button around when registration is disabled, too

+1

Unknwon:
Hi all, I've pushed a fix related to https://github.com/gogits/gogs/pull/908 to `dev` branch, to hide `Sign Up` button by default when registration is disable.

wanghm:
Is it possible to hide "Need an account? Sign up now."?

Unknwon:
@wanghm good catch! Fixed pushed to `dev`.

Unknwon:
As new mirror release `v0.5.13` is going to release, and this issue is labeled as **feedback needed**, so close it for **assumed resolved**, feel free to reopen this if you want to continue discussing. :beer:

NewbieOrange:
Everything works fine, resolved. :smile:

Unknwon:
@NewbieOrange awesome! :beer:

PR #908, opened 2015-02-05
Make Signup button display configurable - #884

nicolai86:
See #884 for the original request. I'd love some feedback for the tiny CR since it's my first.


nicolai86:
I left out all the locales because I have to look them up yet. Will add those soon…

nicolai86:
We now have the translations as well, except for lv and zh-HK

Unknwon:
Thanks your PR!

But we don't do translation in here: see readme, except for English.

nicolai86:
@Unknwon should I leave the new key for every language, or just revert the last commit except for the english translation?

Unknwon:
Actually it doesn't mattar, non-English files will be eventually overwritten. 

Your PR looks good otherwise, will merge soon. :)

nicolai86:
Cool. I'll leave the PR as is then :)

Unknwon:
Merged, thanks again!

Issue #898, opened 2015-02-03
LDAP - User changes password - unable to push/pull over HTTP with new password


theangryangel:
I currently have a small gogs install running with LDAP authentication against Active Directory.

When a user changes their password on the LDAP server they're able to login and out of the web interface with the new password (as expected).

However, any git push/pull, etc. fails saying "wrong password". Use their old LDAP password and it works fine. If we login to their settings page in gogs, and change password (typing in their old LDAP password and new LDAP password for the respective fields) it then allows them to push, pull, etc.

I've had a quick hunt through the gogs source to try and patch this myself, but I don't know the code base or go well enough to find and provide a patch :( (yet). Any pointers in the right direction and I'm willing to take a second look at it, and see if I can sort out a fix.

For what it's worth this is a gogs 0.5.9.1216 (I've not yet had time to compile 0.5.11) on Windows.

Unknwon:
Thanks your report! 

I think there is a related issue that LDAP users do not have a password in gogs database(though HTTP/HTTPS should check with LDAP again, which I consider this is a bug right now), the only way work for now probably would be go to admin panel and reset corresponding user password and try again.


Hope helps!

theangryangel:
I can confirm that the suggested fix does work.

I'll go digging through the code and see if I can come up with a patch :)

Unknwon:
> I'll go digging through the code and see if I can come up with a patch :)

What will be cool even I can imagine that's hard part of the entire project :smile:

Ma-XX-oN:
Was this fixed.  I'm having the same issue under git version `2.21.0.windows.1`

Ma-XX-oN:
Hmmm, had to open a new git-bash session and then do a pull.  I then got a windows dialog asking for the password.

PR #909, opened 2015-02-05
Change SignedInUser to use the models/login.go UserSignIn

theangryangel:
This is a patch for #898 

I hope this is acceptable and works as expected. It should permit users to pull and push using non-local accounts over HTTP, without having to change the password in the local database.


Unknwon:
Thanks your PR!

But please re-PR to `dev` branch :)

PR #910, opened 2015-02-06
Change basic authentication to use the models/login.go UserSignIn, to pe...

theangryangel:
A new PR, pointing at dev branch - sorry I hadn't noticed there was a dev branch :x My bad!

This is a patch for #898


theangryangel:
I've realised this doesn't actually solve the problem. Please ignore. Sorry for the noise. I'll come back when I've fixed the problem properly.

Unknwon:
@theangryangel don't be upset!

PR #911, opened 2015-02-06
Link to previous commited source file (diff.view_file button) instead of returning 404 for deleted files.

TonyTsangHK:
Same behavior as github.com


Unknwon:
Thanks! Will get on this ASAP.

Unknwon:
thanks a lot!

PR #912, opened 2015-02-06
Access Refactoring

phsmit:
This PR builds on #900 . It is not a complete product yet, but it shows what kind of idea I have with the access refactoring.

In this first commit it already removes a great number lines of code and further refactorings might get it down even more. 

Things that still need to be done:
 - Migration needs to be written
 - Testing all changes to see if nothing breaks.


Unknwon:
Since changes are a lot, would you like we create a `access_refactor` branch on main repository to test it?

phsmit:
If a branch works good for you, it is good for me. I can PR both collaboration and access_refactor to it

models/access.go
```diff
@@ -4,77 +4,151 @@
...
+type AccessMode int
 
 const (
-	READABLE AccessType = iota + 1
-	WRITABLE
+	NoAccess AccessMode = iota
+	ReadAccess
+	WriteAccess
+	AdminAccess
+	OwnerAccess
```
Unknwon: I would prefer

```go
ACCESS_MODE_NONE
ACCESS_MODE_READ
ACCESS_MODE_WRITE
ACCESS_MODE_ADMIN
ACCESS_MODE_OWNER
```

phsmit:
Superseded by #926

PR #915, opened 2015-02-07
Get username, name, surname and e-mail from LDAP server

Lafriks:
Add options for getting username, name, surname and e-mail from LDAP server

Don't know if this is the best way but it was fastest way to make this fix and seems to be working

Example values:
username: cn
name: givenName
surname: sn
e-mail: mail


Unknwon:
Thanks your PR! Will get on this ASAP.

Unknwon:
Thanks a lot! Sorry for late.

PR #917, opened 2015-02-08
Updated gogs.css - monospace font for hash

Stefan-Code:
Updated gogs.css to set the font-family of .sha .label in the #repo-files-table to a monospace font to make it look nicer so the label boxes have the same width and don't differ by a few pixels.


Stefan-Code:
oooops... didn't notice css files were less generated. Closing...

PR #919, opened 2015-02-08
Updated gogs.css - monospace font for hash

Stefan-Code:
Updated gogs.css to use monospaced font for sha hash label in repository view to make it look nicer because of equal label width.
Before:
gogs3
After:
gogs4


Unknwon:
Thanks, appreciate!

PR #923, opened 2015-02-09
Fix #922

### LINKED ISSUE #922: "Gogs: auth file format error" when cloning via ssh
Hi.  Fresh install from source on a Debian 7 (32 bit) with go version go1.4.1 linux/386 here.

I added an SSH key but kept getting "Gogs: auth file format error" whenever I tried to clone a repo via ssh.  In the serv.log runServ() was complaining "[F] Invalid auth file format".  Looking serve.go in runServ() that error comes up if it can't split os.Args[2] on a hypen.

I manually edited my .ssh/authorized_keys and moved the "key-1" to before the "--config" parameter and it worked.  (So now it says command="/home/git/go/src/github.com/gogits/gogs/gogs serv key-1 --config=.....").  Looks like the parameter order that is created when an ssh key is uploaded is different to the one expected?

Cheers,
Darren

phsmit:


phsmit:
I quickly fixed the templates first to alleviate the symptom. The second commit by itself would also fix this issue, but I think the new place of --config in the templates is nice anyway.

Unknwon:
thanks!

PR #924, opened 2015-02-09
Fixes #921

### LINKED ISSUE #921: @ mention generates wrong link
See https://try.gogs.io/psmit/test1/issues/1

The @-mention is processed correctly but the resulting link misses the first character.

phsmit:
Fixes #921 and makes the Mention regexp be in line with the others


Unknwon:
thanks!

PR #925, opened 2015-02-09
Collaboration

phsmit:
This PR introduces a new Collaboration struct which is a direct relation between a repository and a non-organization user. Introducing this table will make it much easier to refactor the access table


models/models.go
```diff
@@ -100,6 +101,7 @@ func NewTestEngine(x *xorm.Engine) (err error) {
 		return fmt.Errorf("connect to database: %v", err)
 	}
 
+	x.SetMapper(core.GonicMapper{})
```
Unknwon: Wouldn't this be a problem when we haven't changed all `Id`s to `ID`?

phsmit: Nope, not at all. It only maps names with two successive Uppercase characters different and there are no structs or fields in models/ that have that.

PR #926, opened 2015-02-09
Access Refactoring

phsmit:
This PR builds on #925 . It is not a complete product yet, but it shows what kind of idea I have with the access refactoring.

In this first commit it already removes a great number lines of code and further refactorings might get it down even more.

Things that still need to be done:

Migration needs to be written
Testing all changes to see if nothing breaks.


phsmit:
Superseded by #941

PR #927, opened 2015-02-09
Remove GOGS_CUSTOM and introduce config variable for CUSTOM_PATH

phsmit:
GOGS_CUSTOM environment variable will not be used anymore. robots.txt is now searched inside the custom/conf directory

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -126,7 +126,7 @@ func newMacaron() *macaron.Macaron {
 	m.Use(i18n.I18n(i18n.Options{
 		SubURL:          setting.AppSubUrl,
 		Directory:       path.Join(setting.ConfRootPath, "locale"),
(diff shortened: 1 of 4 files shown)


Unknwon:
How would it possible to read `CUSTOM_PATH` when we don't even know where is the custom path?

PS: default config file isn't allowed to change.

phsmit:
I'm not sure I understand. The CUSTOM_PATH will be defined in the config, and will be used for locating four things: robots.txt, custom gitignores, custom licenses and custom locales (and anything else that we could think off)

The default value will be "the same directory as the custom configuration file", meaning that everything stays the same, except for robots.txt which was originally in one level higher.

What do you mean with "default config file isn't allowed to change"? Isn't it the idea if there is a new option that it gets added to the default config?

PR #937, opened 2015-02-11
Changes to repo-header-download-drop

Stefan-Code:
Changed the repo-header-download-drop url input to behave slightly different. When clicking on the url input it now selects the whole url and you can copy it manually too. (When using a Flash Blocker for example.). Also I have replaced the Red Circle cursor (due to the default disabled class) with the default one. The behavior is now the same as on github.com.
Before:
gogs7
After:
gogs6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 *.db
 *.log
+log/
 custom/
 data/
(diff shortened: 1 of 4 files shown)


Unknwon:
Looks good! Thanks!

PR #941, opened 2015-02-12
Access refactor

phsmit:


phsmit:
Updated version of #926

Unknwon:
Thanks your PR!

I can see some small issues with SQLite3 but I'll do code fix after merge.

PR #943, opened 2015-02-12
Use client certificate for SMTP over TLS

diseaz:
Some SMTP servers may be configured to check client certificate and do not accept arbitrary self-signed ones.

This pull request adds new keys to app.ini to specify cert and key files in the same way as for HTTPS certificate. Key names are the same as for HTTPS (CERT_FILE, KEY_FILE), so in the case there is a single certificate for both HTTPS and mail, it can be specified once in the default section of the app.ini.

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -105,6 +105,10 @@ SUBJECT = %(APP_NAME)s
 HOST =
 ; Do not verify the certificate of the server. Only use this for self-signed certificates
 SKIP_VERIFY = 
+; Use client certificate
+; USE_CERTIFICATE = true
+; CERT_FILE = custom/mailer/cert.pem
+; KEY_FILE = custom/mailer/key.pem
 ; Mail from address, RFC 5322. This can be just an email address, or the "Name" <email@example.com> format 
 FROM =
 ; Mailer user name and password
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -106,9 +106,9 @@ HOST =
 ; Do not verify the certificate of the server. Only use this for self-signed certificates
 SKIP_VERIFY = 
 ; Use client certificate
-; USE_CERTIFICATE = true
-; CERT_FILE = custom/mailer/cert.pem
-; KEY_FILE = custom/mailer/key.pem
+USE_CERTIFICATE = false
+CERT_FILE = custom/mailer/cert.pem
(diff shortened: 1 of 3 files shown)


modules/mailer/mailer.go
```diff
@@ -72,9 +72,15 @@ func sendMail(settings *setting.Mailer, recipients []string, msgContent []byte)
 		return err
 	}
 
+	cert, err := tls.LoadX509KeyPair(settings.CertFile, settings.KeyFile)
```
Unknwon: Should we add a line before this to check if user is willing to use cert and key?

The code right now looks like it forces user to have cert and key files.

modules/setting/setting.go
```diff
@@ -446,11 +446,12 @@ func newSessionService() {
...
-	Host         string
-	From         string
-	User, Passwd string
-	SkipVerify   bool
+	Name              string
+	Host              string
+	From              string
+	User, Passwd      string
+	SkipVerify        bool
+	CertFile, KeyFile string
```
Unknwon: Corresponding to above comment, here we may also want a `UseCertificates` field.

Unknwon:
Thanks your PR!

I've made some comments to help you work better, ping me if you have any other concerns!

modules/mailer/mailer.go
diseaz: Ah, thanks. Missed this.

conf/app.ini
```diff
@@ -105,6 +105,10 @@ SUBJECT = %(APP_NAME)s
 HOST =
 ; Do not verify the certificate of the server. Only use this for self-signed certificates
 SKIP_VERIFY = 
+; Use client certificate
+; USE_CERTIFICATE = true
+; CERT_FILE = custom/mailer/cert.pem
+; KEY_FILE = custom/mailer/key.pem
```
phsmit: There are no other options in this file that are commented out. Maybe uncomment them and give them good defaults?

Also, wouldn't it make sense to interpret an empty cert_file / key_file as not using certificates and when a path is provided using certificates?

E.g. 

    ; Provide paths below to use client certificate for smtp connection. 
    CERT_FILE = 
    KEY_FILE =

Unknwon:
Merging, thanks again!

PR #949, opened 2015-02-13
Rewrite of access migration

phsmit:
The old migration had a few issues:
 - It left old column names around
 - It did not give the right access levels for owners and admins

Also, this includes a migration that fixes the authorization of owner teams, which was previously ORG_ADMIN (instead of ORG_OWNER)


Unknwon:
You don't need another function, only we two are on the `access` branch.

Unknwon:
I most likely will not merge this one but would modify existence code based on your contribution.

Unknwon:
Merged, thanks again!

Issue #948, opened 2015-02-13
A collaborator on a mirror repository can push changes


phsmit:
A collaborator on a mirror repository can push changes. Mirror repositories should have no write actions.

I also think issues should not be enabled for mirror repositories?

An owner can't push to a mirror repository, but get's an authentication failed error instead of something informative.

Unknwon:
Fixed by https://github.com/gogits/gogs/pull/958

PR #958, opened 2015-02-16
Make sure that a mirror can't be written to by http or ssh

phsmit:


phsmit:
Fixes also #948

Unknwon:
Thanks a lot!

PR #959, opened 2015-02-16
Updating context and fixing permission issues

phsmit:
The boolean flags in the repo context have been replaced with mode and two methods

Also, the permissions have been brought more in line with https://help.github.com/articles/permission-levels-for-an-organization-repository/ , Admin Team members are able to change settings of their repositories.


Unknwon:
Looks good, will test after merge.

Unknwon:
Thanks a lot!

PR #960, opened 2015-02-16
Fix that owners also see actions on their repositories

phsmit:
This is a balance between speed and nice code, where speed has won. To prevent a repository query for each action the ownername is match with the current user.

It would be "cleaner" or "better" if we fetch the repository each time. Another option is to add the RepoOwnerID to action

### DIFF
--- a/routers/user/home.go
+++ b/routers/user/home.go
@@ -103,7 +103,12 @@ func Dashboard(ctx *middleware.Context) {
 	feeds := make([]*models.Action, 0, len(actions))
 	for _, act := range actions {
 		if act.IsPrivate {
-			if has, _ := models.HasAccess(ctx.User, &models.Repository{Id: act.RepoId, IsPrivate: true}, models.ACCESS_MODE_READ); !has {
+			repo := &models.Repository{Id: act.RepoId, IsPrivate: true}
+			// This prevents having to retrieve the repository for each action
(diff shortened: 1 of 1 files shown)


routers/user/home.go
```diff
@@ -103,7 +103,12 @@ func Dashboard(ctx *middleware.Context) {
 	feeds := make([]*models.Action, 0, len(actions))
 	for _, act := range actions {
 		if act.IsPrivate {
-			if has, _ := models.HasAccess(ctx.User, &models.Repository{Id: act.RepoId, IsPrivate: true}, models.ACCESS_MODE_READ); !has {
+			repo := &models.Repository{Id: act.RepoId, IsPrivate: true}
+			// This prevents having to retrieve the repository for each action
+			if act.RepoUserName == ctx.User.LowerName {
+				repo.OwnerId = ctx.User.Id
+			}
+			if has, _ := models.HasAccess(ctx.User, repo, models.ACCESS_MODE_READ); !has {
```
Unknwon: Why when current user is owner but still check access?

Should we use `else if` here?

phsmit: Good catch, I'll make it a nested if (because there will be no body for the first if)

phsmit:
I pushed an update

Unknwon:
Thanks a lot!

PR #961, opened 2015-02-16
Remove collaborators when removing and transferring repository

phsmit:


models/repo.go
```diff
@@ -670,6 +670,19 @@ func TransferOwnership(u *User, newOwner string, repo *Repository) error {
 		return err
 	}
 
+	// Remove redundant collaborators
+	collaborators, err := repo.GetCollaborators()
+	if err != nil {
+		return err
+	}
+	for _, c := range collaborators {
+		if c.Id == newUser.Id || newUser.IsOrgMember(c.Id) {
```
Unknwon: Why not check if new user is an organization here?

phsmit: To keep the code simple. In the organization case we have to remove members of the new organization, in the regular user case that user. There is a couple of IsOrgMember queries to many here for the user case, but they don't cause errors and performance is not an issue for an infrequent task like this.

Unknwon:
Thanks again!

PR #962, opened 2015-02-16
Remove the "PHP" style formatting function

phsmit:
The "PHP" formatting function doesn't add anything, except an undocumented date format.

All usages in the templates have been replaced with DateFmtShort and DateFmtLong for convenience.


Unknwon:
Thanks your PR!

PR #963, opened 2015-02-17
Remove GoGet option from repository and handle it with ?go-get=1 instead

phsmit:
The normal go get protocol is to show the go-import meta tag when ?go-get=1 is appended to the url. This commit implements that behaviour and cleans the go-get option from the repository settings page.


Unknwon:
Thanks your PR!

PR #964, opened 2015-02-17
Rewrite/simplify gogs serve

phsmit:
Serve will not leak anymore the existence of repositories when the user does not have access to them. It will however give better feedback in other cases.


Unknwon:
Hi @phsmit , would you like to resolve the conflict?

phsmit:
Conflict resolved

Unknwon:
Some critical issues, but I'll change it after PR, thanks again!

Unknwon:
> Some critical issues

False alarm.

PR #968, opened 2015-02-18
Changed repo-clone url behavior on bare repo page, added LESS compile script

Stefan-Code:
Pretty much the same as  #937 but this time also for the bare repo page. Also added a Script to conveniently compile the LESS files with a single command.


Unknwon:
Thanks a lot!

PR #970, opened 2015-02-18
The new new highlighting

makhov:
Discussed with @phsmit about lines naming style and decide make it github-like, because there was a problem with 1 section 11 line in file and 11 section 1 line.
Now it looks like this: #diff-5922e0b56e4b93646193f3eac37b9202e646946dR1
Hash of file path, left or right and real line num in file.

### DIFF
--- a/templates/repo/diff.tmpl
+++ b/templates/repo/diff.tmpl
@@ -106,13 +106,13 @@
                 <table>
                     <tbody>
                         {{range $j, $section := $file.Sections}}
-                        {{range $k, $line := $section.Lines}}                        
+                        {{range $k, $line := $section.Lines}}
                         <tr class="{{DiffLineTypeToStr .Type}}-code nl-{{$i}} ol-{{$i}}">
                             <td class="lines-num lines-num-old">
-                                <span rel="diff-{{Add $i 1}}L{{$j}}{{$k}}">{{if $line.LeftIdx}}{{$line.LeftIdx}}{{end}}</span>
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks a lot!

templates/repo/diff.tmpl
```diff
@@ -105,14 +105,14 @@
...
                     <tbody>
-                        {{range $j, $section := $file.Sections}}
-                        {{range $k, $line := $section.Lines}}                        
-                        <tr class="{{DiffLineTypeToStr .Type}}-code nl-{{$i}} ol-{{$i}}">
+                        {{range .Sections}}
+                        {{range $k, $line := .Lines}}
+                        <tr class="{{DiffLineTypeToStr .Type}}-code nl-{{$k}} ol-{{$k}}">
                             <td class="lines-num lines-num-old">
-                                <span rel="diff-{{Add $i 1}}L{{$j}}{{$k}}">{{if $line.LeftIdx}}{{$line.LeftIdx}}{{end}}</span>
+                                <span rel="{{if $line.LeftIdx}}diff-L{{Sha1 $file.Name}}{{$line.LeftIdx}}{{end}}">{{if $line.LeftIdx}}{{$line.LeftIdx}}{{end}}</span>
```
phsmit: I think this L is on the wrong place, shouldn't it be after the filename just like the R?

makhov: You're right! Fix it at #972

makhov:
Hm.. I try to represent and fix the second bug.. Worked fine yesterday

phsmit:
I just tried your new PR and it fixes both bugs!

makhov:
Good! But I find another one :) 
Select a block of lines doesn't work when I select two lines with the same line numbers and different L/R prefix. I'll fix it in minutes.

PR #972, opened 2015-02-19
fix typo mistake in highlighting

makhov:
Sorry, my fault. Fix a little mistake in diff.tmpl

### DIFF
--- a/public/ng/js/gogs.js
+++ b/public/ng/js/gogs.js
@@ -209,27 +209,28 @@ var Gogs = {};
             $list.parents('tr').removeClass('end-selected-line');
             $list.parents('tr').find('td').removeClass('selected-line');
             if ($from) {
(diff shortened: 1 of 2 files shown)


phsmit:
:+1: Tested, looks good to me.

makhov:
Fix last one bug

phsmit:
Looks good! I'm not sure if it always makes sense to allow line blocks between different sides, especially not if we later get a "split-view" diff mode.

Anyway a small little bug still. If L2 is lower then R2 and I click on L2 first and click then on R2 this doesn't work. If I click on R2 first and then on L2 it does.

makhov:
I can't represent it. It works fine to me.

phsmit:
I sometimes hate browser caches. I hit Shift-F5 and now everything works perfectly! Sorry for the bother.

Unknwon:
Thanks your PR!

PR #973, opened 2015-02-19
Fix: 504 5.5.2 <localhost>: Helo command rejected

uchti:
Fix: 504 5.5.2 <localhost>: Helo command rejected

### DIFF
--- a/modules/mailer/mailer.go
+++ b/modules/mailer/mailer.go
@@ -100,7 +100,10 @@ func sendMail(settings *setting.Mailer, recipients []string, msgContent []byte) 
 	if err != nil {
 		return err
 	}
-	client.Hello(hostname)
+
+	if err = client.Hello(hostname); err != nil {
+		return err
+	}
 
 	// If not using SMTPS, alway use STARTTLS if available
 	hasStartTLS, _ := client.Extension("STARTTLS")
--- a/modules/mailer/mailer.go
+++ b/modules/mailer/mailer.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/mail"
 	"net/smtp"
+	"os"
 	"strings"
 
 	"github.com/gogits/gogs/modules/log"
@@ -94,6 +95,12 @@ func sendMail(settings *setting.Mailer, recipients []string, msgContent []byte) 
 	if err != nil {
 		return err
 	}
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	client.Hello(hostname)
 
 	// If not using SMTPS, alway use STARTTLS if available
 	hasStartTLS, _ := client.Extension("STARTTLS")


phsmit:
It is a very common practice to send "localhost" to a relay-smtp server. Actually, I have never seen an smtp server that is configured to reject "localhost" in this case. 

The code looks fine, but I'm wondering what will happen if e.g. the hostname of the machine is not in dns, will a smtp server than reject the ehlo?

Could you maybe give some specific details of the smtp server you got this error on?

phsmit:
Would it maybe be better to introduce a config variable "HELO_HOST" that defaults to localhost?

uchti:
Do not worry, in all OS hostname always has a value  = localhost. 
The smtp server not reject the ehlo , if hostname of the machine is not in dns.

phsmit:
But why then does it reject your message when localhost is used?

uchti:
problem in this popular option in postfix config 
http://www.postfix.org/postconf.5.html#reject_non_fqdn_helo_hostname
If this option is set on smtp server , localhost is not valid.

uchti:
more information about this problem in http://www.linuxmagic.com/best_practices/valid_helo_domain.html

phsmit:
But on the same page is the reject_unknown_helo_hostname option, which checks if the hostname is in A or MX.

My question stands, are there relaying smtp servers that actually have this option set?

phsmit:
I'll read the whole document, but this jumps out: "Note! This only applies to MTA to MTA traffic". Gogs is not an MTA, but is much more like a normal user sending email.

phsmit:
I read it completely and actually tells that this is not necessary. Consider that Gogs is your email client. The page you gave reads: http://www.linuxmagic.com/best_practices/valid_helo_domain.html#user

> When you are allowed by your ISP or email provider to use their mail server to send email (eg your 'Outbound Mail Server') they will not check this rule.

phsmit:
Maybe we can just do c.hello("localhost.localdomain"). Do you think this will be rejected by any outgoing smtp server?

uchti:
You can see headers messages from github. It's look something like that

Received: from github-smtp2b-ext-cp1-prd.iad.github.net (github-smtp2-ext8.iad.github.net [192.30.252.199])

From github-smtp2b-ext-cp1-prd.iad.github.net  is not from localhost. You can protest, but doing ehlo with machine hostname before sending is correct.

phsmit:
Sorry, but those headers don't say a thing. github-smtp2b-ext-cp1-prd.iad.github.net is an MTA, but it does not mean that is the host where the message was originally created. 

Let's shorten the hostname to smtp.github.net and say that you have gmail. Now:
    
    dig MX gmail-smtp-in.l.google.com.
    ;; ANSWER SECTION:
    gmail.com.		1872	IN	MX	5 gmail-smtp-in.l.google.com.

So it probably went like this:
Github server (which one, no one knows) -> smtp.github.net -> gmail-smtp-in.l.google.com.

Note that there are 2 arrows here. The first arrow could be any protocol, but let's assume it is smtp. In this case smtp.github.net is the outgoing smtp server (aka smtp relay) and gmail-smtp-in.l.google.com. is the incoming smtp server. The header you showed is about the second connection, from smtp.github.net to the incoming mail server. This is the MTA to MTA connection that your document was talking about. And yes, the ehlo hostname is very important there.

In the case of gogs though, we are talking about the first connection, which is from a smtp client to a mta. In this connection the ehlo hostname is actually not so important and I believe in the real world there are very little outgoing smtp servers that do require the hostname to be correct. That is exactly the reason why go just uses localhost for its connections: its not important in this case. Almost any smtp mailer library you will find in the wild has this behavior. The real world is no RFC :(

phsmit:
More and more I get convinced that your code will not break anything, but doesn't fix anything either, except for maybe better complying to an RFC that no one complies to.

Again, do you have an example of a real world outgoing smtp server that rejects localhost as domain name?

phsmit:
Actually, os.Hostname() can return an error so in that case it does break things...

phsmit:
Please, don't feel attacked by my comments. I do have some experience with email, and I just think that these lines of code don't add anything. The decision to merge this will anyway be with the project owners (I am just a mere contributor)

uchti:
follow RFC is a good motivation to accept this PR :)
My SMTP server is configured with reject_non_fqdn_helo_hostname option.

phsmit:
If it is your own you can easily change it :)

How about if you change this pr like this, to prevent unnecessary errors if Hostname() throw an error:

    if hostname, err := os.Hostname(); err != nil {
        c.Hello(hostname)
    }

Or, if you very much like to send always an FQDN

    if hostname, err := os.Hostname(); err != nil && strings.Contains(hostname, ".") {
        c.Hello(hostname)
    } else {
        c.Hello("localhost.localdomain")
    }

uchti:
If the server for some reason does not return the hostname it is a mistake that could affect the correct execution of the other elements of the program and it should be displayed in the log, do not change the logic of the application in this case.

I think that in my PR do not need to edit.

phsmit:
Gogs never uses the hostname in any other parts. For the operating system it would be good to know, but I don't think Gogs need to fail on that, especially since there is a good alternative. For RFC compliance we should actually use [IP.addr] in case the hostname can not be determined.

uchti:
in go source 
// Hostname returns the host name reported by the kernel.
func Hostname() (name string, err error) {
      return hostname()
}
In which case you will not get an error? Any os can not function properly if it is not returned hostname

uchti:
And yes, hostname can be ip address, but if the function returns err then we have a problem with the os and it would be better to see the information in the log.

Unknwon:
Thanks for the discussion!

Does `client.Hello` returns anything?

uchti:
Yes, I'm sorry about that :) See my next commit. Thanks

Unknwon:
Looks good, merging!

PR #975, opened 2015-02-20
Bugfix for wrong label count

### LINKED ISSUE #933: Issue labels: amount not updated on issue removal
Whenever a label is removed from an issue, the global amount (of that specific label) is unaffected. 

Image of a label with 1 element, but none shown

Here we see `1 Issue` at label `fatal`, but there is no issue listed.

nicolai86:
According to the xorm logs, xorm ignores the num_issues and num_closed_issues columns when updating, even though the values did change.

This leads to wrong number of issues per label.

Listing the num_issue columns explicitly fixes the issue. 

However, we should consider just asking the DB for a correct count to fix existing, wrong counts. The query is basically instant for all use cases.

fixes #933 

\cc @Unknwon

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -561,7 +561,7 @@ func GetLabels(repoId int64) ([]*Label, error) {
 
 // UpdateLabel updates label information.
 func UpdateLabel(l *Label) error {
-	_, err := x.Id(l.Id).Update(l)
+	_, err := x.Id(l.Id).AllCols().Update(l)
 	return err
 }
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for your PR, you can just do `AllCols` :beer:

nicolai86:
updated the PR! Normally don't use xorm, thanks for the heads up.

Unknwon:
Thanks again!

PR #982, opened 2015-02-23
Properly calculate the number of open issues per label

### LINKED ISSUE #933: Issue labels: amount not updated on issue removal

nicolai86:
Instead of doing this on multiple places all over the place I've opt'd
to just fetching the correct issue counts from the database. This only
happens when calling UpdateIssueCounts()

should fix #933

### DIFF
--- a/routers/repo/issue.go
+++ b/routers/repo/issue.go
@@ -549,6 +549,7 @@ func UpdateIssueLabel(ctx *middleware.Context) {
 				label.NumClosedIssues--
 			}
 		}
+
 		if err = models.UpdateLabel(label); err != nil {
 			ctx.Handle(500, "issue.UpdateIssueLabel(UpdateLabel)", err)
 			return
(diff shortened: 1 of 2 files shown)


routers/repo/issue.go
```diff
@@ -538,17 +538,11 @@ func UpdateIssueLabel(ctx *middleware.Context) {
...
-			label.NumIssues++
-			if issue.IsClosed {
-				label.NumClosedIssues++
-			}
-		} else {
-			label.NumIssues--
-			if issue.IsClosed {
-				label.NumClosedIssues--
-			}
+		if err = label.UpdateIssueCounts(); err != nil {
```
Unknwon: I would still prefer fix count problem rather than do DB queries. :)

routers/repo/issue.go
```diff
@@ -767,6 +761,18 @@ func Comment(ctx *middleware.Context) {
...
+					send(500, nil, err)
+					return
+				}
+
+				if err = models.UpdateLabel(label); err != nil {
+					send(500, nil, err)
+					return
+				}
+			}
+
```
Unknwon: Looks like the issue is we didn't recount label when close/reopen

nicolai86:
still not done, tho. need to correct one more thing…

models/issue.go
```diff
@@ -525,6 +525,24 @@ type Label struct {
 	IsChecked       bool `xorm:"-"`
 }
 
+func (m *Label) UpdateIssueCounts() error {
```
Unknwon: Should this function be removed, or would you like to write anther huge function to recount all labels(then I can plugin that function into admin panel)

nicolai86: sorry, forgot about that method. it's gone now.

nicolai86:
I just played through the scenarios and it looks good to me. 
Do you have additional feedback, @Unknwon ?

Unknwon:
Thanks, merging!

PR #990, opened 2015-02-26
do some modify for supervisor run error

jacksonpan:
maybe you will find it, so i find it too, so i fix it. but not the best
solution way, and i am waiting the author will modify the start
scripts, and modify the gogs import conf way. i will update it when
next big version.


Unknwon:
Thanks your PR, but please re-PR to `dev` branch!

PR #992, opened 2015-02-27
Attempt 3 to fix LDAP login issues when git pushing after password change

theangryangel:
This fixes the issue for #898 (and I've actually tested it and got it running production this time).


Unknwon:
Thanks a lot! Appreciate your work!

Issue #738, opened 2014-12-12
At-mentions processed inside of indented code blocks


jcracknell:
Issue #178 brought this up, but looking at the code it was only fixed for fenced (```` ``` ````) code blocks, and not the indented-by-four-spaces variety, which is the most widely supported code block syntax.

A thought is it might be easier to process these as HTML, for example an at mention cannot be inside of `<[^>]+>`, `<pre[^>]*>((?!<\/pre>).)*<\/pre>` or `<code[^>]*>((?!<\/code>).)*<\/code>`.

Edit: or `<a>`.

Unknwon:
Closed by https://github.com/gogits/gogs/pull/1020

PR #1019, opened 2015-03-09
Exclude HTML tags from markdown post-processing

### LINKED ISSUE #637: Inline code markdown with leading hash
Starting an inline code snippet like \`#00ff00\` results in an attempt to create a link to an issue. It should actually result in monospace formatted text like `#00ff00`.

I am running Gogs 0.5.8.1116 Beta.

jcracknell:
Ok, I took a stab at this. Please review carefully, as I have no prior experience programming in go.

The general idea is to split the input into HTML and not-HTML. `GetPostProcessableParts` is fairly ugly, but I don't know that it can really be helped.

This would fix #637 and #738.

### DIFF
--- a/modules/base/markdown.go
+++ b/modules/base/markdown.go
@@ -105,12 +105,31 @@ func (options *CustomRender) Image(out *bytes.Buffer, link []byte, title []byte,
 	options.Renderer.Image(out, link, title, alt)
 }
 
+// See http://www.w3.org/TR/html-markup/syntax.html#attribute
+const HtmlAttributePattern =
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks your PR!

jcracknell:
I'm kind of inclined to retract this, I just noticed I am missing the call to `SanitizeBytes`, and after looking at that it seems this would be better implemented with `x/net/html`.

Unknwon:
@jcracknell go ahead and re-PR :)

PR #1020, opened 2015-03-09
Exclude HTML tags from Markdown post-processing

jcracknell:
Well that was easy. Again, this was for #637 and #738.


Unknwon:
Thanks again!

Unknwon:
Merging, thanks again!

PR #1031, opened 2015-03-10
Project-level Dockerfile & docker-compose script

voxxit:
Using this is easy:

1. Download & install `docker-compose` using instructions from here: https://docs.docker.com/compose/install/
2. Run `docker-compose up` from the project root
3. At the /install page, use "MySQL" for the database with the "root" user and "gogs" as the root password

That should be all you need to do to get running with Docker 1.5.0.


Unknwon:
Thanks a lot!

PR #1041, opened 2015-03-13
Fix for deleting user when gogs instalation is on http://hostname/gogs

sternik:


Unknwon:
Thanks your PR!

But we don't accept PR on `master`, please re-PR to `develop`!

PR #1042, opened 2015-03-13
Add mysql unix socket support.

tomcatzh:
If the host setting looks like a unix socket (leading by char '/'), will use unix(host) as connection string


Unknwon:
Thanks your PR!

But we don't accept PR to `master`, please re-PR to `develop`!

tomcatzh:
OK

PR #1044, opened 2015-03-14
Add mysql unix socket support.

tomcatzh:
If the host setting looks like a unix socket (leading by char '/'), will use unix(host) as connection string


tomcatzh:
should I pull to feature/pull_request?

Unknwon:
@tomcatzh no no no, that's a git-flow thing...

PR #1047, opened 2015-03-14
Various accessibility fixes

ndarilek:


Unknwon:
Thanks your PR!

But we use icon not "words" for a reason, and where are the `.less` changes for `sr-only` class?

Unknwon:
I'm sorry to say this but I would not merge this one.

ndarilek:
"sr-only" is a part of Bootstrap. See its documentation here, along with
other accessibility tips:

http://getbootstrap.com/components/

You may use icons for a reason, but as a blind person that makes your
app unusable to me. If you're not going to use Bootstrap then that's
fine, but please don't dismiss entire populations with this design practice.

Thanks.

Unknwon:
> http://getbootstrap.com/components/

Thanks for the link!

> You may use icons for a reason, but as a blind person that makes your
app unusable to me. If you're not going to use Bootstrap then that's
fine, but please don't dismiss entire populations with this design practice.

I agree, will consider use `icon + short word` pattern.

ndarilek:
So is this still not being merged? If not, I'd appreciate it if you'd
label these icons yourselves. I was looking for an alternative to my
current Redmine/Gitolite setup which is memory-hungry and not as
integrated as I'd like. Gogs looks nice but your icon use makes it
unusable to me. Thanks for using the title attribute, but my screen
reader only reads those links when I tab to them. If I arrow over the
page, which I typically do when exploring a new app, it's a bunch of
"link link link" which is not very speech-friendly. :/ I'm sure I missed
some icons, so it would be helpful for others to catch any I didn't, and
to use this or a similar technique when adding new ones in the future.

Thanks.

dyzajash:
Hymm what about making another option in config for this case? 
Something like:

'readeable menu' : true (false default) 

And if true swap incons in menu for pure text.

ndarilek:
Why not just hide the icon text entirely like sr-only does? Then I have
to ask people's permissions to use text-only when HTML can attach
invisible text to icons just fine.

What's wrong with my patch as submitted? It uses a standard Bootstrap
class that's been present for a long time.

ndarilek:
OK, maybe I'm not understanding this.

The text in the sr-only spans should not be visible. Is it still
appearing on-screen?

I myself am blind so I can't confirm, but the sr-only class should make
the contents of the span invisible to all non-screen readers. There
really should be no need for a config option for this. If there is, I'll
need to ask anyone running a gogs instance to enable it or I can't
participate on their instance, which is unnecessary given that HTML lets
you attach invisible text descriptions to icons. There's no reason that
universal design can't be used here as it is everywhere else.

Thanks.

Unknwon:
> The text in the sr-only spans should not be visible. Is it still
appearing on-screen?

Thanks for the tip, I understand this now. Sorry about it, I never used `sr-only`, I thought it will appear on the UI and so it may break current UI layout.

Not sure how to support `sr-only` in the coming UI(we're not going to use bootstrap anymore), but we'll work it out.

Unknwon:
Hmm...something is going wrong, text are showing the UI and it breaks...

Unknwon:
OK, I realize we are not using bootstrap already in 95% of pages.

Unknwon:
I have to revert changes and see if we can come up a solution after 0.6 release.

@ndarilek 

So to help you use Gogs, what trick does `sr-only` do? Just hide text near icon, or what?

I think we should open an issue discuss on this, what do you think?

ndarilek:
sr-only makes the text in the spans only appear with screen readers. So
if I put:

<span class="sr-only">Star</span>

near a star icon, you won't see the word Star but I will. Without that,
the icon only speaks as "button" or "link" which isn't helpful,
especially when there are half a dozen in a row. The title attribute is
slightly helpful but it only speaks the link text when it is focused,
and much of my webpage interaction involves scrolling through the site
with arrows.

Here is the sr-only class from Bootstrap, maybe it can be included in
any custom styles:

.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  border: 0;
}
.sr-only-focusable:active,
.sr-only-focusable:focus {
  position: static;
  width: auto;
  height: auto;
  margin: 0;
  overflow: visible;
  clip: auto;
}


Sorry for the confusion, I didn't realize the text was visible. I
thought you just didn't want to include the extra classes. :)

Unknwon:
No problem!

I've created issue https://github.com/gogits/gogs/issues/1054 as a backlog to keep in mind for `sr-only` thing when write UI.

Does this approach sound good to you?

PR #1081, opened 2015-03-18
Remove use of PID and fix stderr redirection

CarlosGtrz:
I tested the script in a Centos 6.6 server and it didn't work. Looks like `$!` didn't write the pid. After removing the use of pid it worked correctly.  
Also, the redirection should be `> ${LOGFILE} 2>&1` instead of `2>&1 > ${LOGFILE}` (see http://stackoverflow.com/a/637838/237761)


Unknwon:
Thank you very much!

PR #1105, opened 2015-03-20
Allow to migrate through git:// protocol

linquize:
Allow to migrate through git:// protocol


Unknwon:
Thanks your PR!

Unknwon:
Merging, thanks again!

Issue #692, opened 2014-12-01
Occasional panic on user login


DerDackel:
We've struggled with an issue for a few weeks now in our local Gogs deployment. Occasionally on user login the whole app will crash with a nil pointer dereference. This happens about once a day and I had hoped to be able to track down more info so I had delayed opening this issue, but I couldn't dig up additional info. I couldn't reproduce it or add any meaningful info, so all I have is a goroutine dump happening immediately after authentication.

All I can say at this point
 * All users this has happened on were authenticated using LDAP (but might be unrelated as pretty much only the root user isn't on LDAP.
 * It made no difference whether authentication was successful or not.
 * Authentication usually shows up in the macaron log as the last thing before the crash.
 * I can't exactly trace back to when it started, but it has basically happened to us on any 0.5.x build, up to at least 0.5.7.1105 which we currently have deployed.

Here's the dump that is produced on crash:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0x9349be]

goroutine 73 [running]:
runtime.panic(0xb347a0, 0x10a0eb3)
        /home/sjackel/opt/golang/current/src/pkg/runtime/panic.c:279 +0xf5
github.com/gogits/gogs/modules/asn1-ber.readBytes(0x0, 0x0, 0xc20873600e, 0x2, 0x2, 0x0, 0x0)
        /home/sjackel/opt/golibs/src/github.com/gogits/gogs/modules/asn1-ber/ber.go:181 +0xbe
github.com/gogits/gogs/modules/asn1-ber.ReadPacket(0x0, 0x0, 0x0, 0x0, 0x0)
        /home/sjackel/opt/golibs/src/github.com/gogits/gogs/modules/asn1-ber/ber.go:194 +0xaf
github.com/gogits/gogs/modules/ldap.(*Conn).reader(0xc2082c6480)
        /home/sjackel/opt/golibs/src/github.com/gogits/gogs/modules/ldap/conn.go:259 +0xa5
created by github.com/gogits/gogs/modules/ldap.(*Conn).start
        /home/sjackel/opt/golibs/src/github.com/gogits/gogs/modules/ldap/conn.go:83 +0x2f

goroutine 16 [IO wait]:
net.runtime_pollWait(0x7fc1b2a37e68, 0x72, 0x0)
        /home/sjackel/opt/golang/go-1.3.3/src/pkg/runtime/netpoll.goc:146 +0x66
[...]

PR #1147, opened 2015-03-25
fixed panic error on user login

blemasle:
It does not fix the real error behind the scene, but at least it prevent gogs from going into panic.
See https://github.com/gogits/gogs/issues/692


Unknwon:
Thanks for the patch!

PR #1152, opened 2015-03-26
Fix "Plus" icon size in dashboard

davmrtl:


Unknwon:
Thanks your PR!

But we do not accept PR to `master` and we use `.less` to generate `.css`.

PR #1165, opened 2015-03-27
Hide SSH button when SSH is disabled

linquize:
Hide SSH button when SSH is disabled

### DIFF
--- a/templates/repo/bare.tmpl
+++ b/templates/repo/bare.tmpl
@@ -21,8 +21,8 @@
                 <div class="panel-content">
                     <div id="repo-clone" class="clear text-center">
                         <h2>{{.i18n.Tr "repo.clone_this_repo"}}</h2>
-                        <button class="btn btn-blue current left btn-left-radius" id="repo-clone-ssh" data-link="{{.CloneLink.SSH}}">SSH</button>
-                        <button class="btn btn-gray left" id="repo-clone-https" data-link="{{.CloneLink.HTTPS}}">HTTPS</button>
+                        <button class="btn btn-blue current left btn-left-radius" {{if $.DisableSSH}}style="visibility: hidden" {{end}} id="repo-clone-ssh" data-link="{{.CloneLink.SSH}}">SSH</button>
+                        <button class="btn {{if $.DisableSSH}}btn-blue{{else}}btn-gray{{end}} left" id="repo-clone-https" data-link="{{.CloneLink.HTTPS}}">HTTPS</button>
                         <input id="repo-clone-url" type="text" class="ipt ipt-disabled left" value="{{.CloneLink.SSH}}" onclick="this.select()" readonly />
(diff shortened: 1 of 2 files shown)


d-fens:
Does this not also require changes to the following lines in templates/repo/bare.tmpl

```
git remote add origin <span class="clone-url">{{if $.DisableSSH}}{{.CloneLink.HTTPS}}{{else}}{{.CloneLink.SSH}}{{end}}</span>
```

and 

```
<pre class="text-left radius"><code>git remote add origin <span class="clone-url">{{if $.DisableSSH}}{{.CloneLink.HTTPS}}{{else}}{{.CloneLink.SSH}}{{end}}</span>
```

I also found I had to remove the following from public/ng/css/gogs.css to correctly align the copy button when displaying the new repo page. This appears to have no effect on the clone drop down on the existing repos.
```
.repo-bare #repo-clone-copy {
  /*margin-right: 200px;*/
}
```

lunny:
Any new idea? @linquize @d-fens ? Should I merge this pr?

templates/repo/bare.tmpl
```diff
@@ -21,8 +21,8 @@
                 <div class="panel-content">
                     <div id="repo-clone" class="clear text-center">
                         <h2>{{.i18n.Tr "repo.clone_this_repo"}}</h2>
-                        <button class="btn btn-blue current left btn-left-radius" id="repo-clone-ssh" data-link="{{.CloneLink.SSH}}">SSH</button>
-                        <button class="btn btn-gray left" id="repo-clone-https" data-link="{{.CloneLink.HTTPS}}">HTTPS</button>
+                        <button class="btn btn-blue current left btn-left-radius" {{if $.DisableSSH}}style="visibility: hidden" {{end}} id="repo-clone-ssh" data-link="{{.CloneLink.SSH}}">SSH</button>
```
tboerger: I think adding a ```hidden``` css class would be a better approach instead of using inline CSS.

templates/repo/header.tmpl
```diff
@@ -18,9 +18,7 @@
                 </a>
                 <div id="repo-header-download-drop" class="drop-down">
                     <div id="repo-clone" class="clear">
-                        {{if not $.DisableSSH}}
-                        <button class="btn btn-blue left btn-left-radius" id="repo-clone-ssh" data-link="{{$.CloneLink.SSH}}">SSH</button>
-                        {{end}}
+                        <button class="btn btn-blue left btn-left-radius" id="repo-clone-ssh" {{if $.DisableSSH}}style="visibility: hidden" {{end}} data-link="{{$.CloneLink.SSH}}">SSH</button>
```
tboerger: The same as above.

tboerger: I just realized it afterwards, why do you replace the existing check? Just not printing it out if SSH is disabled looks even better for me.

linquize: Because there are layout issues on Firefox / narrower browser width.
I do not want to change the left button's css (relates to layout issues),
so I just hide the button

PR #1166, opened 2015-03-28
Do not hardcode your personal path to gogs executable

linquize:
Do not hardcode your personal path to gogs executable.
Use the path of current gogs instead.


lunny:
This is embbed ssh server feature, it's not be open because it's on process.

Knorkebrot:
Why was this even commited? :)
Please merge this.

tboerger:
:+1: for merging this bug fix, even if this feature is not used yet.

PR #1167, opened 2015-03-28
Set Content-Type to text/plain for http status 401

linquize:
Set `Content-Type` to `text/plain` for http status 401
This is because git command line shows the failure reason only if `Content-Type` is `text/plain`.


Unknwon:
Thanks a lot!

PR #1177, opened 2015-03-31
fix typo in centos init script

rauchrob:
We should be *sourcing* `/etc/sysconfig/gogs`, not *executing* it, don't we?


Unknwon:
Thanks!

But please PR to `develop` branch next time!

PR #1182, opened 2015-04-02
Remove extra space from SSH authentication message

lukad:
This simply changes the message that a user gets when connection via ssh without a given command from

> Hi user ! You've successfully authenticated, but Gogs does not provide shell access.

to

> Hi, user! You've successfully authenticated, but Gogs does not provide shell access.

[Plenks](https://en.wikipedia.org/wiki/Plenken) are evil ;)


lukad:
Sorry, wrong destination branch... I have no clue how to change it so I'm going to close this PR and create a new one.

PR #1185, opened 2015-04-04
Update build.sh for Windows

linquize:
On Windows, the executable name is `gogs.exe`

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -4,8 +4,14 @@ rm -rf $outPath
 mkdir $outPath
 
 go build ../gogs.go
-chmod +x gogs
-mv gogs $outPath/
+PLATFORM=`uname | cut -d _ -f 1`
+if [ $PLATFORM = "MINGW32" ] || [ $PLATFORM = "MINGW64" ] || [ $PLATFORM = "CYGWIN" ]; then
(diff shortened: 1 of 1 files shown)


lunny:
The build.sh is only provided for unix-like system. I think we need make `build.bat` to build on windows.

linquize:
On Windows, we can also use bash to execute.

tboerger:
But a ```.bat``` would make much more sense as we don't need to rely on Cygwin.

linquize:
He can use git bash, which comes with git-for-windows, does not need cygwin

linquize:
Another question: the script runs with error, even on Linux
Is it for designed for normal people who wants to build their own gogs binaries?

where is `start.sh`?
```
cp: cannot stat ‘dockerfiles/’: No such file or directory
cp: cannot stat ‘../cert.pem’: No such file or directory
cp: cannot stat ‘../key.pem’: No such file or directory
cp: cannot stat ‘start.bat’: No such file or directory
cp: cannot stat ‘start.sh’: No such file or directory
cp: cannot stat ‘../wercker.yml’: No such file or directory
```

Unknwon:
THanks!

PR #1190, opened 2015-04-06
Add scala as a repository language

tszym:
Hi,

This is a work in progress for scala integration, as parts are still missing.

I tried to add support for Scala as a repository language.
It's sad that I didn't found docs about how to easily add a language.

As a non-go programer, the most painful part was the `byte[]` for default introduced by e6cf83b. This is why I didn't fill it for now.

The `public/js/lib.js` also seems to be related to languages support, but is not really easy to read, so not easy to modify.

I hope we could work on this to be able to support new languages and edit this PR with more informations.

Regards


tboerger:
If you follow https://github.com/tszym/gogs/blob/scala-language/conf/README you should get the bindata stuff generated.

tszym:
Thanks for your input!

I tried ti use go-bindata 3.1.0 with go 1.4.2.
Unfortunately it looks like I cannot generate bindata in the good way.

I assume the error could be in the input directories I specify. The README file says `conf/...` but this obviously generate errors. So I tried `conf`, `conf/..` and `conf/gitignore/Scala`.
For all the attempts, the generated file contains way less stuff than the commited one.
I can find a binary representation for the default Scala gitignore, so I just generated it and added it manually in the last commit. But I don't think this is the right way to do.

Hope you can help about this.

tboerger:
@tszym gogs is not maintained anymore as the only guy with write permissions does not respond anymore. Feel free to open a new pull request at https://github.com/go-gitea/gitea which is our fork of the project. There is already an open issue (https://github.com/go-gitea/gitea/issues/10) where i suggested that we need a proper documentation how to update the bindata content.

tszym:
Thanks @tboerger I'll do this soon.

tboerger:
```
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/jteeuwen/go-bindata.safeFunctionName(0xc20801e360, 0x12, 0xc20803c450, 0x0, 0x0)
	/home/tboerger/Projects/golang/src/github.com/jteeuwen/go-bindata/convert.go:235 +0x78f
github.com/jteeuwen/go-bindata.findFiles(0xc208032650, 0xe, 0x0, 0x0, 0x1, 0xc20801e100, 0xc208058040, 0x1, 0x1, 0xc20803c450, ...)
	/home/tboerger/Projects/golang/src/github.com/jteeuwen/go-bindata/convert.go:210 +0xbd8
github.com/jteeuwen/go-bindata.findFiles(0x7fff6f604fb4, 0x4, 0x0, 0x0, 0x7fff6f604f01, 0xc20801e100, 0xc208058040, 0x1, 0x1, 0xc20803c450, ...)
	/home/tboerger/Projects/golang/src/github.com/jteeuwen/go-bindata/convert.go:176 +0x610
github.com/jteeuwen/go-bindata.Translate(0xc208062000, 0x0, 0x0)
	/home/tboerger/Projects/golang/src/github.com/jteeuwen/go-bindata/convert.go:34 +0x1fe
main.main()
	/home/tboerger/Projects/golang/src/github.com/jteeuwen/go-bindata/go-bindata/main.go:20 +0x34

goroutine 2 [runnable]:
runtime.forcegchelper()
	/usr/lib64/go/src/runtime/proc.go:90
runtime.goexit()
	/usr/lib64/go/src/runtime/asm_amd64.s:2232 +0x1

goroutine 3 [runnable]:
runtime.bgsweep()
	/usr/lib64/go/src/runtime/mgc0.go:82
runtime.goexit()
	/usr/lib64/go/src/runtime/asm_amd64.s:2232 +0x1

goroutine 4 [runnable]:
runtime.runfinq()
	/usr/lib64/go/src/runtime/malloc.go:712
runtime.goexit()
	/usr/lib64/go/src/runtime/asm_amd64.s:2232 +0x1
```

This is what you get?

tboerger:
Ok, latest master branch of go-bindata is broken. tag 3.0.7 seems to work but there are big differences.

tszym:
Using `-data` I have a different output than whithout it in the bindata file. Also without `-data`, the stdout is empty, and not with the option. I find it weird.

Do we really need default files binary-hardcoded for `.gitignore` and licence files? I would much more prefer to have Gogs failing to start with an explicit error message when the file is missing in config.

tboerger:
I'm working on an update within gitea for the bindata stuff, afterwards there will be some properly documentation.

vtolstov:
I think we need go generate with stringer or simething like this. To store data inside binary.

tszym:
Or we could just not store data inside binary as it brings an overhead that seems not really useful.

vtolstov:
@tszym in case of languages licenses and other stuff - yes, i think this is need to be in plain files on fs.

tboerger:
Maybe this issue will be obsolete for gitea if we integrate https://github.com/go-gitea/gitea/issues/46

Unknwon:
Hmm.. there is a conflict... cannot merge. :flushed:

PR #1194, opened 2015-04-08
in routers/repo/download.go, don't set mime type, let go figure it out.

donbowman:
don't set mime type to text/plain. This will break serving e.g. raw images, html, etc

According to http://golang.org/pkg/net/http/#DetectContentType
"If the Header does not contain a Content-Type line, Write adds
a Content-Type set to the result of passing the initial 512 bytes
of written data to DetectContentType."

### DIFF
--- a/modules/base/markdown.go
+++ b/modules/base/markdown.go
@@ -198,7 +198,7 @@ func RenderRawMarkdown(body []byte, urlPrefix string) []byte {
 	extensions |= blackfriday.EXTENSION_FENCED_CODE
 	extensions |= blackfriday.EXTENSION_AUTOLINK
 	extensions |= blackfriday.EXTENSION_STRIKETHROUGH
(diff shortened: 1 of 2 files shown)


fiorix:
I also need this, it's currently breaking files served by /raw/* because it sets all Content-Type to text/plain instead of letting the ResponseWriter figure out the mime type.

lunny:
Please pull request to develop. Sorry for late.

donbowman:
I don't understand, this is a pull request?

fiorix:
Hey @lunny, this is a pull request already. You just have to merge.

brunogaspar:
@donbowman You'll need to send your Pull Request against the `develop` branch as @lunny requested as at the moment it was sent to the `master` branch.

@fiorix ^^

donbowman:
ah sorry, misunderstood. done

brunogaspar:
No biggie :)

PR #1216, opened 2015-04-18
SSH config options

### LINKED ISSUE #1215: Allow SSHDomain in configuration
Hi,

Users in some hosting situations, such as cloudflare, may not be able to use cloudflare + ssh from the same domain. In this case, they can get around this by being able to use their set their own SSH host and port. This way the clone address that shows up in repositories will be correct.

tony:
Fix #1215.


tony:
@Unknwon I built this branch on a production environment I'm using and it works well. Let me know if I overlooked anything.

lunny:
Thanks.

PR #1217, opened 2015-04-20
Blank name fix

### LINKED ISSUE #1207: Bad LDAP Filter Breaks Gogs
An LDAP filter doing something dumb like including all objects (oops), results in creating users with blank user names. You'll get authed as this blank user account, and can create repos and wreck all sorts of havock.

This pretty much breaks everything in Gogs, in a really bad way, you can't delete these repos from the UI (404 errors trying to access them) due to the invalid path that is given due to the lack of a user name, I didn't notice any issues on the filesystem caused by this but it makes me paranoid.

Mainly under no condition should Gogs ever accept a blank user name, including over LDAP.

StrangeWill:
Fixes #1207.

Fails my Travis build, but so does the develop branch which succeeds on the official Travis builds, so I'm interested in why my Travis builds from my repos fail but gogs' official repos are fine...

Should pass on your Travis server.


StrangeWill:
Mumble grumble, didn't select the right branch...

PR #1226, opened 2015-04-23
Add PAM authentication

nacho:
We've done this patch to add the PAM authentication and it seems to work for us. There are a couple of issues though:
1. we don't know how to make it conditional for just linux servers
2. the new label for some reason is not translated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ go:
   - 1.4
   - tip
 
-sudo: false
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -y libpam-dev
 
 script: go build -v
 
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -619,6 +619,7 @@ auths.smtp_auth = SMTP Authorization Type
 auths.smtphost = SMTP Host
 auths.smtpport = SMTP Port
 auths.enable_tls = Enable TLS Encryption
+auths.pam_service_name = PAM Service Name
 auths.enable_auto_register = Enable Auto Registration
 auths.tips = Tips
 auths.edit = Edit Authorization Setting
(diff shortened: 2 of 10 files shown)


pbor:
I fixed the conditional compilation, you now need -tags pam, otherwise you get a stub that always refuses auth.

lunny:
Discuss:  use `pam_linux.go` or `+build linux` and `+build !linux` to instead `+build pam`. Because Go compiler will recognize the conditional compiling `linux`. Then for linux binary, it will include the pam support and others will not.

pbor:
I used +build pam because pam is available also on osx. But I can change it as you prefer.

lunny:
OK. Then maybe you can use `+build !windows`.

nacho:
Changed. Do you have any other comments?

lunny:
No. But the CI report it not build success. Could you checked if it has conflicts with other pr?

nacho:
Maybe because the build now depends on pam-devel?

lunny:
`pam-devel` is a c++ library?

nacho:
It is a C library see https://git.fedorahosted.org/cgit/linux-pam.git/

nacho:
Is there a way to tell to travis that now we depend on that pam library?

nacho:
It seems to build now. I added libpam-dev to travis.

lunny:
Thanks.

nacho:
Thanks to you for the quick review.

Unknwon:
Hi @nacho , thanks your work again!

I just made one change about this PR:

I noticed some linux versions do not come with required C header files, so make install process `go-get`able, I change to use build tag `pam` to support PAM auth: https://github.com/gogits/gogs/commit/81adbe17caf4ff8d7fc920dfe661640ab248ad12 , also updated `.travis.yml` file as an example of how to use build tag.

Best regards,
Jiahua Chen

pbor:
@Unknwon , thanks, having a pam tag is ok.

It would be great if future binary releases are built with the tag enabled

Unknwon:
@pbor It would be certain that binary releases are built with the tag enabled. :dizzy:

PR #1234, opened 2015-04-25
Add commands to generate apache/nginx config

pbor:
This gives a handy way to generate configuration files for apache and nginx with all the right options properly filled in based on the contents of app.ini

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -17,12 +17,13 @@ import (
 
 var CmdGenerate = cli.Command{
 	Name:  "generate",
-	Usage: "Generate web server configuration files",
+	Usage: "Generate self-signed certificate, apache and nginx configuration files",
 	Description: `Generate configuration files to use a web server as a proxy for Gogs.`,
 	Flags: []cli.Flag{
 		cli.StringFlag{"config, c", "custom/conf/app.ini", "Custom configuration file path", ""},
 	},
 	Subcommands: []cli.Command{
+		CmdCert,
 		{
 			Name: "apache",
 			Usage: "generate Apache configuration file",
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -1,0 +1,145 @@
+// Copyright 2015 The Gogs Authors. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+package cmd
+
+import (
+	"os"
+	"bufio"
+	"log"
+	"strings"
+	"text/template"
+	"github.com/codegangsta/cli"
+
+	"github.com/gogits/gogs/modules/setting"
+)
+
+var CmdGenerate = cli.Command{
+	Name:  "generate",
+	Usage: "Generate web server configuration files",
+	Description: `Generate configuration files to use a web server as a proxy for Gogs.`,
+	Flags: []cli.Flag{
+		cli.StringFlag{"config, c", "custom/conf/app.ini", "Custom configuration file path", ""},
+	},
+	Subcommands: []cli.Command{
+		{
+			Name: "apache",
+			Usage: "generate Apache configuration file",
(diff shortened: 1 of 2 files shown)


tboerger:
I'm not sure if this is really useful as webserver configs are most of the time really opinionated.

vtolstov:
I think this is not needed, simple add example config to /usr/share/gogs/ (in case of packages) and use it if need. For me - i use nginx with custom domains and sometimes with upstreams, so this is not needed feature for me.

tboerger:
Maybe it makes more sense to add an example to a contrib directory that gets packaged into ```/usr/share/gogs/contrib``` as stated by @vtolstov. that would make much more sense. So this feature is for me :-1:. Let's wait for @lunny or @Unknwon.

pbor:
The idea is to have an official reference of the basic configurations: if you look at the documentation today there are many different configs. Beside the patch helps generating the files from the gogs config file instead of having to keep them in sync manually.

Obviously this "not needed" and you are not forced to use it, but it would make setting up gogs even faster, make the documentation for a standard installation more streamlined and allow to more easily script a deployment.

I do not see any downside in having these commands as opposed to shipping examples in a contrib dir that would probably get out of date or even worse having a zillion different configs on the wiki...

tboerger:
And why shouldn't it get out of sync as a command? This is a weak argument :)
I think adding it into the project source (not on a wiki) in a contrib folder should be totally fine.

pbor:
They will not go out of sync because if the doc says "execute gogs nginx" and copy the file, people will complain if it does not work. If instead the file is in contrib and people have to manually edit it anyway, they will just fix it up locally and not care.

Besides, if for instance I change the local port in the ini file I can regenerate the file automatically instead of having to manually edit.

As I said I do not consider this a fundamental feature, but I think it is pretty common that gogs is deployed by a developer (like me) and not by a sysadmin, and I never remember the right apache/nginx configuration, so I had to spend time finding documentation, reading examples, figuring out the small differences on the different wiki pages.
When I set up gogs I would have very much enjoyed this command and that is why I spent some extra time contributing this patch, so that I can avoid going through the same process when I'll have to redeploy gogs in 6 months for some other project... I still fail to see what are the downsides in having this included.

pbor:
I have update the pull request to use a single "generate" command, with "apache", "nginx" and "cert" subcommands

tboerger:
@Unknwon what's up with that? Close it?

Unknwon:
@tboerger yes, will do!

pbor:
I do not particularly care, but can you provide the rationale for wontifxing this? I see the question about properly setting up apache/nginx etc regularly asked and I still do not see any downside to this patch...

tboerger:
I think mostly because it's a onetime command, take it once and never again. That can be simply better documented... Or just take docker than it's eveen easier.

vtolstov:
I think that apache/nginx configuration is for wiki not for command. Somebody want public address of gogs, somebody wants localhost, so you always need to edit generated file. Mos of the time nginx config equal for many projects on go because it simply proxy_pass

pbor:
@tboerger cert is one time command and it is included

@vtolstov if you looked at the patch you see that the generated file takes the values from gogs configuration. I am not sure about "most of the time", but I am sure that this would have made my life easier when I set thing up in my deployment and I am pretty sure I am not the only one judging by the questions that are frequent on the chat

Anyway you already replied before, my comment was mostly for @Unknwon since he is the one calling the shots

tboerger:
> @tboerger cert is one time command and it is included

I don't like that command as well, but it have already a part of it before i got to gogs. Beside that generating correct certs is a bigger effort than a simple nginx/apache config ;). But true, that can be documented on the wiki as well.

Unknwon:
I'm trying to make a simple generator in admin panel other than a command.

PR #1249, opened 2015-04-30
in routers/repo/download.go, don't set mime type, let go figure it out.

donbowman:


tboerger:
An explanation for the blackfriday change should be interesting. The mime change LGTM.

Unknwon:
Thanks!

PR #1256, opened 2015-05-08
Add/Fix app sub url for color picker js lib inclusion

mrburns:
Currently, the js file doesn't take into account the possibility of a sub url being used (wich is my case :) wich prevents its inclusion.


tboerger:
That's a useful bugfix. Never used Gogs in a subdirectory :)
:+1:

Unknwon:
Thanks!

PR #1262, opened 2015-05-09
fix systemd "Trailing garbage, ignoring." warning.

chadoe:
Environment should be a space-separated list instead of comma-separated.
No need for double quoting when the values don't contain spaces.

http://0pointer.de/public/systemd-man/systemd.exec.html#Environment=


tboerger:
I would prefer a single line per environment variable.

Unknwon:
Thanks!

Issue #1025, opened 2015-03-10
[SECURITY] Debian init script runs daemon as root


FiloSottile:
https://github.com/gogits/gogs/blob/5e763baa12e346f8eb577073399068802fa5d67a/scripts/init/debian/gogs#L51-L53

The gogs binary itself is usually owned by a regular user - as encouraged, but the debian init script will run it as root with `--test`, allowing privilege escalation.

All the scripts must take great care to only run the daemon under the target user.

The project might consider getting a CVE ID for this.

Unknwon:
Thanks for the report!

All scripts are from 3rd-party, we personally don't use init.d, so don't know how to correct it.

Would you like you to make a PR on this? :blush:

FiloSottile:
I guess it's the least I can do, considering how I like the project :)

But with my security hat on I'll have to stress the importance of making sure "blessed" system scripts that will run as root are secure. I can take of the Debian one.

Please don't sacrifice users security to convenience.

Unknwon:
I should definitely learn init.d when I got time(famous false statement)!

Would you mind or possible change the script to run under user `git`? That is the recommend user to run Gogs.

FiloSottile:
Yeah, I'm making that change soon and reviewing the init script. It does run under `git` except on one call, but that's enough to become root on a system.

What I'm saying is that stuff that runs as root is delicate, and should not be shipped/blessed unless fully understood.

Unknwon:
> What I'm saying is that stuff that runs as root is delicate, and should not be shipped/blessed unless fully understood.

True.

hickscorp:
I have my own script that runs as root but starts the daemon as another user. Give me a moment and I will post it.

Pierre.

Unknwon:
> I have my own script that runs as root but starts the daemon as another user. Give me a moment and I will post it.

cc @FiloSottile

hickscorp:
Hello again,

```shell
#! /bin/sh
### BEGIN INIT INFO
# Provides:          gogs
# Required-Start:    $syslog $network
# Required-Stop:     $syslog
# Default-Start:     2 3 4 5
# Default-Stop:      0 1 6
# Short-Description: A self-hosted Git service written in Go.
# Description:       A self-hosted Git service written in Go.
### END INIT INFO
# Author: Danny Boisvert

# PATH should only include /usr/* if it runs after the mountnfs.sh script
PATH=/sbin:/usr/sbin:/bin:/usr/bin
[...]

PR #1274, opened 2015-05-13
fix debian init script

Knorkebrot:
My git user has it's shell set to `/bin/false`, which makes `su -c` impossible, also a flag was misplaced and this fixes that the test run is executed as root. (#1025)

Disclaimer: I'm not an experienced Debian nor System-V init user, maybe someone could review this commit before it gets merged.


tboerger:
:+1:

lunny:
Hi, could you pr to https://github.com/go-gitea/gitea?

Unknwon:
Thanks!

PR #1275, opened 2015-05-13
Add file permalink button

jstasiak:
The button appears when a file is viewed in a branch or a tag. It points
to a URL containing the branch's (or tag's) current commit id so that
it'll always point to the same content.

### DIFF
--- a/conf/locale/locale_de-DE.ini
+++ b/conf/locale/locale_de-DE.ini
@@ -341,6 +341,7 @@ releases=Veröffentlichungen
 file_raw=Roh
 file_history=Verlauf
 file_view_raw=Ansicht Roh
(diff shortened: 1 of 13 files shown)


tboerger:
IMHO The translation string have to be in all files.

jstasiak:
@tboerger I just pushed an updated commit rebased on current develop.

tboerger:
:+1:

tboerger:
@jstasiak can you create a pull request for go-gitea/gitea? :)

jstasiak:
@tboerger I didn't know about that project, I'll send a pull request there too, thanks for the info :)

Unknwon:
Thanks!

PR #1277, opened 2015-05-14
Fix: "Port is blocked" issue

winfr34k:
Fixed an issue where the post intall file for Debian enforced port 6000
which should have been port 3000 because it renders the software virtually unusable
because it's a blocked port in Firefox, Chrome and alikes.

For further reference see:
http://www-archive.mozilla.org/projects/netlib/PortBanning.html


tboerger:
This change looks ok, but anyway i would suggest to put Gogs always behind a reverse proxy like Nginx.

winfr34k:
@tboerger: If you can't finish the initial install of gogs, putting it behind a reverse proxy is impossible because competing the setup process doesn't work.

tboerger:
That's true

Unknwon:
Thanks!

PR #1284, opened 2015-05-20
allow to specify WorkDir via env var

vtolstov:
Sometimes (when package software) binaries lie on /usr/bin/ or /bin/, some path can be disable for execute for security reason, so providing WorkDir as env var may be very useful.

Signed-off-by: Vasiliy Tolstov <v.tolstov@selfip.ru>

### DIFF
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -162,8 +162,12 @@ func ExecPath() (string, error) {
 
 // WorkDir returns absolute path of work directory.
 func WorkDir() (string, error) {
-	execPath, err := ExecPath()
-	return path.Dir(strings.Replace(execPath, "\\", "/", -1)), err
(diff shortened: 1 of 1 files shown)


tboerger:
:+1:

vtolstov:
@Unknwon ping... i'm use this on production by two weeks and nothing bad happening =)

tboerger:
As it got a fallback to the current behavior it should be fine.

tboerger:
Another note, maybe we should use another environment variable name? ```WORK_DIR``` is a pretty common env variable that can be already set. Maybe something like ```GOGS_WORK_DIR``` should be better?

Dr-Terrible:
:+1: 

As a side note, all _environment variables_ belonging to Gogs should be prefixed with _GOGS__ prefix to avoid clashing with env vars from other applications.

PR #1288, opened 2015-05-22
changing concordance and the first letter to upper-case

alarangeiras:
In addition I want to make a proposal: In portuguese it is better to leave some technical terms in english. I realize that you are using "Branch" as "Ramo", we all know this term as branch here in Brazil.
If you agree i can make the change.


tboerger:
:+1:

Unknwon:
Hi, I've made your change at https://crowdin.com/translate/gogs/1/enus-ptbr#q=manage_access_token where all Gogs translation should be done. Thanks!

PR #1309, opened 2015-06-03
settings/password: no minimal required characters for OldPassword

stettberger:
If users are added to gogs due a reverse proxy automatically, their password is
set to their initial username. If their username is shorter than 6 characters,
they can never set a password. But since this password is required for some
operations (e.g. repository deletion), they could not use all features of gogs
until now.


Unknwon:
Thanks!

Issue #1304, opened 2015-06-01
Is gogs dead?


jsieler:
Is the project mainly dead? Since over a month there is nothing new... Anyone knows?

fuxiaohei:
we are waiting for @Unknwon coming back from US. He's a student.

If something gets worse, we will make new plan for gogs in June or July.

tboerger:
@fuxiaohei accepting for contributors for accepting pull requests would already help. Or just opening the contribution to the long missing feature for pull requests :)

fredghostkyle:
@fuxiaohei Do we have a ETA to when he's back? Also, just because you're missing your leader doesn't mean nothing gets done. That might sound mean and really harsh but it's true.

tboerger:
If more people would have write access, yes... We will get an answer really pretty soon, i think tomorrow.

vheuken:
@tboerger I don't see why that prevents work from getting done. Why can't someone just fork it and merge pull requests to that fork, then merge that when he gets back?

tboerger:
https://github.com/go-gitea/gitea

srvg:
What's the deal with go-gitea exactly? A real fork? The project repo doesn't even mention it's a fork from here. What are your intentions @tboerger ?

Unrelated: can we have a mailing list and an official irc room? Github issues s**k for support stuff.

tboerger:
go-gitea will mention that, just give us sone time. There is already the official Gitter chat :)

rojaro:
@srvg ... check https://gitter.im/gogits/gogs

redelman:
Since gogs has been forked to go-gitea, does that mean that gogs is indeed dead? Wondering if I should bother updating my gogs at work to go-gitea.

tboerger:
@redelman we need a little bit more time to get a new stable release of gitea out of the door, afterwards it makes sense.

iraycd:
This topic is trending on Hacker News. https://news.ycombinator.com/item?id=9708832

jh86:
It seems pretty ridiculous that people are complaining about a lack of updates for free (as in beer) software. "Be the change you wish to see in the world", fork the project, make the changes you need, and stop waiting for someone else to do it. Unless I'm misunderstanding something here -- are you guys paying his salary?

tboerger:
Nobody earns money with gogs/gitea. And it looks like some people on HN thought that i have forked gogs, that's not true, I just started to dive into contribution ;)

[...]

PR #1317, opened 2015-06-08
fix .keys route

smcquay:
This change fixes the output from /{{ username }}.keys so that it can work in
a ~/.ssh/authorized_keys file


Unknwon:
Thanks!

smcquay:
So what is the plan with regard to go-gitea?

Unknwon:
See https://github.com/gogits/gogs/issues/1304

smcquay:
Ah, thanks!

PR #1321, opened 2015-06-09
Fix command used by FreeBSD init script

wfriesen:
I recently installed Gogs on a FreeBSD system and found that the provided init script refers to start.sh which does not exist any more (since 6c84223). This change gets it working again.

Thanks.


Unknwon:
Thanks!

PR #1355, opened 2015-07-09
Fix build on packager.io, enable Debian 8 and CentOS / RHEL 7

crohr:
Looks like the pam library is now required to properly build the binary. Therefore I added the missing dev and runtime dependencies to the package, and enabled Debian 8 and CentOS / RHEL 7 now that they are supported on https://packager.io


Unknwon:
Thanks!

Issue #1231, opened 2015-04-24
generated certificate should have /CN with the hostname


pbor:
I used the command suggested in the FAQ to generate a certificate, however when I tried to use it with apache and chrome, chrome refused to open the site (firefox gave the warning about the fact that it was self-signed, but then let me through).

[paolob@code gogs]$ ./gogs cert -ca=true -duration=8760h0m0s -host=myhost
2015/04/24 16:30:00 Written cert.pem
2015/04/24 16:30:00 Written key.pem
[paolob@code gogs]$ openssl x509 -in cert.pem -subject -noout
subject= /O=Acme Co

I asked for help to my sysadmin and he told me that to be a valid certificate it should at least have a /CN=hostname in the openssl dump

flotho:
@pbor, self certificate are always a problem with git.
You have 2 frequent issues that I've already faced to. 
The first one is that the SSL is self signed and it could be passed by the git conf for this repository.
Additionnaly, if the SSL host haven't been correctly set (CN parameter MUST be the same as your host), you won't be able to clone the repo without chnaging the conf of git globally.

Hope it helps

pbor:
Thanks Florent. Yes, I am aware of those issues and I have solved the
problem in my installation. This report is about improving the certificate
generation tool included in gogs.
Il 26/Apr/2015 10:06, "Florent THOMAS" <notifications@github.com> ha
scritto:

Unknwon:
Closed by https://github.com/gogits/gogs/pull/1358

PR #1358, opened 2015-07-11
Add CommonName to the self signed certificate

pbor:
Without a CN the self signed certificate is considered
invalid by chrome.
You can check with: openssl x509 -in cert.pem -subject -noout

Fixes issue #1231


Unknwon:
Thanks!

Issue #1361, opened 2015-07-13
Constructing SSH address with default port error


coenm:
Gogs renders a wrong ssh address to clone the respository. It is not prefixed with "ssh://" and the domain and path are sperated by ":" instead of a "/".

Unknwon:
Commented on your PR. :dizzy:

wqweto:
For ssh links to work it seems that `gogs-repositories` has to be home directory of `git` or any other `setting.RunUser` which is weird.

There is no way to specify a directory beneath `setting.SSHDomain` that could be symlinked to `gogs-repositories`. I tried putting it in `DOMAIN = server:gogs-repositories` but this produces invalid ssh links in the end.

tboerger:
You can put the repositories whereever you want, you just need to tell gogs where they are. The SSH routes are always the same, but this is related to the magic within authorized_keys.

wqweto:
Any pointers/what to search for on the topic of "the magic within authorized_keys"?

tboerger:
Just take a look at the generated ~/.ssh/authorized_keys, this is openSSH magic. It's limited to specific commands like the gogs handler (which knows the path to the repositories from the configuration of course)

wqweto:
Thank you, I will.

Unknwon:
Close for now and open for continuing discussion.

PR #1362, opened 2015-07-13
#1361 Bugfix constructing SSH address with default port 

coenm:
Now, Gogs renders the right ssh clone address when running the ssh deamon on the default port.

Issue #1361 (https://github.com/gogits/gogs/issues/1361)


linquize:
But the old one is also a valid URL format in git ssh protocol.

Unknwon:
Hmm... the current one is how GitHub and other git hosts produce a SSH clone URL, I don't think it's a bug? Or do you have any specific reason that it can actually cause a problem? Thanks!

tboerger:
This change is indeed useless :)

Unknwon:
Close for now and open for continuing discussion.

Issue #1392, opened 2015-07-25
Admin Config broken in develop


manfer:
```
template: admin/config:105:50: executing "admin/config" at <.Webhook.TaskInterva...>: TaskInterval is not a field of struct type interface {}
```

I'm not familiar enough with the project, nor with go, to totally understand the error but by inspecting models/webhook.go it looks like there is no such TaskInterval. So, maybe an option that has been removed?

Removing lines 104-105 from templates/admin/config.tmpl solves the problem but that would be the solution only if TaskInterval makes no sense anymore. Otherwise I don't know how to debug the problem. 

By the way, what is that admin/config page for? It is just an info page, isn't it? I can't change any value in there.

tboerger:
Rhis page is for display only. This option have been reppaced with the realtim webhook commit. This fix have to be added.

manfer:
Ok, I'm going to prepare a PR removing those lines from the template and the corresponding locale value/pairs.

tboerger:
This is the replacement: https://github.com/gogits/gogs/commit/fa298a2c30c358dbfa47fc123c6aca83fe9eb999#diff-c4893a32340ee2cee87e664d67f8abf3R95

manfer:
TaskList was showing minutes as unit, Does QueueLength integer a value in any unit?

Not sure if I follow the code properly but looks like if TaskInterval had 1 as default value, so 1 minute. And now QueueLength default is 1000. There is not a correlation between 1 minute and 1000 something. Don't know if it should be any correlation.

manfer:
I understand Queue Length has no special units so I have created the PR https://github.com/gogits/gogs/pull/1393 that should solve this issue. Thanks @tboerger for the explanation.

I updated the locale files but there must be something that I don't understand as the translation for QueueLength is not working yet even after a rebuild with `go build`. Some different build for locales maybe? Actually there are more locales that doesn't work on my installation as those from the issues pages.

Unknwon:
Hi @manfer , thanks your PR and for your questions:

1. `QUEUE_LENGTH` is the maximum length of queue(a Go cached channel size). It may not make sense to you but just like a bucket size if this way helps.

2. As @tboerger mentioned in #1393, you need have to run `go-bindata` command as described [here](https://github.com/gogits/gogs/blob/master/conf/README)

Unknwon:
I've did go-bindata thing for you and pushed to `develop` branch.

PR #1393, opened 2015-07-25
admin config fix

manfer:
The admin config page is broken because TaskList has been replaced by QueueLength as explained by @tboerger in the following issue https://github.com/gogits/gogs/issues/1392

This pull request should solve the problem.


tboerger:
The go-bindata needs an update :)

Unknwon:
Thanks your PR!

PR #1394, opened 2015-07-26
Fixed issues spanish localization

manfer:
The spanish localization file doesn't include the strings that are already localized in issues templates.

The bindata.go file contains hexadecimal values where someone could include fake data so I think the needed go-bindata should be done by project owner in a different commit. In my opinion is reasonable to do it this way. Hope that is ok.

### DIFF
--- a/conf/locale/locale_es-ES.ini
+++ b/conf/locale/locale_es-ES.ini
@@ -354,6 +354,34 @@ commits.date=Fecha
 commits.older=Anterior
 commits.newer=Posterior
 
+issues.new = Nueva Incidencia
+issues.new_label = Nueva Etiqueta
+issues.new_label_placeholder = Nombre etiqueta...
+issues.open_tab = %d abiertas
+issues.close_tab = %d cerradas
+issues.filter_label = Etiqueta
+issues.filter_label_no_select = Ninguna etiqueta seleccionada
+issues.filter_milestone = Milestone
+issues.filter_assignee = Asignada por
+issues.filter_type = Tipo
+issues.filter_type.all_issues = Todas las incidencias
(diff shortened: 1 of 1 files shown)


tboerger:
you can read [here](https://github.com/gogits/gogs/blob/master/conf/README) how to build the bindata on your own as part of this pull request ;)

manfer:
@tboerger Thanks. I understand how it works but as the bindata.go file will contain hexadecimal data and I could easily fake data there without the project responsibles finding easily that I had done that I think it is a better idea if go-bindata is not done by me.

Anyway, if I include the bindata.go file the only way in which they can be sure that I didn't include any fake encoded data is by doing the go-bindata themselves and comparing.

So in my opinion it is safer for the project if only people with enough rights do the go-bindata when necessary and include the new bindata.go file in a new commit that they can append to the PR if they think is required.

Unless the CI checks the localization files and the binddata.go file content match in which case I don't mind to include it on this PR or on future PRs,

tboerger:
To be honest, i18n data can be faked quite easy. Who confirms that you don't write something like "Smells like chicken"? ;)

tboerger:
And if you generate the bindata @Unknwon can clone the pr, run bindata command an check if only the timestamps are getting updated.

Unknwon:
Thanks your work!

However as [documented](https://github.com/gogits/gogs#overview) that Gogs uses https://crowdin.com/project/gogs to manage translations and be updated weekly.

Unknwon:
It is sad but everything you did will be overwrote when we download INI files and replace.

manfer:
No problem. I will look in that translation website. Thanks.

Unknwon:
@manfer sorry about it! And I think someone already did the translation which you can help look and check them!

PR #1395, opened 2015-07-26
User Issues Template fix

manfer:
```
template: user/issues:28:66: executing "user/issues" at <.Id>: Id is not a field of struct type *models.Issue
```

On the template we have to use ID instead of Id

The code sometimes use Id in some structures and ID in others. Even on same file like in `models/issue.go`. It is not something good and it is easy to make a mistake later.


Unknwon:
Thanks!

Issue #1088, opened 2015-03-18
Diff changeset view 


wqweto:
In .git/config I'm using i18n for non-unicode source code files to use cyrillic code page

    [i18n]
        commitencoding = cp1251

and in gogs source file view everything is ok. HTML page view source shows `content="text/html; charset=UTF-8"` and cp1251 to utf-8 is done server-side just fine.

Browsing commit diff view seems to always be using cp1252/Latin code page and my cyrillic texts go berserk. Gogs is setup on CentOS 6.6 behind nginx.

What tool do you use for diff or is it built-in? Is there something OS level I can tweak?

Unknwon:
Thanks your feedback!

Currently we use git-shell to generate diff and parse, then show on the web UI. 

BTW, what is your Gogs version?

wqweto:
I'm using Version: 0.5.16.0316 Beta

Unknwon:
OK, the way it works is when parse the diff, Gogs will auto-detect encoding and try to convert to UTF8. Haven't tried when set i18n in git/config.

wqweto:
I searched for "encoding" setting in `app.ini` with no luck. Have no idea idea how i18n is supposed to work in git-scm in first place.

SourceTree and GfW seem to understand `commitencoding` just fine. Will test github.com behavior and post here.

Unknwon:
Sure!

Actually I'm not if `.git/config` is actually for server side, maybe it's just for client side.

ZafiRUS:
Actually, diff view encoding breaks even if I use utf-8. File view is fine.
I have no encoding settings in git, I just use utf8 everywhere.

Unknwon:
Close by https://github.com/gogits/gogs/pull/1399

wqweto:
@ZafiRUS: After this patch putting `ANSI_CHARSET = Windows-1251` under `[repository]` section showed cyrillic letters in view and diff of ANSI files just fine here.

PR #1399, opened 2015-07-27
Allow explicit charset/codepage setting for ANSI source files to alle…

wqweto:
…viate issue #1088 

This is a simple fix to alleviate charset issues with ANSI (single-byte) files for old languages in online view file and diffs.

The trouble with auto-detection per se is that most of the source files in any programming language are usually in English. Only a small portion of the files (usually in string literals) can be for instance in cyrillic letters for Russion.

Current auto-detection is based on confidence levels and string literals are usually much less that language keywords in English, thus it fairly confidently defaults to ISO-8859-1 which is useless.

This patch allows to swap ISO-8859-1 with any other charset at service level as a whole by configuring `ANSI_CHARSET` value in `[repository]` section (ANSI files in all projects are affected). Obviously best would be to be able to configure this on a per project level but I'm not sure how this can be done.

### DIFF
--- a/modules/base/template.go
+++ b/modules/base/template.go
@@ -55,6 +55,9 @@ func ShortSha(sha1 string) string {
 func DetectEncoding(content []byte) (string, error) {
 	detector := chardet.NewTextDetector()
 	result, err := detector.DetectBest(content)
+	if result.Charset == "ISO-8859-1" {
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks!

Issue #1137, opened 2015-03-24
404 on files with url-encoded names


wqweto:
When I commit a file with name `ifcfg-eth1%3A1` it is not accessible from the web interface.

`/{project}/src/master/{folder}/ifcfg-eth1%3A1` returns 404

`ifcfg-eth1%3A1` is the Windows name of `ifcfg-eth1:1` copied over sftp from a Linux box.

Unknwon:
Hi, I've tested both file name `ifcfg-eth1:1` and `ifcfg-eth1%3A1` and commit, there seems no problem(404) and page looks normal.

Can you provide more details on this?

wqweto:
I still get 404 with Application Version: 0.6.3.0725 Beta

These are last couple of lines from `gogs.log`
```
[Macaron] Started GET /wqweto/Test/src/master/aaa for 127.0.0.1
[Macaron] Completed /wqweto/Test/src/master/aaa 200 OK in 28.754599ms
[Macaron] Started GET /wqweto/Test/src/master/ifcfg-eth2%3A1 for 127.0.0.1
[Macaron] Completed /wqweto/Test/src/master/ifcfg-eth2%3A1 404 Not Found in 24.6563ms
2015/07/26 22:42:17 [1;31m[view.go:56 Home()] [E] GetTreeEntryByPath: GetTreeEntryByPath: error not exist[0m
```
I'm on a CentOS 6.6 with nginx in front in simple proxy_pass mode
```
  location / {
    proxy_pass      http://localhost:3000;
  }
```
404 is returned in both http status code and a pretty customized page from gogs.

Unknwon:
Well, my question is how `ifcfg-eth2%3A1` file is generated?

wqweto:
I made some progress, I was able to access the file by manually editing url to `/{user}/{project}/src/master/ifcfg-eth1%253A2` i.e. I had to url_encode the % in the file name.

The real file name is `ifcfg-eth2%3A1` on an NTFS volume (Windows box). This name has to be url_encoded to `ifcfg-eth1%253A2` in the actual link address when listing files in repo directory for the link to be correctly decoded w/o converting `%3A` to `:` because file `ifcfg-eth2:1` is not part of the repo and consequently `GetTreeEntryByPath` complains.

Unknwon:
Closed by #1402

PR #1402, opened 2015-07-28
Allow file and directory names to contain percent character

wqweto:
This is a quick fix for #1137

`EscapePound` could be expanded to full blown url_encode probably using URL from `net/url` but I'm not sure how this will affect performance as it's a fairly frequently used function in templates.

Speaking of templates, every href attribute has to use `EscapePound` on user generated content i.e. various names (repo, user, directory, file) or risk invalid links.

### DIFF
--- a/templates/repo/home.tmpl
+++ b/templates/repo/home.tmpl
@@ -48,7 +48,7 @@
                     {{ $l := Subtract $n 1}}
                     {{range $i, $v := .Treenames}}
                         {{if eq $i $l}}
-                        <span class="bread">{{EscapePound $v}}</span>
+                        <span class="bread">{{$v}}</span>
                         {{else}}
                         <span class="bread"><a href="{{EscapePound $.BranchLink}}/{{index $.Paths $i}}">{{$v}}</a></span>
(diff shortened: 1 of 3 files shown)


templates/repo/home.tmpl
```diff
@@ -48,7 +48,7 @@
                     {{ $l := Subtract $n 1}}
                     {{range $i, $v := .Treenames}}
                         {{if eq $i $l}}
-                        <span class="bread">{{EscapePound $v}}</span>
+                        <span class="bread">{{$v}}</span>
```
Unknwon: Why did you remove this one? Is it related to #1137?

wqweto: Yes, instead of `fcfg-eth1%3A1` it was showing filename as `ifcfg-eth1%253A2` i.e. double escaped. Same would happen for filenames with # I suppose, don't know what the original intent was.

Unknwon: Make sense, maybe just was a mistake!

Unknwon:
Thanks!

PR #1404, opened 2015-07-28
Allow space and/or percent in file and directory names

wqweto:
These are just the hrefs that my quick testing revealed to be most painful for my repos.

### DIFF
--- a/models/git_diff.go
+++ b/models/git_diff.go
@@ -167,8 +167,8 @@ func ParsePatch(pid int64, maxlines int, cmd *exec.Cmd, reader io.Reader) (*Diff
 				return diff, nil
 			}
 
-			fs := strings.Split(line[len(DIFF_HEAD):], " ")
-			a := fs[0]
+			beg := len(DIFF_HEAD)
+			a := line[beg : (len(line)-beg)/2+beg]
 
 			curFile = &DiffFile{
 				Name:     a[strings.Index(a, "/")+1:],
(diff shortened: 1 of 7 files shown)


models/git_diff.go
```diff
@@ -164,11 +163,11 @@ func ParsePatch(pid int64, maxlines int, cmd *exec.Cmd, reader io.Reader) (*Diff
...
 		if strings.HasPrefix(line, DIFF_HEAD) {
 			if isTooLong {
-				return diff, nil
+				break
 			}
 
-			fs := strings.Split(line[len(DIFF_HEAD):], " ")
-			a := fs[0]
+			beg := len(DIFF_HEAD)
+			a := line[beg : (len(line)-beg)/2+beg]
```
Unknwon: Just to confirm, is it guess based on assumption that length of two files are equal?

wqweto: Yes, because these two are the same file with different prefixes like `a/file` and `b/file`. A string split on " b/" is the stronger alternative (than current " ") but could fail on some file names too.

Someone else has to confirm if file rename is impl with remove + add by git though.

Unknwon: Thanks for explanation!

Let me do some tests and see if handles normal remove/add cases.

Unknwon: @wqweto looks like normal add/remove/rename still uses same file name, so split in half should work in most of cases, thanks!

Unknwon:
Thanks again!

PR #1410, opened 2015-07-30
Show pagination only if needed on repo issues page

manfer:
On repository issues page only show the pagination if needed. It is needed if one of the following conditions is true:
* It is not the first page (not IsFirst) because in that case is needed always.
* Or if we are in the first page (previous condition is false and that's why this second condition is evaluated, so it is the first page) and has more than one page (HasNext).

### DIFF
--- a/templates/repo/issue/list.tmpl
+++ b/templates/repo/issue/list.tmpl
@@ -87,6 +87,7 @@
       {{end}}
 			
 			{{with .Page}}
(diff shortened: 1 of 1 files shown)


manfer:
Better to do it just with:
```
{{if gt .Total 1}}
```
I changed the PR.

Unknwon:
Looks good! Thanks!

PR #1414, opened 2015-07-31
Factor out function to get the current user

pbor:
The same logic was duplicated in three places. Factor it
out in a single utility and improve it to also try
os/user.Current(). We need all the fallbacks because in
some cases it is better to not rely on the env vars, but
in other cases os/user.Current() fails with
"user: Current not implemented on linux/amd64".

### DIFF
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -24,6 +24,7 @@ import (
 	"github.com/gogits/gogs/modules/bindata"
 	"github.com/gogits/gogs/modules/log"
 	// "github.com/gogits/gogs/modules/ssh"
+	"github.com/gogits/gogs/modules/user"
 )
 
 type Scheme string
@@ -309,10 +310,7 @@ func NewConfigContext() {
 	}[Cfg.Section("time").Key("FORMAT").MustString("RFC1123")]
 
 	RunUser = Cfg.Section("").Key("RUN_USER").String()
-	curUser := os.Getenv("USER")
-	if len(curUser) == 0 {
-		curUser = os.Getenv("USERNAME")
-	}
+	curUser := user.CurrentUsername()
 	// Does not check run user when the install lock is off.
 	if InstallLock && RunUser != curUser {
 		log.Fatal(4, "Expect user(%s) but current user is: %s", RunUser, curUser)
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks your PR!

If I'm not remember wrong, `curUser, err := user.Current()` runs incredibly slow in Windows and it uses CGO which Gogs tries to avoid at most in current stage.

pbor:
maybe we should first try the env vars and then fallback to user.Current() ?

The reason to use user.Current() is that another user reported that the env vars are not available during the init scripts.

I can also split the PR in two patches, one that just factors out the module/function and then we can add more fallback logic in a later step, at least it would be done in a single place

Unknwon:
> The reason to use user.Current() is that another user reported that the env vars are not available during the init scripts.

Does user `os/user` solve his/her problem? If he/she uses `linux/amd64`, `os/user` will not work either.

> I can also split the PR in two patches, one that just factors out the module/function and then we can add more fallback logic in a later step, at least it would be done in a single place

Yes, please, and put get env vars before `os/user` could be better.

Unknwon:
How's going on the PR?

Unknwon:
I really don't see the value of using `"os/user"`, just change this PR to a simple factor out is good enough.

tboerger:
For reference the pull requests merged to gitea related to that:

https://github.com/go-gitea/gitea/pull/17
https://github.com/go-gitea/gitea/pull/29

Unknwon:
Thanks @tboerger , but I hold my opinion on not using `os/user`.

tboerger:
And what about a fallback if the env stuff fails?

pbor:
(note that the PR has been amended to just do the refactoring in the mean time)

Unknwon:
Thanks @pbor .

@tboerger OK, this is the problem. It is not implemented in linux, and slow on Windows, so what does `os/user` serves for? Mac only?

tboerger:
https://golang.org/src/os/user/lookup_unix.go there is a bunch of systems that support it :)

tboerger:
And as mentioned it seems to be not in general slow on windows :)

Unknwon:
> And as mentioned it seems to be not in general slow on windows :)

People don't blame Go, they blame Gogs.

linquize:
What about https://github.com/go-gitea/gitea/blob/6e15dcd076dc5520d13fe987f94f62b7e11cc5c4/modules/user/user.go#L24 ?

tboerger:
@linquize as @Unknwon said it seems to be damn slow on windows and is not implemented on linux

Issue #1415, opened 2015-07-31
Markdown in Gogs incorrectly renders nested lists when nested list uses bullets


trejder:
According to [Markdown Basics](https://help.github.com/articles/markdown-basics/#nested-lists) you can have nested list in form of:

```markdown
1. Item 1
  1. A corollary to the above item.
  2. Yet another point to consider.
2. Item 2
  * A corollary that does not need to be ordered.
    * This is indented four spaces, because it's two spaces further than the item above.
    * You might want to consider making a new list.
3. Item 3
```

In Gogs only first scenario (both outer and inner list uses numbers) works fine. When you have second scenario (outer list uses number, inner list uses bullets), this is ignored and you end up with dual number nested list. See below images as examples.

From left to right: Markdown code | rendering effect in GitHub | rendering effect in Gogs:

Case 1: Inner list uses dashes and items are separated with blank line:

![c1-items-separated-by-blank-line-using-dashes](https://cloud.githubusercontent.com/assets/2903392/9003037/c47b01b8-376a-11e5-84f5-11699e9a22a1.png)

Case 2: Inner list uses asterisks and items are separated with blank line:

![c2-items-separated-by-blank-line-using-asterisks](https://cloud.githubusercontent.com/assets/2903392/9003059/fa9e22ca-376a-11e5-9512-4000ea9f584c.png)

Case 3: Inner list uses asterisks, but there is no blank line separation:

![c3-no-blank-line-separation](https://cloud.githubusercontent.com/assets/2903392/9003068/0871e21a-376b-11e5-92b0-1d99081f2a49.png)

You can also notice, that in case 1 and 2, when there is a blank line separation after each item, inner list also has an ugly empty line intro in the beginning.

Please, consider fixing this ASAP, as nested lists are very useful for step-by-step guides or TODO lists. Currently, they can't be used in Gogs, because rendered final effect is quite ugly.

Issue somehow related to #1416.

Unknwon:
Thanks a lot!

manfer:
I made this pull request to solve the nested unordered list inside an ordered list appearing as an ordered list.

About the the ugly line it must be solved in [blackfriday](http://github.com/russross/blackfriday) library. Or maybe there is an option to configure blackfriday in a way that it won't happen. We will need to append a new commit with the configuration solution if there is one.

trejder:
Thank you for a super-fast response and fix! I'm shocked to see how quick things happens here! :>

Unknwon:
[...]

PR #1418, opened 2015-07-31
nested unordered list inside order list issue

manfer:
This will solve the nested unordered list inside an ordered list issue. It is just a CSS error.

Issue: https://github.com/gogits/gogs/issues/1415

About the other problem reported about an ugly line appearing that is a blackfriday issue (so it must be reported to that library) or a blackfriday configuration issue (don't know if there is a blackfriday extension to solve that or other configuration that could be included in gogs that would solve that).


Unknwon:
Thanks!

Issue #1422, opened 2015-07-31
Autofocus issue title field and fix tab index in issue editing screen


trejder:
Can we have `autofocus="autofocus"` added to `<input name="title">` field in Gogs, so clicking `New issue` would place cursor in that field by default? This would allow user to start typing issue title right away, without need for an additional click to title field.

Exactly as it works here, on GitHub. It speeds up work extremely, if you have to enter a lot of issues at once (i.e. upon project manual migration).

When editing existing issue, issue body field should be auto-focused, instead of issue title, as in most cases you're changing / fixing issue body, not title.

In addition, you should consider changing tab-index of objects in Issue edit screen (if this is possible) to require presing `Tab` button only once to go from editing issue title to writing issue body. Again, exactly as this happens in GitHub issues. Other elements in-between (assigned / milestone dropdowns, tabs) shouldn't be "reachable"  by `Tab` button at all or should have tab index set so would be active _after_ issue body, not _before_ it. Thank you.

Unknwon:
- [x] New issue
- [ ] Edit issue
- [x] New milestone
- [x] Edit milestone

Unknwon:
It is now possible to edit issue and comments! https://try.gogs.io/unknwon/gogs/issues

PR #1426, opened 2015-07-31
autofocus for new and edit issue forms

manfer:
Issue https://github.com/gogits/gogs/issues/1422 autofocus implemented.

About the tab index not sure what is explained there. Github has not such a simple tabindex as described. In gogs maybe it can be enhanced but I would not prioritize it.


Unknwon:
Thanks!

PR #1431, opened 2015-08-02
Install: Fix DB host not changed when selecting another DB type

linquize:
Install: Fix DB host not changed when selecting another DB type.


Unknwon:
Thanks your PR!

But the new `install.tmpl` is using JS file https://github.com/gogits/gogs/blob/master/public/js/gogs.js#L25 .

Did you run into any trouble by this?

linquize:
No problem at all.

Issue #1396, opened 2015-07-26
Copy link button


wqweto:
... seems to be implemented w/ Adobe Flash Player and the latter often gets disabled for security reasons. This is the result on Chrome. The button beneath the player can be clicked but nothing happens.

![Flash](https://dl.unicontsoft.com/upload/pix/ss_gogs_flash.png)

tboerger:
This is an external script... Maybe it can be configured different with a better fallback detection.
This is used: http://zeroclipboard.org/

wqweto:
[Cut and Copy Commands](https://developers.google.com/web/updates/2015/04/cut-and-copy-commands?hl=en) on Google Developers.

Firefox 41 seems to support these too.

tboerger:
> Cut and Copy Commands on Google Developers.
> Firefox 41 seems to support these too.

That's absolute not an option, to quote the support: "IE 10+, Chrome 43+, and Opera 29+ support these commands. Firefox supports this but requires a preference change (See cut and copy commands here). Without this Firefox throws an error. Safari does not support these commands.". That new requirements are just a no-go for this kind of project.

wqweto:
@tboerger: What kind of project is this? Can you just take a look how they did clipcopy on github.com... I bet there is no sign of Flash. (I'm positive as it works with Flash disabled in chrome)

manfer:
If I'm not wrong github uses [zeroclipboard](http://zeroclipboard.org/) which fallback to flash when necessary. Which is exactly the one used on this project as @tboerger said. I suggest rereading his first message more carefully.

manfer:
Current major version of zeroclipboard: 2.2.0
Version of zeroclipboard in github: 2.1.6
Major 1.x version of zeroclipboard: 1.3.5
Version of zeroclipboard in gogs: 1.1.1

Looks like the library needs to be updated.

But besides I'm not sure if zeroclipboard can work without flash. Maybe github is only using zeroclipboard as fallback if access to clipboard is not directly available on the browser.

manfer:
@wqweto I can confirm Github uses the clipboard API from the browser (cut and copy commands as in the link you provide) when it is available and only fallbacks to zeroclipboard if it is not available. So doing something similar would be an enhancement for sure.

manfer:
I link here the [discussion](https://github.com/zeroclipboard/zeroclipboard/issues/171) on zeroclipboard because it can contain useful info for ideas.

[...]

PR #1432, opened 2015-08-03
Enhance copy clipboard

manfer:
I have added clipboard API support when available. First commit parses User-Agent header to add a class to html tag if the browser supports copy to clipboard. The second commit implements the javascript that copies directly to clipboard if the browser supports it or otherwise uses jquery zclip which uses an invisible flash to deal with the copy.

More details on https://github.com/gogits/gogs/issues/1396

Probably the minified version `public/ng/js/min/gogs-min.js` needs to be rebuilded.

### DIFF
--- a/modules/middleware/repo.go
+++ b/modules/middleware/repo.go
@@ -16,6 +16,14 @@ import (
 	"github.com/gogits/gogs/modules/git"
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/setting"
+
+	"github.com/hashicorp/go-version"
+	"github.com/mssola/user_agent"
+)
+
+const (
+	FIREFOX_COPY_SUPPORT = "41.0"
+	CHROME_COPY_SUPPORT = "43.0.2356"
 )
 
 func ApiRepoAssignment() macaron.Handler {
@@ -345,6 +353,24 @@ func RepoAssignment(redirect bool, args ...bool) macaron.Handler {
 
 		ctx.Data["BranchName"] = ctx.Repo.BranchName
 		ctx.Data["CommitId"] = ctx.Repo.CommitId
+
+		userAgent := ctx.Req.Header.Get("User-Agent")
+		ua := user_agent.New(userAgent);
+		browserName, browserVer := ua.Browser()
+
+		sliceVer := strings.Split(browserVer, ".")
+		var max int
+		if max = len(sliceVer); 3 < max {
+			max = 3
+		}
+		browserVer = strings.Join(sliceVer[:max], ".")
+
+		browserVersion, err := version.NewVersion(browserVer)
+		chromeConstraint, err := version.NewConstraint(">= " + CHROME_COPY_SUPPORT)
+		firefoxConstraint, err := version.NewConstraint(">= " + FIREFOX_COPY_SUPPORT)
+
+		ctx.Data["BrowserSupportsCopy"] = (browserName == "Chrome" && chromeConstraint.Check(browserVersion)) || (browserName == "Firefox" && firefoxConstraint.Check(browserVersion))
+
 	}
 }
 
--- a/.gopmfile
+++ b/.gopmfile
@@ -33,6 +33,8 @@ golang.org/x/net = commit:937a34c9de13
 golang.org/x/text = commit:5b2527008a4c
 gopkg.in/ini.v1 = commit:caf3f03ad9
 gopkg.in/redis.v2 = commit:e617904962
+github.com/hashicorp/go-version = commit:999359b6b7
+github.com/mssola/user_agent = commit:f659b98638
 
 [res]
 include = etc|public|scripts|templates
--- a/public/js/app.js
+++ b/public/js/app.js
@@ -269,36 +269,62 @@ var Gogits = {};
         if ($(selector).hasClass('js-copy-bind')) {
             return;
         }
-        $(selector).zclip({
-            path: "/js/ZeroClipboard.swf",
-            copy: function () {
-                var t = $(this).data("copy-val");
-                var to = $($(this).data("copy-from"));
-                var str = "";
-                if (t == "txt") {
-                    str = to.text();
-                }
-                if (t == 'val') {
-                    str = to.val();
-                }
-                if (t == 'html') {
-                    str = to.html();
-                }
-                return str;
-            },
-            afterCopy: function () {
+
+        if ( document.documentElement.classList.contains("is-copy-enabled") ) {
+
+            $(selector).click(function(event) {
                 var $this = $(this);
-                $this.tooltip('hide')
-                    .attr('data-original-title', 'Copied OK');
+
+                var cfrom = $this.attr('data-copy-from');
+                $(cfrom).select();
+                document.execCommand('copy');
+                getSelection().removeAllRanges();
+
+                $this.tipsy("hide").attr('original-title', $this.data('after-title'));
                 setTimeout(function () {
-                    $this.tooltip("show");
+                    $this.tipsy("show");
                 }, 200);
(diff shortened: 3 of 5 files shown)


modules/middleware/repo.go
```diff
@@ -16,6 +16,14 @@ import (
 	"github.com/gogits/gogs/modules/git"
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/setting"
+
+	"github.com/mcuadros/go-version"
```
Unknwon: I would recommend to use https://github.com/hashicorp/go-version which has methods like `LessThan` `GreaterThan`, Gogs may use this one in the future, so it's better to start using it now.

modules/middleware/repo.go
```diff
@@ -16,6 +16,14 @@ import (
...
+	"github.com/mssola/user_agent"
+)
+
+const (
+	FIREFOXCOPYSUPPORT = "41.0"
+	CHROMECOPYSUPPORT = "43.0.2356"
```
Unknwon: Use format `CHROME_COPY_SUPPORT` could be better. :beer:

modules/middleware/repo.go
```diff
@@ -345,6 +353,19 @@ func RepoAssignment(redirect bool, args ...bool) macaron.Handler {
...
+		ua := user_agent.New(userAgent);
+		browserName, browserVersion := ua.Browser()
+		browserVersion = version.Normalize(browserVersion)
+
+		var chromeSupportsCopy bool
+    var firefoxSupportsCopy bool
+    chromeSupportsCopy = browserName == "Chrome" && version.Compare(browserVersion, CHROMECOPYSUPPORT, ">=")
+		firefoxSupportsCopy = browserName == "Firefox" && version.Compare(browserVersion, FIREFOXCOPYSUPPORT, ">=")
+
+		ctx.Data["BrowserSupportsCopy"] = chromeSupportsCopy || firefoxSupportsCopy
```
Unknwon:
```go
if (browserName == "Chrome" && version.Compare(browserVersion, CHROMECOPYSUPPORT, ">=")) ||
    (browserName == "Firefox" && version.Compare(browserVersion, FIREFOXCOPYSUPPORT, ">=")) {
    ctx.Data["BrowserSupportsCopy"] = true
}
```

Is it better or you prefer current one? Because variable `chromeSupportsCopy` and `firefoxSupportsCopy` only be used once.

Unknwon:
Thanks your PR!

I've made some comments to help your PR become better.

manfer:
Ok, I have tried and it is impossible. Until that library `github.com/hashicorp/go-version` is corrected it can't be used at all.

It generates a Panic error.

```
package main

import (
	"fmt"

	"github.com/hashicorp/go-version"
)

const (
	FIREFOX_COPY_SUPPORT = "41.0"
	CHROME_COPY_SUPPORT = "43.0.2356"
)

func main() {

  browserName := "Chrome"
  //browserVer := "44.0.2403.125 m (64-bit)" // NOT WORKING
  //browserVer := "44.0.2403" // WORKING
  //browserVer := "44.0" // WORKING
  //browserVer := "44" // WORKING
  browserVer := "44.0.2403.125" // NOT WORKING

  browserVersion, err := version.NewVersion(browserVer)
	chromeConstraint, err := version.NewConstraint(">= " + CHROME_COPY_SUPPORT)
  firefoxConstraint, err := version.NewConstraint(">= " + FIREFOX_COPY_SUPPORT)

  if (browserName == "Chrome" && chromeConstraint.Check(browserVersion)) || (browserName == "Firefox" && firefoxConstraint.Check(browserVersion)) {
    fmt.Println("IT IS TRUE")
  } else {
    fmt.Println("IT IS FALSE")
  }

  if err != nil { fmt.Println("JUST TO AVOID err NOT USED") }

}
```

manfer:
The `github.com/hashicorp/go-version` says in its README:

> Versions used with go-version must follow [SemVer](http://semver.org/).

I opened an issue there providing the code above but probably their reply is going to be. Those Chrome version numbers doesn't comply to Semver specs.

The only thing that comes to my mind in order to use `github.com/hashicorp/go-version` is just to remove anything but the three former numbers on the browser version which should be enough for the comparison.

manfer:
An interesting reading [Semver has failed us](http://www.jongleberry.com/semver-has-failed-us.html)

manfer:
I don't see anything on `github.com/hashicorp/go-version` which is not available on `github.com/mcuadros/go-version`. And the latter looks not tied to a specific versioning scheme like hashicorp version.

In hashicorp version you use v1.GreaterThan(v2) and in mcuadros version you use Compare(v1,v2,">").

In hashicorp version you have to convert your version strings into vesion objects. In mcuadros version you use the strings directly.

In hashicorp version to do a >= comparison you need to do two operations GreaterThan and Equal in mcuadros you can use >= operator in the Compare function directly.

I really don't know why you want to use hashicorp version in the future.

I would need to know if you still want to use hashicorp version in which case I would have to crop the browser version numbers coming from the user-agent header before feeding them to that library.

If mcuadros go-version is OK I will revert code back to use that library.

tboerger:
You are missing some `;`, that won't work minified.

manfer:
@tboerger Thanks. Good catch. I will solve that too in the javascript code. Let's see how we proceed with the go-version library and I will solve both problems.

manfer:
Now I pushed a working version using `github.com/hashicorp/go-version`. I had to add code to remove from the browser version anything above the third dot.

Unknwon:
@manfer thanks again!

My apology, do you mean that the old `version` package can deal with format `x.x.x.x m`? I just thought use method `Greater` `Less` `Equal` is more robust than string `>` `<` `=` and so if anything changes Go compiler can check for us. 

I think I'll just merge your PR and make a internal package `version` for Gogs `modules`.

Unknwon:
My apology again, changed back to use `github.com/mcuadros/go-version`. :mask:

Unknwon:
Now you can try it on https://try.gogs.io

manfer:
Looks fine.

The only thing I can see is in the last commit https://github.com/gogits/gogs/commit/e6792796cc25c9a6db91998599a47d3f64016975 in `.gopmfile` you are not including a commit SHA1 from mcuadros library. I suppose that mean it will always download the last commit from master branch. It should be ok as I doubt the developer is going to introduce a bug because it is unlikely he is going to create a new commit soon. Besides as you are only going to use the library temporarely must not be a problem.

If you want to add the commit SHA1 it will be `d52711f8d6`

manfer:
But what I can't do is check it on try gogs. Something is wrong there and shows these errors on the console.

> Refused to load the script 'https://code.jquery.com/jquery-1.11.1.min.js' because it violates the following Content Security Policy directive: "script-src 'none'".
> Refused to load the script 'https://try.gogs.io/ng/js/lib/lib.js' because it violates the following Content Security Policy directive: "script-src 'none'".
> Refused to load the script 'https://try.gogs.io/ng/js/lib/jquery.tipsy.js' because it violates the following Content Security Policy directive: "script-src 'none'".
> ...

Unknwon:
> The only thing I can see is in the last commit e679279 in .gopmfile you are not including a commit SHA1 from mcuadros library. I suppose that mean it will always download the last commit from master branch. It should be ok as I doubt the developer is going to introduce a bug because it is unlikely he is going to create a new commit soon. Besides as you are only going to use the library temporarely must not be a problem.

Oh yes, in the development, we always use the latest version unless there is a specific reason(build fail) for a commit, otherwise, only update `.gopmfile` before release. What's your opinion on this work flow?

> I forgot my browser scripts blocker. I had to allow the scripts. Works fine.

:joy: :beer:

PR #1440, opened 2015-08-05
syntax highlight with prism

manfer:
New syntax highlight. Bye bye prettyprint, welcome prism.

This is an starting point for this pull request as it is going to be difficult it doesn't need some tweak.

What is on it?

First of all we need to relax sanitizer of markdown parse output to allow the class attribute on the code tag. This is done on markdown.go:
```go
Sanitizer.AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code")
```
The regexp is to protect against inyections. Regexp borrowed from the example provided for the title  attribute in [bluemonday readme policy building section](https://github.com/microcosm-cc/bluemonday#policy-building)

Once we have our code with the `language-xxx` class we can apply syntax highlight only and just only on those fenced code blocks with a specified language instead of trying to highlight just everything.

The highlight has to be applied to the file view and not only on markdown output. So the `templates/repo/view_file.tmpl` has been edited to start using prism too.

But for that to work we have to guess which programming language is on the file. That is done in `routers/repo/view.go` where the programming language is guessed from the file extension. I have included by now all languages supported by prism. The extensions/languages map included on `routers/repo/view.go` has to be reviewed as we can maybe include more extensions (for example xhtml, xslt, ... can be rendered as markup language). Besides the extensions are case sensitive so maybe we have to include some extensions in different case too. Needs to be case sensitive as for example we want to identify .c as C language but .C as objective-C.

What is [Prism](http://prismjs.com/)?

Well, it is a javascript syntax highlighting library which in my opinion is getting really good. It is growing on languages supported as there are already PRs for new languages and its functionality can be extended with plugins. It already has some useful plugins.

The way to get prism is by downloading a customizing version from the page provided by the project, [Customize and Download Prism](http://prismjs.com/download.html). I have downloaded a development version (meaning not minified) with prism core (mandatory), default theme, all programming languages supported, and line highlight, line numbers, file highlight, show language, highlight keywords plugins.

I included the line highlight because it can be useful. I still don't know where but looks it can be. We need line numbers for sure. Show language is cool. Highlight keywords is great too and gives the oportunity to tweak the highlight CSS theme in the future a lot more. About, file highlight I decided not to use it as it will use ajax to read the file that has to be highlighted and not sure if it could be a security risk (probably not but I prefer to feed the code string to the pre tag in the template file).

After I had `prism.js` and `prism.css` I include then on gogs. The CSS file has been tweaked for gogs as `public/ng/css/gogs-prism.css`

The rest on the pull request has been get rid of prettyprint js code and minor CSS tweaks to `markdown.less` and `repository.less`.

What I'm not sure if needs to be edited?

I'm not sure about `public/ng/js/gogs.js` line 137- `Gogs.renderCodeView`. Looks its purpose is to deal with diff code views and that is still working properly but if this is the case I would suggest changing it and applying it only to `.code-diff` class and not `.code-view` which is more generic.

I would need to know more clearly the purpose of all that code.

And I don't know if some of the changes I made have to be included too in other files. Well, now I realize I didn't compiled just markdown.less to create `public/css/markdown.css but that should be done anyway automatically later if I'm not wrong.

But I don't know about any other file as it always confuses me having in `public/ng` and `public`, css, js, less, folders. And the same on templates don't know if I would have to edit other head.tmpl file too.

Well, in brief, **needs deep review**.


modules/base/markdown.go
```diff
@@ -250,6 +250,7 @@ func PostProcessMarkdown(rawHtml []byte, urlPrefix string) []byte {
 func RenderMarkdown(rawBytes []byte, urlPrefix string) []byte {
 	result := RenderRawMarkdown(rawBytes, urlPrefix)
 	result = PostProcessMarkdown(result, urlPrefix)
+	Sanitizer.AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code")
```
Unknwon: Doesn't `OnElements` return a policy object? Seems it is discard here?

routers/repo/view.go
```diff
@@ -238,3 +239,148 @@ func Home(ctx *middleware.Context) {
 	ctx.Data["BranchLink"] = branchLink
 	ctx.HTML(200, HOME)
 }
+
+func Language(ext string) string {
```
Unknwon: Does prismjs has something like auto-match language? So we don't have to rely on the extension?

Unknwon:
Thanks your PR!

But be sure ping me next time before giving this kind of big PR. :bowtie:

> But I don't know about any other file as it always confuses me having in public/ng and public, css, js, less, folders. And the same on templates don't know if I would have to edit other head.tmpl file too.

It is confusing, but at this point you only need to deal with `ng/` files, I'll migrate them when needed.  

How's Prismjs compares to https://highlightjs.org/ ?

Also, https://highlightjs.org/ seems to be able to change code theme at runtime, can Prismjs do this?(somehow I see ratio buttons in download page which only allows me choose one theme. Download all themes and let user choose which one they prefer, it feels awesome to me).

Unknwon:
Also, I've opened branch https://github.com/gogits/gogs/tree/feature/highlight , would you mind re-PR to that branch when we finish discussion here(comments that have been post)? :grinning:

modules/base/markdown.go
manfer: Sanitizer is a variable declared in tools.go and an its value is `bluemonday.UGCPolicy`. The only thing I'm doing is adding to that policy a new rule that allows `class` attribute on `code` tag matching the providing regexp.

Unknwon: Yeah, `Sanitizer` is a global variable, but since `OnElements` has return value, I suppose it has to be re-assign back to `Sanitizer`?

Also, shouldn't only do this once instead of repeating whenever renders a markdown?

Not sure about how `bluemonkey` works, I can be wrong.

manfer: We can read bluemonday documentation more in depth but I think it is ok how I did it. Just the UGCPolicy is extended with the rule I add. It works fine.

Unknwon: Well, then... Does this operation only need to do once or need to set it every time? I think it only needs to do once since the object is global.

routers/repo/view.go
manfer: No there is no auto-match to use in the file viewer. We have to provide the language ourselves to let the library know which language on the file. Prettyprint works the same if I'm not wrong and that's why it was barely working.

And I doubt it is easy to do it differently. Guessing the language from the source text would mean that the library must know all languages grammar and decide in which grammar is that code written. I bet is an almost impossible task or at least a huge task considering that even the code could have syntax errors.

Unknwon: Fair point, thanks for your explanation .

manfer:
Highlightjs is an option too. Looks good but it doesn't have [line numbers feature](https://github.com/isagalaev/highlight.js/blob/master/docs/line-numbers.rst). Nor I see if it is extendable with plugins. Looks like it is just extendable with language grammars and themes.

It shouldn't be difficult to include various themes of prism on the code and allow to change it dinamically. I haven't looked into highlightjs in depth to understand if it has an easy API to do such things but wouldn't be difficult anyway.

Unknwon:
Okay. 

Only the Sanitizer thing left for discussion before re-PR.

modules/base/markdown.go
manfer: Yes. It only needs to be done once. So maybe it is better to do it in other place. I tried in `tools.go` just after the line that initializes it but I got some error about that line of code not being allowed there outside a function and I inmediately thought my best option was to put it just in markdown.go. I'm too new to go to know a better place to put that line of code but probably there is a better place indeed.

Unknwon: The same place where Sanitizer is declared would be my choice. 

Something like: Sanitizer = UGCPolicy.AllowAttr.Match.OnElements

This way we can define rules all at once.

manfer: :+1:  I didn't thought about placing it just in the same line. I tried behind and I couldn't. I will send the PR to the other branch them with this change already applyed.

Unknwon: Great. See you in another PR. And,... It's 1am here, get back to you tomorrow!

Unknwon:
Closed by merging #1442

PR #1442, opened 2015-08-05
syntax highlight with prism

manfer:


Unknwon:
Thanks!

Unknwon:
All looks good but one important thing is missing, that line number(s) no longer clickable/selectable. :flushed:

Unknwon:
Single line: https://try.gogs.io/unknwon/gogs/src/master/gogs.go#L14
Multiple lines: https://try.gogs.io/unknwon/gogs/src/master/gogs.go#L18-L20

Unknwon:
In view file, white padding around code could be problem, too.

Unknwon:
A plugin might be helpful: http://prismjs.com/plugins/line-highlight/ I'm still looking.

Unknwon:
image

Style can be easily fix, the only thing left is the select line number, I looked over line-number plugin, I guess there is a way to add hook(either inside line-number plugin or another independent hook)

Unknwon:
More problems found: 

Plain file does not have line numbers anymore... https://try.gogs.io/unknwon/gogs/src/master/Dockerfile#L3

Unknwon:
Just for reference: #1440

manfer:
@Unknwon That's exactly what I was not sure about

> I'm not sure about public/ng/js/gogs.js line 137- Gogs.renderCodeView. Looks its purpose is to deal with diff code views and that is still working properly but if this is the case I would suggest changing it and applying it only to .code-diff class and not .code-view which is more generic.

Never realized there was that option on file view. I understand that is a useful feature to link to a line or to a range. I already included the line highlight plugin just in case something like that were possible. I will try to think how to implement it.

> In view file, white padding around code could be problem, too.

About the padding it is included on purpose on every view but it can be changed. But I wouldn't do it that way. First we have to agree about the html structure then change the CSS.

My idea is that `.panel-content` is a wrapper and must be there always. If we don't want padding on that wrapper we can just remove it.

Then if it is a markdown it will have `.markdown` wrapper div inside the `.panel-content` where we can put the extra padding.

If on the other hand it is a binary file it has `.view-raw` and we can add the padding there too.

And if its a source code file then it will have as wrapper `.code-view` and we can remove any padding there.

After having an agreement on that the CSS tweak shouldn't be a problem. Removing padding from `.panel-content`. Increasing a little the padding on '.markdown` wrapper. Adding padding to `.view-raw`if neccesary. And removing any padding if any from `.code-view`.

> Plain file does not have line numbers anymore...

I think that is [reported on prism](https://github.com/PrismJS/prism/issues/645) as an issue so I hope it will be solved soon.

manfer:
By the way now that we have the `language-xxx` class in code tag using one or other library would be just very simple. Any of prettyprint, prism, or highlightjs can be used. It is just a matter of adding the library and in the case of prettyprint and highlightjs just add the line of code that triggers the highlight. It the case of prettyprint being sure not to ask to highlight everything but only the codes that has a language-xxx class.

This is just as info so we can have it in consideration if in some moment we decide is better to replace the highlight library.

I like prism library anyway so I'm going to look for the solution to the lines highlight. Maybe there is a way by using what is already available on that library or otherwise tweaking those plugins to have a selection API for the lines.

manfer:
It should be possible already to highlight pieces of code dinamically if I understand correctly this [issue in prism](https://github.com/PrismJS/prism/issues/356).

```
Prism.highlightElement()
```
and data-line attribute should be what we are looking for. I'm going to play on the console how it works.

manfer:
Something like:
```
$("pre").attr("data-line", 16);
Prism.highlightAll();
```
works fine. Calling `Prism.highlightElement(our_pre_element)` which would be a better choice triggers an error. Maybe a bug in the library. But `highlightAll` is just fine. Added that to an onclick event to the line numbers shouldn't be a problem, I hope.

About links I would say that instead of an anchor link as it is now ...#L16 we should pass the highlight as a queryString so using your previous examples they would be turned into:
```
Single line: https://try.gogs.io/unknwon/gogs/src/master/gogs.go?lines="14"
Multiple lines: https://try.gogs.io/unknwon/gogs/src/master/gogs.go?lines="18-20"
```
Reading that queryString, maybe validating it with a regexp, and using it on the pre tag data-line attribute is just piece of cake.

The library allows even multiple ranges:
```
Multiple ranges: https://try.gogs.io/unknwon/gogs/src/master/gogs.go?lines="14, 18-20"
```

manfer:
The dynamic was not so simple that it seem at first glance. I had to edit the line numbers plugin to add that feature. It is just very little code but must be on that place. Here is the full code of the line numbers prism plugin. My added code is just at the end, the on click event binding.

```javascript
Prism.hooks.add('after-highlight', function (env) {
	// works only for <code> wrapped inside <pre> (not inline)
	var pre = env.element.parentNode;
	var clsReg = /\s*\bline-numbers\b\s*/;
	if (
		!pre || !/pre/i.test(pre.nodeName) ||
		// Abort only if nor the <pre> nor the <code> have the class
		(!clsReg.test(pre.className) && !clsReg.test(env.element.className))
	) {
		return;
	}

	if (clsReg.test(env.element.className)) {
		// Remove the class "line-numbers" from the <code>
		env.element.className = env.element.className.replace(clsReg, '');
	}
	if (!clsReg.test(pre.className)) {
		// Add the class "line-numbers" to the <pre>
		pre.className += ' line-numbers';
	}

	var linesNum = (1 + env.code.split('\n').length);
	var lineNumbersWrapper;

	var lines = new Array(linesNum);
	lines = lines.join('<span></span>');

	lineNumbersWrapper = document.createElement('span');
	lineNumbersWrapper.className = 'line-numbers-rows';
	lineNumbersWrapper.innerHTML = lines;

	if (pre.hasAttribute('data-start')) {
		pre.style.counterReset = 'linenumber ' + (parseInt(pre.getAttribute('data-start'), 10) - 1);
	}

	env.element.appendChild(lineNumbersWrapper);

	$(".line-numbers-rows > span").on("click", function() {
		var line = $(this).index() + 1;
		$(pre).attr("data-line", line);
		Prism.highlightAll();
	});

});;
```
First for that on click to work I had to relax in the CSS the rule that disables events on those elements in gogs-prism.css. Changing `pointer-events` to `auto` instead of `none`.

So there is no problem to add that feature.

We should agree about the HTML structure that I explained before. In my opinion it should be as explained.

###### For Markdown
```html
<div class="panel-content">
  <div class="markdown">
    the markdown output here
  </div>
</div>
```
###### For Raw files
```html
<div class="panel-content">
  <div class="raw-view">
    the file here. if an image just the image.
  </div>
</div>
```
###### For Source Code Files
```html
<div class="panel-content">
  <div class="code-view">
    the highlighted source code here.
  </div>
</div>
```
And that's why the template `view_file.tmpl` looked as follows in my pull request.
```
    <div class="panel-content">
    	{{if .ReadmeExist}}
        <div class="markdown">
    	  {{if .FileContent}}{{.FileContent | Str2html}}{{end}}
        </div>
      {{else if not .IsFileText}}
        <div class="view-raw">
            {{if .IsImageFile}}
            <img src="{{EscapePound .FileLink}}">
            {{else}}
            <a href="{{EscapePound .FileLink}}" rel="nofollow" class="btn btn-gray btn-radius">{{.i18n.Tr "repo.file_view_raw"}}</a>
            {{end}}
        </div>
    	{{else if .FileSize}}
        <div class="code-view" id="repo-code-view">
          <pre class="line-numbers"><code class="language-{{if .FileLanguage}}{{.FileLanguage}}{{else}}nohighlight{{end}}">{{.FileContent}}</code></pre>
        </div>
    	{{end}}
    </div>
```

manfer:
I forgot to mention that line numbers are not selectable and that's done on purpose on prism. And I agree. Almost 100% of the times people wants to copy just the code.

I never had the need to copy the line numbers. In fact it has been many times in the past that I had to remove line numbers manually because a website code snippets had numbers and there were no way to select just the code.

So only the code is selectable and it is on purpose.

About the dynamic highlight I don't see it very usefull either but anyway adding the snippet to line numbers plugin will provide that feature.

What is very very very useful for sure is the possibility to link to the code with highlighted parts. And that can be coded very very easily with a queryString.

I can start implementing all that but I would prefer to first agree on the html structure to change that either again and edit CSS to show source code without padding.

Unknwon:
Thanks for your explanation!

About the highlight thing I think we are on the same page.

But about line number, I insist that line numbers should be selectable, it is very useful to point other people to a specific position in a large file.

manfer:
But I wouldn't call it selectable but linkable. And yes I agree that linkable is a very very very useful feature and should be coded and it is very easy to implement using prism.

I'm going to start coding it all and will do a pull request againts this branch so you can test it all.

Unknwon:
Sure... linkable! :grinning:

manfer:
I have all implemented but for some reason the line numbers are out of place. I'm not able to figure it out just now, as much I looked into it I don't see anything strange and the line numbers are placed in a really strange place. I will look into it tomorrow. I will start again from the previous commit applying the changes to see if I figure out the problem.

Just a question that maybe you can help. As explained before the lines to highlight are going to come from a queryString on the url, like
```
Single line: https://try.gogs.io/unknwon/gogs/src/master/gogs.go?lines=14
Multiple lines: https://try.gogs.io/unknwon/gogs/src/master/gogs.go?lines=18-20
Multiple ranges: https://try.gogs.io/unknwon/gogs/src/master/gogs.go?lines=14,18-20,23-24
```
The code I have just now to read the query parameter looks like this
```go
					lines := ctx.Query("lines")
					//TODO: sanitization, remove spaces and validation if needed
					ctx.Data["HighlightLines"] = lines

```
And the question is what we have to do in that TODO. Because I don't know if Query already sanitizes the input. And what you think about removing spaces and validating that input with a regexp.

Unknwon:
Sounds great!

For your question:

No, Query only returns value of the key. 

For this specific case, I think only need to check if those are valid int numbers, if not then simply ignore them. Correct me if I understand your question wrong.

manfer:
Ok then we validate the string with a regexp that ensures is valid: it only can contain values separated by commas, and each of those values must be a valid range, being a valid range just a number or two numbers separated by an hyphen. I can allow spaces which don't have meaning too.

Tomorrow I will finish it.

Unknwon:
No problem, take your time!

PR #1447, opened 2015-08-07
Syntax highlight

manfer:


Unknwon:
Thanks, should I merge now or wait until more pushes?

Unknwon:
Can't wait to see how far it goes, merging...

Unknwon:
Amazing! I can highlight a line(although it's just one line and link doesn't automatically change for me). :+1:

manfer:
I think it is still not prepared to go into develop. I'm realizing the aproach by prism to include line numbers and to highlight is not so useful and that something like what was done by prettyprint and what is done in github is better.

So I will try to implement a plugin for prism to make line numbers and highlight in github style. And maybe look how highlightjs works either.

Unknwon:
Sure, I think GitHub is using custom version of highlightjs.

manfer:
For what I read github makes the highlight server side with a custom ruby version of [pygments](http://pygments.org/) called [linguist](https://github.com/github/linguist/tree/master/lib/linguist). Both are open source but there is no golang version.

Unknwon:
Well... that's sad. :disappointed_relieved:

PR #1450, opened 2015-08-08
scroll to first highlighted line

manfer:
* Scroll to first highlighted range (line) implemented.
* Line numbers even if the source language is not supported implemented.


Unknwon:
Thanks!

Unknwon:
Tested :+1: awesome!

PR #1451, opened 2015-08-08
Add check to see if comments on an issue are by the repo owner

jamesa:
Fixed a bug where all comments on an issue would be marked as the "Owner" of the repository.

Before:
screen shot 2015-08-08 at 5 18 54 pm

After:
screen shot 2015-08-08 at 5 19 39 pm


Unknwon:
Thanks!

Issue #1453, opened 2015-08-09
Empty repository page show "ssh" even when off


gustavopaes:
Hello,

In my app.ini I have defined DISABLE_SSH = true, but when I create an empty project, the SSH option is shown to clone.

![screenshot from 2015-08-09 13 51 17](https://cloud.githubusercontent.com/assets/227400/9155881/c4f1c03e-3e9d-11e5-926f-f52d0b2408d6.png)

Thanks.

Unknwon:
Thanks your report!

gustavopaes:
Can I resolve it?

Unknwon:
Hmm... if you mean make a Pull Request, Yes.

Otherwise, you can just wait for me to fix it for you.

gustavopaes:
Yes, I mean make a PR.

Unknwon:
Sure, I guess just hide "SSH" button can fix the problem.

gustavopaes:
Not just it, need to change the urls in field and help text too. But it's simple.

Unknwon:
Well, yes. Should be simple:)

Unknwon:
Closed by #1454

PR #1454, opened 2015-08-09
Remove SSH button when SSH it is disabled

gustavopaes:
#1453

### DIFF
--- a/public/ng/css/gogs.css
+++ b/public/ng/css/gogs.css
@@ -1496,7 +1496,7 @@ The register and sign-in page style
 .repo-bare #repo-header-meta {
   display: none;
 }
-.repo-bare #repo-clone-ssh {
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks! 

However Gogs actually uses less file to generate css files. 

I can do it for you tomorrow morning or you can update the PR. 😀

gustavopaes:
Oh really? I didn't know that. I can update the PR, no problem.

Unknwon:
In this case, it should be in somewhere public/ng/less/Gogs/repository.less

gustavopaes:
yes, I found it. How can I compile less in go projects? I didn't see any Grunt or Gulp (common in nodejs projects)

Unknwon:
There is a sh in scripts directory could be helpful. I personally use codekit.

PR #1460, opened 2015-08-10
new top menu

manfer:
This is my suggested top menu.

In some point we will need to remove the mouse over effect on the menu links (the logo, issue, and the dropdown trigger buttons). I decided to avoid doing it because that's probably better done by customizing semantic theme. So my tweaks to css are minimal.


manfer:
##### User Dropdown Menu
newtopmenu1

##### New (repo, organization, ...) Dropdown Menu
newtopmenu2

##### Popups
newtopmenu3

The logo acts as the link `/` and there is nothing else to do. If is a signed user it will link to dashboard and if it is anonymous user will link to home.

Unknwon:
Thanks!

PR #1461, opened 2015-08-10
Fix broken collaboration link in repo settings

jamesa:
Really small fix - broken link. :)


Unknwon:
Thanks! Very good fix!

Issue #1107, opened 2015-03-20
"Change your avatar" tooltip/link on any user page


xPaw:
Visiting any user profile and hovering on any profile will show a tooltip to change avatar on gravatar.com.

![2015-03-20-2048ljl05](https://cloud.githubusercontent.com/assets/613331/6758143/c95d11a6-cf3e-11e4-8537-0af2a4b7149e.png)

---

On an unrelated note, it would be a good idea to add width/height attributes to avatar so that content doesn't jump while image is being loaded. It also seems that gogs resizes images to 200px, but user profile displays avatar in 234px.

Unknwon:
> It also seems that gogs resizes images to 200px, but user profile displays avatar in 234px.

Good catch, thanks!

Unknwon:
Fixed by https://github.com/gogits/gogs/pull/1464

PR #1464, opened 2015-08-10
Hide avatar tooltips unless you're looking at your own profile

jamesa:
This hides the tooltips and links to configure your profile picture, unless it's your profile. This also links directly to the profile image, similar to how GitHub does, if you're looking at another user's profile.

Also included a commit to change the size of the resized uploaded avatars, from 200px to 234px, like was talked about in issue #1107. But not sure if this is how you wanted this fixed!

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -194,7 +194,7 @@ func (u *User) UploadAvatar(data []byte) error {
 	if err != nil {
 		return err
 	}
-	m := resize.Resize(200, 200, img, resize.NearestNeighbor)
+	m := resize.Resize(234, 234, img, resize.NearestNeighbor)
 
 	sess := x.NewSession()
 	defer sess.Close()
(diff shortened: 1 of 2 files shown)


Unknwon:
Looks good, thanks!

Unknwon:
https://try.gogs.io/unknwon is updated as well. :grinning:

jamesa:
@Unknwon not sure why I get an error on that new statement on that link. Obligatory "works on my PC." :P

Unknwon:
There is a template error, has pushed fix

jamesa:
Thanks, I thought I had to compare IDs not names.

Unknwon:
Normally, yes, compare ID.

But when there is no logged in user, `.SignedUser` is `nil` and Go template cannot compare with it. So it panics

PR #1466, opened 2015-08-11
modified repo header

manfer:
##### Private
newrepoheader

##### Fork
newrepoheaderfork

##### Mirror
newrepoheadermirror

Still not totally sure about the left part. So I maybe try a little more tomorrow. What do you think about this repo header?

Tomorrow I will set the breadcrumb to huge size again and add some padding to the left of forked from. I will review how it looks when mirror, repo and private.

### DIFF
--- a/public/less/_base.less
+++ b/public/less/_base.less
@@ -38,6 +38,9 @@ img {
 	.top.menu a.item:hover {
 		color: rgba(0,0,0,.45);
 	}
+	.top.menu .menu {
+		z-index: 900;
+	}
 	.head.link.item {
 		padding-right: 0!important;
 		.dropdown.icon,
@@ -74,12 +77,21 @@ img {
 		&.right {
 			text-align: right;
 		}
+		&.small {
+			font-size: 0.75em;
+		}
 	}
 
 	.message {
 		text-align: center;
 	}
+
+	.header > i + .content {
+		padding-left: 0.75rem;
+		vertical-align: middle;
+	}
 }
+
 footer {
 	margin-top: @footer-margin!important;
 	height: @footer-margin;
@@ -125,7 +137,8 @@ footer {
   .generate-img(@n, (@i + 1));
 }
 
-.octicon.icon {
+.octicon.icon,
+.mega-octicon.icon {
 	font-family: octicons;
 }
 
--- a/public/less/_repository.less
+++ b/public/less/_repository.less
@@ -3,32 +3,16 @@
 
 	padding-top: 15px;
 	padding-bottom: @footer-margin * 3;
+
 	.head {
-		height: 40px;
-		background-color: #FCFCFC;
-		.mega-octicon {
-			width: @mega-octicon-width;
-			font-size: 30px;
+		.ui.compact.menu {
+			margin-left: 1rem;
 		}
-		a,
-		.fork-flag {
-			font-weight: 300;
-		}
-		.ui.label {
-			margin-top: 5px;
-			vertical-align: top;
-		}
-		.fork-flag {
-			margin-left: @mega-octicon-width + 8px;
-			display: block;
-			font-size: 11px;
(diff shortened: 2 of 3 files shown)


Unknwon:
Thanks your PR!

But it looks really strange to me with this style of number...

manfer:
I have updated the PR with this new version

newrepoheader

Unknwon:
Thanks your PR!

I've made some comments on the code.

manfer:
I'm going to change this again as I found something that can be done better. Now I'm using a two column grid one for the header and another for the buttons (watch, star and fork). But It is better to use just one column.

I will reintroduce the .head class and remove all no needed CSS rules.

manfer:
ups, I did a force push without my commit and the PR was closed. Let's see if the PR can be reopened or I will have to PR again.

manfer:
Reopening worked.

Just in case I would like to point something out again. The PR doesn't include and updated version of `gogs.min.css` because it will conflict with changes already made on develop branch that, if I'm not wrong, will be difficult to merge. So to test this PR it must be checked out and the `less` files must be compiled and minified to generate the updated `public/css/gogs.min.css` file.

Unknwon:
Thanks, merge now.

Unknwon:
I‘m OK with right part, but the left part, the new design really doesn't feel right for me...

manfer:
I prefer the `huge breadcrumb` default font-size and weight but that's a matter of taste.

I bet you have a monitor with large resolution because your design looks in general large in 1366x1024.

That probably should be handle with CSS media queries to have different starting font sizes for different resolutions if Semantic-UI is not already doing that.

If I'm right and you have a large resolution monitor and Semantic-UI is doing nothing about that (and I'm almost sure it isn't), for such a device these rules would be needed:
```css
html {
  font-size: 16px;
}

body {
  font-size: 15px;
}
```
Not sure if 16/15 would be enough or you would need even larger starting sizes 18/17. But I suppose 16/15 is enough because your design is not too huge. Don't know if there are very very huge resolution monitors that would need higher starting font-sizes.

That should be the starting point to design for the large monitor in a way that it looks later the same on a smaller resolution.

Semantic-UI has `14px` on the html tag and `13px` on the body. They use different one on both tags because I suppose they want the `rem` relative unit to start at a larger size. I mimic that in my suggested starting font-size for large monitors.

The header I designed is already too huge for my taste on 1366x1024 though I have not been able with Semantic-UI classes to make it smaller. So if my header is looking small to you I'm sure you have a large resolution monitor.

Maybe I'm totally wrong but I think this is what is happening.

Just out of curiosity if you look at github body tag CSS font how it reads? And does github looks small in your monitor or just fine? I'm curious to know if github is using media queries or what they are doing about large resolution monitors.

This is how that font in github body looks for me and github page looks nice in my monitor resolution.
```css
body {
  font: 13px/1.4 Helvetica, arial, nimbussansl, liberationsans, freesans, clean, sans-serif, "Segoe UI Emoji", "Segoe UI Symbol";
}
```
Maybe I'm totally wrong about what is happening.

PS: In last [Semantic-UI version 2.0.8](https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.0.8/semantic.css) both html and body have a 14px font-size, so the developer must have changed mind in some moment. This means custom CSS needed for very large monitors would be just:
```css
html, body { font-size: 16px; }
```

PS2: In my opinion would be a good idea at least while we are starting to port everything to Semantic-UI to use a not minified version. Would be a lot easier to debug.

PS3: An update to v2.0.8 which should break nothing would be great.

Unknwon:
Yes, my screen is 1680*1050. 

> PS: In last Semantic-UI version 2.0.8 both html and body have a 14px font-size, so the developer must have changed mind in some moment. This means custom CSS needed for very large monitors would be just:

It is true for all 2.x versions.

Issue #1462, opened 2015-08-10
Error when entering the Admin config path


uriel-fanelli:
Hi, I found a strange message error when entering the /admin/config path. Instead of returning the config dump, after the last update it returns:

<pre>
template: admin/config:105:50: executing "admin/config" at <.Webhook.QueueLength>: QueueLength is not a field of struct type interface {}
</pre>

version is Version: 0.6.4.0809 Beta
Go version is 1.4.2 

Is is a bug?

Unknwon:
It looks like you updated Gogs but didn't recompile it...

uriel-fanelli:
What you mean? I rebuild it from the source code after a git clone...

Unknwon:
The missing field in your error are newly added, so if your template files are up-to-date but Gogs program didn't recompile or you didn't notice you are executing the old/wrong file, this type of errors could occur.

uriel-fanelli:
Again. I deleted the complete folder. I git-cloned again from github. I compiled from scratch. Again, the error is there.

It is physically impossible there is a mistake like "old binaries"

Unknwon:
What is the version shows at the bottom of the page in other pages?

uriel-fanelli:
© 2015 Gogs · Version: 0.6.4.0809 Beta ·

You can see it at boseburo.ddns.net , just ignore the certificate error, it is selfsigned by me.

Unknwon:
> Again. I deleted the complete folder. I git-cloned again from github. I compiled from scratch.

Would you mind to provide some step infos(commands)?

uriel-fanelli:
of course. 

I did a 

git clone https://github.com/gogits/gogs.git  

which created the folder gogs

then I enter gogs

go clean all
go get
go build

and then i get a brand new executable. 

Then I copy the custom config folder I have saved before, and 

./gogs web

please notice it works. It only has several issues. Btw, the SSH problem happens only for the account which used to have a ssh key saved. I created a new one, and this doesn't happens.

Where exactly gogs stores the keys?

Unknwon:
Very strange, I just checked out and tested that everything works fine. :sweat:

> Where exactly gogs stores the keys?

Currently, same file as other SSH keys `~/.ssh/authorize_keys`

uriel-fanelli:
I check, and the old key is there. Also I asked gogs to rewrite the file with keys, which has worked, and still the keys are there. 

[...]

PR #1467, opened 2015-08-11
#1462 , should be .Id and not .ID

uriel-fanelli:
see issue #1462 .


Unknwon:
Continue comment from #1462.

If you take a look at https://github.com/gogits/gogs/blob/master/models%2Fpublickey.go#L89 , you can see that actual field name is `ID` not `Id`(this is the old one)

uriel-fanelli:
I can confirm my code is the same. Still, with .ID it is not working, putting .Id it works.

I check the code in general, and I see a jeopardy of .Id and .ID, so I can't say which one is used.

Unknwon:
Close by #1462

PR #1469, opened 2015-08-11
fix dashboard padding

sorl:
Dashboard padding is 0 in left right and it would look alot nicer if they were 1.5em as the header, 1.5em is also a nice replacement for 24px top and bottom.


Unknwon:
Thanks!

Unknwon:
PS: Gogs uses `less` file to compile `css`, be sure to change `less` file next time! :beer:

sorl:
doh, of course, sry bout that

PR #1471, opened 2015-08-12
column belongs always into a ui grid

manfer:
Looks like a lot of changes but the only change is a new `.ui.grid` layer that contains the `.column` layer.

Because columns always belong to a `ui grid` and we forgot to include that layer.

All looks exactly the same after this change though there are some new margins and paddings that change nothing because they collapse in the same way than before the change.

Anyway for completeness this is needed and could lead to strange, not easy to debug, CSS problems if in future the semantic theme is changed.

### DIFF
--- a/templates/base/head.tmpl
+++ b/templates/base/head.tmpl
@@ -52,106 +52,108 @@
 
 		{{if not .PageIsInstall}}
 		<div class="following bar light">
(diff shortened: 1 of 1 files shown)


Unknwon:
Make sense, thanks!

Issue #1009, opened 2015-03-06
Config option for preserving hard line breaks


Unknwon:
By @jcracknell at https://github.com/gogits/gogs/issues/345#issuecomment-77509948

@Unknwon I am suggesting you can make the line break behavior user and repository-configurable, which you can then toggle at [modules/base/markdown.go#L211](https://github.com/gogits/gogs/blob/18c06973292a3d20b8ad72575c819fbe3287a9ad/modules/base/markdown.go#L211).

```go
type User struct {
  Id int64
  // ...
  Etc string `xorm:"TEXT"` // { "commentOptions": { 
                           //     "markdown": { "preserveLineBreaks": true }
                           //   },
                           //   "repositoryOptions": {
                           //     "markdown": { "preserveLineBreaks": false }
                           //   }
                           // }
}
```

```go
type Comment struct {
  Id int64
  Type CommentType
  PosterId int64
  // ...
  Etc string `xorm:"TEXT"` // { "markdown": { "preserveLineBreaks": true } }
}
```
```go
type Repository struct {
  Id int64
  OwnerId int64 `xorm:"UNIQUE(s)"`
  // ...
  Etc string `xorm:"TEXT"` // { "markdown": { "preserveLineBreaks": false } }
}
```
A column storing a blob of JSON would be convenient for storing this and other non-essential data without over-complicating future migrations.

matze:
Why all the extra complexity? I haven't found a place besides Gogs that turned on automatic hard linebreaks when rendering Markdown content. This is especially true since Markdown [allows](http://daringfireball.net/projects/markdown/syntax#p) hard linebreaks by appending two or more spaces at the end of the line. Just remove [this](https://github.com/gogits/gogs/blob/e0036ebb0cdcafede2670d120f911290815b5c57/modules/base/markdown.go#L201) pesky line and everyone will be happy and not surprised.

fiorix:
+1 for @matze's suggestion. The current hard line break is annoying.

jcracknell:
Well that was the general motivation, but the impression I had is that there are misguided persons who actually like hard line breaks. ;)

fiorix:
fwiw neither GitHub or Gitlab have hard line breaks in their markdown renderer.

jcracknell:
I believe the exception is
GitHub comments. ;)

Unknwon:
How about a repository level hard line breaks setting? And default is **false**.

jcracknell:
[...]

PR #1475, opened 2015-08-12
Fix race condition on ssh key delete, and revert re-revert of hard line break

donbowman:
https://github.com/gogits/gogs/commit/c5be40decbfdb488b79bb4f53bf9d238fae20f0c re-added the hard line breaks to the markdown, this is undesirable (it was just fixed in previous pull request).

Also, there is a race condition in the ssh keys... on deleting one key, between writing the temp file and the rename, a new key can be added, and this will delete it.

I don't believe this fix is entirely correct, there is a problem still. one problem is that a user can have 2 keys w/ the same fingerprint, and the system will pick one (e.g. key-####) and then lookup (key-YYYY) and get an error that they don't own it.

I think the correct fix is to always rewrite the whole file from db (which code exists but is currently manual). But i didn't put that in the pull.


models/publickey.go
```diff
@@ -386,8 +386,6 @@ func ListPublicKeys(uid int64) ([]*PublicKey, error) {
 
 // rewriteAuthorizedKeys finds and deletes corresponding line in authorized_keys file.
 func rewriteAuthorizedKeys(key *PublicKey, p, tmpP string) error {
-	sshOpLocker.Lock()
-	defer sshOpLocker.Unlock()
```
Unknwon: This empty line should be removed.

Unknwon:
Thanks your PR!

I've made some comments to help your PR be better.

Besides,

These [two lines](https://github.com/donbowman/gogs/blob/master/models/publickey.go#L267) should be removed, and add them to two other places:

- https://github.com/donbowman/gogs/blob/master/models/publickey.go#L337
- https://github.com/donbowman/gogs/blob/master/models/publickey.go#L588

donbowman:
i'm not sure i understand. the lock is used to control concurrent access to ~/.ssh/authorized_keys, why would the lock move to those spots?

Unknwon:
OK, I see. My bad :flushed:

Just remove that empty line should be fine.

Unknwon:
ping

donbowman:
blank line removed as per req

Unknwon:
Thank you very much!

manfer:
```
	extensions |= blackfriday.EXTENSION_HARD_LINE_BREAK
```
Had been removed with this merge.

There is no problem to have that extension enabled or disabled. People will have to adapt their markdown to one situation or the other. But blackfriday with hard line breaks ON renders markdown more similar to what people are used to which is github flavoured markdown.

It is ok to remove it but given a reason would be the very least. This is undesirable doesn't look to me a  good reason. Undesirable for whom or for what reason exactly? This was fixed in previous PR is not accurate. This is not a fix of anything because it is not a bug having it turned on. At most it would be an enhancement if it is considered that turning hard line break off is an enhancement.

Maybe the best solution, if possible, is to add an option to the [config](http://gogs.io/docs/advanced/configuration_cheat_sheet.html) or in the admin settings so anyway can choose to enable or disable hard line breaks at will.

tboerger:
I'm still searching for a reason why it have been removed

Unknwon:
I would really like to add a repo-level settings as someone else has proposed on another thread.

manfer:
> I would really like to add a repo-level settings as someone else has proposed on another thread.

That's even more fine grained option though I would include a way to config the default value in global config or admin settings anyway.

Unknwon:
@manfer per user maybe make more sense since different users have different tastes.

Unknwon:
@manfer but yes, global config is easier to accomplish, would be my final choice.

manfer:
> @manfer per user maybe make more sense since different users have different tastes.

They are not mutually exclusive. First the admin, the owner of gogs installation, sets the default value, ON or OFF. Then users have the option to decide to set it ON or OFF at will for themselves.

The first should be very easy to implement. Just an if based on the setting should be enough.

The user option is a larger task and can be implemented later if needed.

Unknwon:
OK, let move on to https://github.com/gogits/gogs/issues/1009

Issue #1429, opened 2015-08-01
LDAP : connect to an OpenLDAP (yunohost)


dolanor:
Hi,
I try to activate an LDAP authentication that is configured by the yunohost team.
The schema is as followed (from ldapvi)
```
11 uid=dolanor,ou=users,dc=yunohost,dc=org
uid: dolanor
objectClass: mailAccount
objectClass: inetOrgPerson
objectClass: posixAccount
uidNumber: 1377
displayName: Dolanor Tharivae
gidNumber: 1377
homeDirectory: /home/dolanor
loginShell: /bin/bash
givenName: Dolanor
cn: Dolanor Tharivae
maildrop: dolanor
sn: Tharivae
mail: dolanorgit@mydomain.org
mail: dolanor2@mydomain.org
```

What I tried was :
```
Authorization name : mydomain.org
Domain: mydomain.org
Host: IP
Port: 389
Base DN: ou=users,dc=yunohost,dc=org
Username attribute: uid
First name attribute: givenName
Surname attribute: sn
E-mail attribute: mail
Search Filter: (&(objectClass=mailAccount)(uid=%s))
Ms Ad SA: uid=%s,ou=users,dc=yunohost,dc=org
```

It doesn't work, and I have no idea where it fails. No log, nothing.
I tried to `go test` in `modules/ldap` by replacing some info with my ldap server and the test seem to work. But it doesn't work for the login.

Also, the screenshots from http://gogs.io/docs/features/authentication.html should be updated with the new Authentication screen.

Unknwon:
Thanks your feedback!

What behavior of Gogs did you see when `it fails`?

dolanor:
It just says "username or password is not correct". 
I see nothing in the logs. Even though I tried to activate the `Debug` or `Trace` version in the `/data/gogs/conf/app.ini` for all log Modes. Because I want to know if it fails at the bind, the connect or if my filters are not good.

From what I understand, I should try to connect just with my uid, and not with the mail (which is uid@mydomain.org).

By the way, I have plenty of email address on this account which are just alias that redirect to my main one. Do you handle that case too ? (but I think it is not the problem here. It is just an `off topic` question)

Unknwon:
Thanks your info!

Not very familiar with LDAP(but trying hard), so it takes time to debug on the problem.

BTW,

> Even though I tried to activate the Debug or Trace version in the /data/gogs/conf/app.ini for all log Modes.

[...]

Issue #1257, opened 2015-05-08
LDAP Base DN and Search Filter Has No Effect on User Authorization


msimpson:
The [**LoginUserLdapSource**](https://github.com/gogits/gogs/blob/master/models/login.go#L229) method relies upon [**SearchEntry**](https://github.com/gogits/gogs/blob/18c06973292a3d20b8ad72575c819fbe3287a9ad/modules/auth/ldap/ldap.go#L58) to perform an authentication check during the LDAP login process. This happens by way of the **SearchEntry** method binding to the domain with the given user's credentials, thereby providing authentication, and applying both the configured *base dn* and *search filter* to query the user's attributes for the possible use in auto registration.

However, if an administrator (such as myself) attempts to restrict those who could feasibly login by way of altering the *search filter* to a specific group, like the example given below:

```
(&(objectClass=user)(memberOf=cn=developers,cn=users,dc=domain,dc=com)(sAMAccountName=%s))
```

Then the **SearchEntry** method could end up returning a successful authentication but with blank attributes (`"", "", "", "", true`) for those users who successfully bind but ultimately fall outside of the *search filter*, or even *base dn*. And if this is the initial login for that user as well, then the profile which is created in turn is also blank (see below). This causes a situation where the user can never login a second time once they have logged out of the first.

![bad](https://cloud.githubusercontent.com/assets/730334/7537507/75e886c8-f567-11e4-951e-6cd7d4097938.png)

A quick fix would simply entail checking more of the returned attributes from the **SearchEntry** query, such as ***name***, instead of relying on the bind itself as the sole method of authentication and authorization.

Although, a better alternative would be to rewrite the method by which Gogs users are authenticated and authorized against LDAP entirely. Possibly providing a bind user in the LDAP configuration for all queries during auto registration would allow for an empty return from the **SearchEntry** query to be treated as "user not found" and hence unauthorized, instead. Whereas a successful query would be treated as an authorization and move on to a test bind for the given user's credentials to prove authentication as well.

**The main point being that a successful bind is only authentication and not authorization as well.**

#### Information:
**Gogs Version:** 0.6.1.0325 Beta
**Git Version:** 2.4.0
**Gogs Host System:** Arch Linux (lts)
**LDAP:** Windows Server 2008 R2 Active Directory

Unknwon:
[...]

Issue #1238, opened 2015-04-26
Crash when connect to LDAP with SSL


xujinzheng:
linux 14.04.2 amd64 release v0.6.1

panic 内容如下

[Macaron] Started POST /user/login for 123.118.11.224
panic: runtime error: index out of range

goroutine 1938 [running]:
github.com/gogits/gogs/modules/ldap.(*Conn).reader(0xc208bd5600)
	/home/wuwen/go/src/github.com/gogits/gogs/modules/ldap/conn.go:267 +0x30f
created by github.com/gogits/gogs/modules/ldap.(*Conn).start
	/home/wuwen/go/src/github.com/gogits/gogs/modules/ldap/conn.go:83 +0x32

goroutine 1 [IO wait]:
net.(*pollDesc).Wait(0xc20816de90, 0x72, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:84 +0x47
net.(*pollDesc).WaitRead(0xc20816de90, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:89 +0x43
net.(*netFD).accept(0xc20816de30, 0x0, 0x7f28c10427c0, 0xc2083e83e0)
	/usr/local/go/src/net/fd_unix.go:419 +0x40b
net.(*TCPListener).AcceptTCP(0xc2084bbc40, 0x6e2b2e, 0x0, 0x0)
	/usr/local/go/src/net/tcpsock_posix.go:234 +0x4e
net/http.tcpKeepAliveListener.Accept(0xc2084bbc40, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/http/server.go:1976 +0x4c
net/http.(*Server).Serve(0xc20886f860, 0x7f28c105a758, 0xc2084bbc40, 0x0, 0x0)
	/usr/local/go/src/net/http/server.go:1728 +0x92
net/http.(*Server).ListenAndServe(0xc20886f860, 0x0, 0x0)
	/usr/local/go/src/net/http/server.go:1718 +0x154
net/http.ListenAndServe(0xc208279c10, 0xc, 0x7f28c105a6f8, 0xc20806c0e0, 0x0, 0x0)
	/usr/local/go/src/net/http/server.go:1808 +0xba
github.com/gogits/gogs/cmd.runWeb(0xc2080e20e0)
	/home/wuwen/go/src/github.com/gogits/gogs/cmd/web.go:490 +0x21da
[...]

Issue #1145, opened 2015-03-25
TLS and BindDN


ArKam:
As discussed earlier on the LDAP documentation page of the project, it would be really cool to improve some functionalities on Gogs regarding the way it talk to LDAP Authentication Endpoint.

**1. Support of the BindDN property to allow more complex filters and LDAP requests against some secured LDAPs.**

**2. It seems that GO-LDAP support TLS Transport channel, both, ActiveDirectory AND OpenLDAP support LDAP over TLS which is the successor of Older SSL transport.**

As I pointed it out on the documentation page, when people talk about SSL support they mostly talk about TLS support now a day.

As I said, those two options are very important one when a SysAdmin benchmark future solutions for enterprise grade services (among many others).

flotho:
could it be a duplicate of #472 for the first point?

ArKam:
Absolutly, I was in direct talk with the author regarding this issue entry and so made it on github as requested without checking previous entry but from the screens and discussion on your link (I'm sorry I can't read Mandarin), the first is a duplicate request.

Gogist creator asked me to create this entry as I explained him clearly why and how works this request.

flotho:
Hy @ArKam 

Great thanks for your answer

ArKam:
You're welcome ;-)

smarmit:
+1 I believe we would be using gogs in our enterprise if we had the authenticated (bind) support for LDAP. Thx.

Unknwon:
Hi there, by mering PR #1476 into develop, this problem is claimed to be fixed, would you mind to test again and give us feedback? Thanks! :beer:

ArKam:
Hi @unknown, I can confirm that you can close this issue as the proposed merge is everything we needed ! The merger made a tremendous job on its patch and refactor of the ldap code, good job and I'm happy to hear back on this topic!

Unknwon:
@ArKam Thanks your confirmation!

PR #1476, opened 2015-08-13
Significantly enhanced LDAP support in Gogs

SergioBenitez:
Hi,

After a cluster of headaches working with LDAP in Gogs, I spent sometime revamping the Gogs LDAP auth module. Below is a brief summary (from the updated README) of the functionality in the new module:

This authentication module attempts to authorize and authenticate a user against an LDAP server. Like most LDAP authentication systems, this module does this in two steps. First, it queries the LDAP server using a Bind DN and searches for the user that is attempting to sign in. If the user is found, the module attempts to bind to the server using the user's supplied credentials. If this succeeds, the user has been authenticated, and the account information is retrieved and passed to the Gogs login infrastructure.

Here's an image of the new LDAP configuration form:
gogs-ldap

These changes should close the following issues:
#1257, #1238, #1145, #840, #472, #1429, #672 

This pull request also fixes several undocumented issues, such as a null pointer exception caused by https://github.com/gogits/gogs/blob/develop/models/login.go#L264, and adds logging to the LDAP authentication stream to aid in addressing LDAP configuration issues.

- Sergio


lunny:
Looks great!

Unknwon:
Thanks a lot!

But take a bit of time to review it.

stevenroose:
Nice, you are a hero!

SergioBenitez:
@Unknwon Let me know if you have any questions on the implementation or need me to rebase.

Unknwon:
@SergioBenitez no problem. I'll let you know :)

Unknwon:
PS: GitHub can merge with conflict on command line, so if no changes needed, I can still preform merge operation.

Unknwon:
@SergioBenitez Hi, this PR has been merged, all looks awesome just one question, how is the compatibility between the new config option of LDAP and old ones(current ones that people are using)

SergioBenitez:
@Unknwon They're not particularly compatible. Unless their LDAP server supports anonymous search requests and the filter they had set previously identified one user, the LDAP configuration will break. Since the previous filter was used in a different manner, this is unlikely to be the case, so I suspect most if not all LDAP users will need to reconfigure their Gog's LDAP parameters.

That being said, most users should find this form of LDAP configuration to be rather standard, and so reconfiguring their LDAP support should be rather easy. Furthermore, the README is rather detailed and should give even beginning LDAP users enough information to set this up.

Perhaps we should also update the LDAP information on the Gogs documentation using the image in the PR and the text in the README.

Unknwon:
> @Unknwon They're not particularly compatible. Unless their LDAP server supports anonymous search requests and the filter they had set previously identified one user, the LDAP configuration will break. Since the previous filter was used in a different manner, this is unlikely to be the case, so I suspect most if not all LDAP users will need to reconfigure their Gog's LDAP parameters.

Thanks for the info.

> Perhaps we should also update the LDAP information on the Gogs documentation using the image in the PR and the text in the README.

Yes, I'll add link on Gogs README to LDAP README.

Unknwon:
@SergioBenitez another note:

This PR will not be included in the next release(v0.6.5) which will be released in one day or two, because people need time to response on the issues that has been claimed to be fixed.

Hope you can understand! :blush:

ArKam:
Hi, I'm the requester for the #1145 issue and I can confirm that your patch is everything I was looking for ! Awesome and well executed job, your readme is wonderful and really understandable by novices ! Thanks for the share to the community !

Unknwon:
@SergioBenitez FYI, documentation has been updated: http://gogs.io/docs/features/authentication.html

SergioBenitez:
Awesome, thanks @Unknwon. I noticed a few issues with the style in the README. I've submitted PR #1492 which fixes them.

Unknwon:
@SergioBenitez I have updated gogs.io as well. :100:

dolanor:
Just to thank @SergioBenitez . It works for me. :100:

PR #1478, opened 2015-08-13
fixed chinese character to japanese KANJI

yaggytter:
fixed chinese character '语' to japanese KANJI '語' in language selection.


Unknwon:
Good catch! Thanks!

PR #1485, opened 2015-08-13
43bfee0d4872932c1da185e8367e21795bc4ceb5 reverted my change

donbowman:
423b1c69be7828c331a47467b66217d190e916b9, which in turn
causes all content to be served as text/plain. Let go decide
what the mime type is (e.g. text/html instead of text/plain).


fiorix:
+1

Makes no sense to force the content type to text/plain.

Unknwon:
LGTM. Merging

PR #1486, opened 2015-08-14
Fix Fail to load gopmfile: could not parse line: github.com/mcuadros/go-version

linquize:
Fix Fail to load gopmfile: could not parse line: github.com/mcuadros/go-version


Unknwon:
Thanks your PR!

But be sure PR to `develop` branch as documented: https://github.com/gogits/gogs#notices

linquize:
The problem is `gopm` / `go get` only fetches from `master` branch. So I have to PR to `master`

Unknwon:
It does not matter, I can push `develop` to `master`, but right now it generates a redundant merge commit when merge back to `develop` :joy:

PR #1504, opened 2015-08-18
Change input field to type=password for the authentication templates.

LydiaSevelt:
Title pretty much says it all, the input box for the password field wasn't set to "type=password" so passwords were being displayed in plaintext on the screen. The change is pretty trivial.

Oops, changing to develop branch.

### DIFF
--- a/templates/admin/auth/edit.tmpl
+++ b/templates/admin/auth/edit.tmpl
@@ -49,7 +49,7 @@
                                 </div>
                                 <div class="field">
                                     <label for="bind_password">{{.i18n.Tr "admin.auths.bind_password"}}</label>
-                                    <input class="ipt ipt-large ipt-radius {{if .Err_BindPassword}}ipt-error{{end}}" id="bind_password" name="bind_password" value="{{.Source.LDAP.BindPassword}}" />
+                                    <input class="ipt ipt-large ipt-radius {{if .Err_BindPassword}}ipt-error{{end}}" id="bind_password" name="bind_password" type="password" value="{{.Source.LDAP.BindPassword}}" />
                                 </div>
                                 <div class="field">
                                     <label class="req" for="user_base">{{.i18n.Tr "admin.auths.user_base"}}</label>
--- a/templates/admin/auth/new.tmpl
+++ b/templates/admin/auth/new.tmpl
@@ -45,7 +45,7 @@
                                     </div>
                                     <div class="field">
                                         <label class="req" for="bind_password">{{.i18n.Tr "admin.auths.bind_password"}}</label>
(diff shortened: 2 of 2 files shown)


Unknwon:
Thanks again!

SergioBenitez:
The bind password field was quite purposefully not set as type="password". This is because the bind password field is stored in plain text. I feared that setting the field's type as "password" might give the impression that it's safe to use a sensitive password for the Bind DN when it's not. It would be a shame if users started using their LDAP admin as the Bind DN account because they believe they're password is stored in some secure fashion.

In light of this, I vote for this commit to be reversed.

Unknwon:
@SergioBenitez thanks your explanation! 

How about I add a help text says something like "this password is saved as plaintext in database"?

LydiaSevelt:
SergioBenitez, I see where you are coming from on that, however, to give you my opinion as a user when I first saw that my first reaction was "This is broken" and then after that "This isn't secure". I would recommend you go with Unknwon's suggestion (which is what I was going to recommend) as there are other problems with having the password displayed on the screen. In a corporate environment when others can readily see you screen this leads to an issue with actually doing the configuration, not to mention when you are building a setup to test (as I was) and you just use your employee account to get things working for testing (with the intention of creating a service account if it moves to production) it can be very disconcerting to have your password displayed prominently on the screen.

Maybe something like "Warning: This password is stored in plain text in the database, consider creating a permissions limited service account to use as your Bind DN."

Just my two cents on the issue.

Unknwon:
Thanks @LydiaSevelt 's comment.

> "Warning: This password is stored in plain text in the database, consider creating a permissions limited service account to use as your Bind DN."

:+1:

SergioBenitez:
@LydiaSevelt I'm very aware of the issue of showing the password on the screen. In fact, if you read my previous comment closely, I allude to this being a very real issue: "...while preventing the password from being shown on the screen."

Nonetheless, from a UI/UX perspective, I'd prefer a "show password" checkbox type of thing, but a warning message seems okay too. Where would the warning message go?

Unknwon:
@SergioBenitez understand the `input` element:

image

LydiaSevelt:
Ah, I overlooked that in your previous post Serigio, my apologies.

SergioBenitez:
@Unknwon That looks good. @LydiaSevelt No worries.

I suggest the following text, a slight modification of @LydiaSevelt's:
> Warning: This password is stored in plain text. Do not use a high privileged account.

Unknwon:
@SergioBenitez looks better :100:

Unknwon:
This warning has been added.

PR #1506, opened 2015-08-19
add newline at the end of file for each gitignore

Ralph-Wang:
`git ignore` will add new ignore items fail if the origin file have no newline at the end.

make sure all .gitignore have newline at the end of file.


Unknwon:
Good catch, thanks!

PR #1507, opened 2015-08-19
Set IsAdmin using LDAP

gramakri:
The IsAdmin flag is set based on whether the admin filter
returned any result. The admin filter is applied with the user dn
as the search root.

In the future, we should update IsAdmin as well on each login.
Alternately, we can have a periodic sync operation.


Unknwon:
Thanks your PR!

> In the future, we should update IsAdmin as well on each login.

It might not be ideal solution, because I think Gogs does not talk to LDAP server always, only when user session is expired, or preform operations that need password validation.

Unknwon:
Looks all good.

Issue #1514, opened 2015-08-20
Missing 'auth login' in communication with smtp server


haoyixin:
here is the log:
[.../runtime/asm_amd64.s:2232 goexit()] [E] Async sent email 0 succeed, not send emails: xxx@xxx.com, info: Subject: [test] this is a test(#6), send issue notify emails err: 530 5.7.1 Client was not authenticated
EVERYTHING got ok when by telnet.
SO, i think it's missing 'auth login' ...
here is the config:
[mailer]
ENABLED = true
HOST = mail.xxx.com:25
FROM = xxx@xxx.com
USER = xxx
PASSWD = xxx

Unknwon:
Thanks your feedback!

Are you using gmail?

haoyixin:
not gmail, it's Microsoft ESMTP MAIL Service in my company
yeah, no 'auth plain' or 'auth md5...', as there isn't a 'auth login' handler in go,
i've added it in https://github.com/gogits/gogs/pull/1516
AND, thanks, :-)

Unknwon:
OK!

So does the code in your PR is actually working and solved your problem?

haoyixin:
[T] Async sent email 1 succeed, sent emails: xxx@xxx.com
yeah, that's working fine.(i've suffered a boring meeting just now -_-''')

Unknwon:
@haoyixin haha, OK. Thanks for your confirmation!

Unknwon:
Closed by https://github.com/gogits/gogs/pull/1517

PR #1515, opened 2015-08-20
Added supported of 'AUTH LOGIN'

haoyixin:
As there isn't a 'AUTH LOGIN' handler, i had added it. 
This for issue#1514https://github.com/gogits/gogs/issues/1514


Unknwon:
What's going on with this PR? :scream:

PR #1519, opened 2015-08-21
Modify minimumKeySize of RSA to 1024 for old keys.

haoyixin:
A WAY to fit existing environment :-)


Unknwon:
Sounds reasonable! Thanks!

PR #1524, opened 2015-08-21
Supported for binding with username in LDAP login

haoyixin:
As our LDAP server doesn't support anonymous search and I'm not admin, we need bind as combination like 'domain\username' with login password.
SO, i think the most economic way is adding a placeholder in BindDN setting and leaving Password blank.
How about you :-)

### DIFF
--- a/modules/auth/ldap/ldap.go
+++ b/modules/auth/ldap/ldap.go
@@ -42,6 +42,9 @@ func (ls Ldapsource) FindUserDN(name, passwd string) (string, bool) {
 
 	log.Trace("Search for LDAP user: %s", name)
 	if ls.BindDN != "" {
+		var bd, bp string
+		// With palceholder in BindDN and no password,
+		// It will bind as combination define in BindDN with login password
 		if ls.BindPassword == "" {
 			bd = strings.Replace(ls.BindDN, "<username>", name, -1)
 			bp = passwd
--- a/modules/auth/ldap/ldap.go
+++ b/modules/auth/ldap/ldap.go
@@ -8,6 +8,7 @@ package ldap
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/gogits/gogs/modules/ldap"
 	"github.com/gogits/gogs/modules/log"
@@ -30,7 +31,7 @@ type Ldapsource struct {
 	Enabled          bool   // if this source is disabled
 }
 
-func (ls Ldapsource) FindUserDN(name string) (string, bool) {
+func (ls Ldapsource) FindUserDN(name, passwd string) (string, bool) {
 	l, err := ldapDial(ls)
 	if err != nil {
 		log.Error(4, "LDAP Connect error, %s:%v", ls.Host, err)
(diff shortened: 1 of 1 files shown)


haoyixin:
SO, here is the log, it works well.
2015/08/21 17:05:53 [T] Search for LDAP user: xxx
2015/08/21 17:05:53 [T] Bound as BindDN xxx\xxx
2015/08/21 17:05:53 [T] Searching using filter (&(objectClass=user)(sAMAccountName=xxx))
2015/08/21 17:05:53 [T] Binding with userDN: CN=xxx,OU=xxx,OU=xxx,OU=xxx,DC=xxx,DC=xxx
2015/08/21 17:05:53 [T] Bound successfully with userDN: CN=xxx,OU=xxx,OU=xxx,OU=xxx,DC=xxx,DC=xxx

SergioBenitez:
I strongly dislike this approach for several reasons.

1) You're potentially leaking the user's password needlessly. If the LDAP server doesn't communicate over TLS an the user isn't in the LDAP database, then you'll be transmitting the user's password in the clear needlessly.

2) You've introduced a new way to replace the username string. The user filter field uses %s and here you've introduced the unconventional <username>.

3) This is essentially a very odd way of performing 'direct bind' authentication where a bind with the user is attempted to both authorize and authenticate. But, perhaps only some subset of users should be authorized to login, and that authorization is happening at the server - this method removes that ability.

In all, I believe the problem is not with Gogs, but with your LDAP setup. You should ask the administrator of your LDAP server to add a new DN with read access to the DNs of users who are allowed to login via Gogs or other services.

An alternative is to add a new form of authentication to Gogs, "Direct LDAP" or something of that sort, that doesn't use a Bind DN at all and binds directly with the user's username and password. Those with privacy concerns can use the current method and those without them or with restrictions such as yours can use the "Direct LDAP" method.

haoyixin:
About your '3 POINTS'，uhmm, that's a nice shooting, championship ring is yours, lol~
About POINT ONE:
OK, If the LDAP server doesn't communicate over TLS, everything is plain, no matter what you did or what I did.AND, protect the user neither in LDAP database nor goes' from leading password? Why? (I've often wondered， why can't I login Gmail with my Facebook account, -_-''')
About POINT TWO:
Sorry, I forgot it, it could be anything else NOT IN YOUR DISLIKES.
About POINT THREE:
The filter work well, by the log upside.
AND Something ELSE:
That's optional, purpose is decoupling, may odd but productive.(Our company is a tiny workshops with about 3000 employee, so ADD STH to LDAP server is a lengthy process.Enterprises must face up to this problem, especially in China.SO, I mean, this's promotional.)

Unknwon:
Thank you both @haoyixin @SergioBenitez 

Hmm... I'm not pro on LDAP so I can't comment about the concerns you're discussing, but I agree with @SergioBenitez 's opinion that this should be another type of LDAP(e.g. Direct LDAP) to distinguish from normal(current) version of LDAP in Gogs.

What do you guys think?

SergioBenitez:
@haoyixin My apologies if you took offense to my statements. I certainly didn't mean to offend you. The goal behind my comment was to explain how your proposed solution to the problem is slightly unconventional, which may confuse users, and to suggest an alternative. I understand that server administration may be out of your control, as seems to be the case here.

On point 1, I agree with you. This is why I explicitly used the word _needlessly_. If you use a Bind DN without TLS, the user's password will only be transmitted in the clear if the Bind DN can see that the user exists, not every time as the direct bind method would do.

On point 3, I wasn't saying that the filtering doesn't work in your example, I was saying that using a Bind DN gives you an additional server-side component to authorization. Since the server can be configured to allow the Bind DN to search/view only certain entries, it can be used as a method of server-side authorization.

@Unknwon After thinking more about the problem, that would be my preferred way to solve the issue @haoyixin is having. This would involve remove the Bind DN, Bind Password, and User Search Base fields from the LDAP form and adding a "User DN" field in their place that allows a '%s' parameter.

haoyixin:
@SergioBenitez SORRY for my radical comments :-) After careful consideration, I know it indeed to confuse the user. And I'm truly agree with the last suggestion in your last comment(Maybe, in UI, adding radio buttons in that form to choose this mode would be more clear for user @Unknwon )

samlcharreyron:
+1 for the Direct LDAP mode. I had to downgrade Gogs to an earlier version so I could continue using LDAP.

dasJ:
:+1: Would be very useful in my deployment.

SergioBenitez:
I'm working on this.

Unknwon:
@SergioBenitez hmm.... what is `this` are you working on ?

SergioBenitez:
Sorry, I should have been clearer. I'm working on implementing a new "direct LDAP" authentication method that functions as I described above.

Unknwon:
@SergioBenitez That is what I'm thinking you're trying to do, but needs your confirmation on it :smiley: so others can understand. 

Hmm, so is it OK to you? @haoyixin because this PR may result be closed without merging.

haoyixin:
ok, nop,

dasJ:
Tested it, lgtm

SergioBenitez:
See #1581.

PR #1536, opened 2015-08-24
Updated - Lang Selection Width

goldbattle:
Seems to have a small problem because of the long language `Português do Brasil`:

image image

This pushed down the rest, and ruined the 31px assumption. This just makes it wider so it doesn't run into that problem. Let me know if you run into any problems.

### DIFF
--- a/public/ng/less/gogs/base.less
+++ b/public/ng/less/gogs/base.less
@@ -101,7 +101,7 @@ clear: both;
         height: -3+31*@langNum;
         z-index: 100;
         font-size: 12px;
-        width: 120%;
+        width: 170%;
         min-width: 140px;
         li > a {
             padding: 3px 9px;


Unknwon:
Thanks! Looks good right now.

goldbattle:
For some reason I just checked the try.gogs.io and it doesn't seem to have
that problem with the 120%.
I am looking into why my version has problems.

My version is Version: 0.6.5.0815 Beta

Unknwon:
@goldbattle yes, I've noticed that. 

Also, I'll push compiled css for this less change with some other changes later together.

goldbattle:
Ah, seems that my version doesn't have the `min-width:140px` and instead has `min-width:100px`, I didn't see that was changed. 170% should be a width of 136px , while 120% is far below the width of 140px. My apologies.

PR #1581, opened 2015-09-05
Added LDAP simple auth. support

SergioBenitez:
As discussed in #1524, this pull request adds a simple authentication LDAP auth method to Gogs. 
This pull request also refactors the main login function in `models/login.go` and the auth form javascript code in `public/ng/js/gogs.js`.

From the README:
> LDAP simple authentication does not utilize a Bind DN. Instead, it binds directly with the LDAP server using the user's supplied credentials. If the bind succeeds and no filter rules out the user, the user is authenticated.

The simple auth LDAP form looks as follows:
screen shot 2015-09-04 at 9 20 59 pm


Unknwon:
Thanks!

SergioBenitez:
Checking in on this. Let me know if you need any input @Unknwon.

Unknwon:
Oops, didn't notice this has conflicts :mask:

Some other things are needed to be done before this, sorry about waiting!

And if you don't mind, would you like to update this PR and resolve conflict?

SergioBenitez:
Done.

Unknwon:
@SergioBenitez thank you, appreciate!

Unknwon:
Merged! Thanks again!

PR #1583, opened 2015-09-05
remove dashboard operations table header

manfer:
The operations section of dashboard has a double header. The panel header that reads "Operations" and a table header that reads "Operation Name   |   Switch". The table header is redundant and can be removed.


Unknwon:
OK, cool, thanks!

Although this page is going to be converted semantic UI shortly.

Unknwon:
But.... hmm, can you re-PR to `develop` branch?

manfer:
> But.... hmm, can you re-PR to develop branch?

https://github.com/gogits/gogs/pull/1584

PR #1589, opened 2015-09-06
Set IsBare true if migrated repo is empty

### LINKED ISSUE #1588: 500 after migrating repo from local bare repo
I migrated a repo using its local file path successfully, but when I go to the page I get a 500 error:

    [Macaron] Started GET /org/repo for 127.0.0.1
    2015/09/06 11:43:46 [...s/middleware/repo.go:361 func·009()] [E] GetBranches:
    [Macaron] Completed /org/repo 500 Internal Server Error in 16.84093ms

What could be the reason?

siadat:
Set `repo.IsBare` true if the migrated repository is empty.

Closes #1588

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -470,6 +470,12 @@ func MigrateRepository(u *User, name, desc string, private, mirror bool, url str
 		return repo, fmt.Errorf("create update hook: %v", err)
 	}
 
+	// Check if repository is empty.
+	_, stderr, err = com.ExecCmdDir(repoPath, "git", "log", "-1")
+	if err != nil && strings.Contains(stderr, "fatal: bad default revision 'HEAD'") {
+		repo.IsBare = true
+	}
+
 	// Check if repository has master branch, if so set it to default branch.
 	gitRepo, err := git.OpenRepository(repoPath)
 	if err != nil {


models/repo.go
```diff
@@ -470,6 +470,12 @@ func MigrateRepository(u *User, name, desc string, private, mirror bool, url str
 		return repo, fmt.Errorf("create update hook: %v", err)
 	}
 
+	// Check if repository is empty.
+	_, _, err = com.ExecCmdDir(repoPath, "git", "rev-parse", "--verify", "--quiet", "HEAD")
```
Unknwon: Will there be an error if it's not empty?

siadat: Yes, the exit status will be non-zero if the repository is empty!

Unknwon: Thanks for following up! And what about stderr? Will there just be an error and no stdout or stderr if it's empty?

Unknwon: How about use `git log` to test? 

It always complains error `fatal: bad default revision 'HEAD'` for empty repository, which more precise in case there is another real error happens.

siadat: stderr is silenced by `--quiet`. I running `git log` on a repository with thousands of commits might not be a good idea because it might take a long time. This approach only finds the commit that is referenced by `HEAD`, the exit status will be 0 (a commit found) or non-zero (no commit found).

Unknwon: Hmm... that's easy, just use `git log -1`, so it only finds latest 1 commit and stops.

Unknwon: OK, so if you don't use `--quite`, you can check error string for `fatal: Needed a single revision`.

siadat: So, we these alternatives:

    git log -1
    fatal: bad default revision 'HEAD'

    git rev-parse --verify HEAD
    fatal: Needed a single revision

If you think checking stderr is more reliable than checking err, I could update the commit to do something like this instead:

    _, stderr, err := com.ExecCmdDir(repoPath, "git", "log", "-1")
    if err != nil && strings.Contains(stderr, "fatal:") { ... }

Let me know what you think.

Unknwon:
```go
_, stderr, err := com.ExecCmdDir(repoPath, "git", "log", "-1")
if err != nil && strings.Contains(stderr, "fatal: bad default revision 'HEAD'") { ... }
```

would be prefect!

siadat: :+1:

siadat: @Unknwon commit updated!

Unknwon:
Great! Can you confirm that this PR actually fixes your problem?

siadat:
Yup, I just tested it again by migrating an empty repository, and it showed the expected bare.tmpl page instead of the old 500 page.

For empty repositories already migrated, they should either be deleted and re-migrated, or manually set their `is_bare = 1` in the database.

Unknwon:
Thank you very much!

siadat:
Thank you! :tada:

Issue #1528, opened 2015-08-24
Always use Mailer FROM setting when sending email


goldbattle:
Currently the mailer seems to error if the email section `FROM` in the config is supplied.
When the config line reads: `FROM = "Gogs" <noreply@****.com>` I get this error
```log
2015/08/23 21:47:11 [.../runtime/asm_amd64.s:2232 goexit()] [E] Async sent email 0 succeed, not send emails: noreply@****.com, info: Subject: [Test-Repo] Test(#1), send issue mention emails err: mail: missing phrase`
```

When I have this `FROM = noreply@****.com`, it still does not use this value and instead uses the email of the user that triggers the event.

I think it would be better if people want to functionality of being able to reply to the person that triggered the email event is to use the `FROM` address and then use the `Reply-To:` email header with the user's email.

goldbattle:
Tested with `0.6.3.0802 Beta` and `0.6.5.0815 Beta` according to the configuration page

Unknwon:
Thanks your feedback!

> I think it would be better if people want to functionality of being able to reply to the person that triggered the email event is to use the FROM address and then use the Reply-To: email header with the user's email.

I'm confused by this line. So you say ` it still does not use this value and instead uses the email of the user that triggers the event.`, and then you want to reply to the user who triggered the event(which is the way Gogs does right now).

goldbattle:
It does not use the `FROM` email as the email the email has been sent from.
As you said it sends it as the user sent the email, not the server that is sending it.

Unknwon:
OK... Gogs uses `FROM` you configured on `app.ini` when sends so-called `system` e-mails, and use user's e-mail as `FROM` for sending mention e-mails is on purpose that shows who actually mentioned you. 

Does my explanation of the behavior make sense to you?

goldbattle:
Yes!

That make sense.
I am looking for all messages to be sent from the server, and the ones that the user triggered instead use the `Reply-To:` flag. All emails are being mail from the server and not the person themselves, the current way makes it seem that the person sent that email, not the server. This causes issues as a smtp server is sending mail as a non-verified address that it doesn't own.

> If I use noreply@domain.com it should always send from that as it, as domain.com is the domain that I own and I have rights over, not gmail.com or whatever email domain that the user's email is hosted on.

The server should always send from the from address, but the user can still reply to the original sender using the `Reply-To:` header.

Unknwon:
OK, I understand now! 

[...]

PR #1592, opened 2015-09-07
add `FixedNotifyFrom` feature

denghongcai:
#1591


Unknwon:
Thanks.

But #1591 looks same as https://github.com/gogits/gogs/issues/1528 .

denghongcai:
确实是一样的，我重新提交一个PR上来

Unknwon:
For the new PR, I think you just need to use `[mailer] FROM` for everything.

PR #1609, opened 2015-09-09
Add allow attribute `class` to `<code>` tag

denghongcai:
PR for #1608 

我这里没有Mac环境没法用codekit，还需要修改highlight.js的配置为

```javascript
hljs.configure({
  classPrefix: 'language-'
})
```

### DIFF
--- a/modules/base/tool.go
+++ b/modules/base/tool.go
@@ -15,6 +15,7 @@ import (
 	"hash"
 	"html/template"
 	"math"
+	"regexp"
 	"strings"
 	"time"
 
@@ -26,11 +27,8 @@ import (
 	"github.com/gogits/gogs/modules/setting"
 )
 
-var Sanitizer = bluemonday.UGCPolicy()
+var Sanitizer = bluemonday.UGCPolicy().AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code")
 
-func init() {
-	Sanitizer.AllowAttrs("class").OnElements("code")
-}
 
 // Encode string to md5 hex value.
 func EncodeMd5(str string) string {
--- a/modules/base/tool.go
+++ b/modules/base/tool.go
@@ -28,6 +28,10 @@ import (
 
 var Sanitizer = bluemonday.UGCPolicy()
 
+func init() {
+	Sanitizer.AllowAttrs("class").OnElements("code")
+}
+
 // Encode string to md5 hex value.
 func EncodeMd5(str string) string {
 	m := md5.New()


manfer:
Don't know when this regression happened but there were already code to allow this.
https://github.com/gogits/gogs/pull/1442/files#diff-88e948dbecd728e2513ba9bafe876faaR30

I don't suggest to use this code but the one from the link I provide here. The code provided in this pull request is not safe.

manfer:
> Don't know when this regression
No regression the feature/highlight branch must be not merged

denghongcai:
I'm interested in the reason which will cause a `class` attribute exploit. Can you provide a link about it?

I think the best way to do language syntax highlight is to do it at the browser, not backend. It has more features and easy to modified. I also want to change the current way of rendering markdown to front-end way.

Maybe it's worth to make a discussion?

manfer:
If you allow the class to be anything you can just use something like this as value:

```
"></code>And here I can add anything that wont be sanitized<code class="
```

denghongcai:
input :
```
"></code><script>alert("test");</script><code class="
```

output (.AllowAttrs("class").OnElements("code")) :

```
&#34;&gt;</code>
```

output:

```
&#34;&gt;</code>
```

manfer:
I can't provide you with a working XSS example. But if bluemonday developers on their readme prevents you againts XSS attribute injection should be better in my opinion not to ignore them.

And I'm not sure what you have tested there. What you should test would be something like this.
```
<code class=""></code><script>alert("test");</script><code class=""></code>
```
Which is not invalid html at all. Probably would be sanitized correctly but not sure if it is the correct test to do.

I would take the advice of bluemonday developers and use a regex filter to only allow safe inputs. In this case as really the only valid input should be "language-something", it can be restricted even more than with the suggested bluemonday regex.

denghongcai:
Though I believe there is no exist way to exploit at `class` attribute, it based on bluemonday using a parser to analyze html code. I agreed that we'd better restrict the input format.

So there begins a new discussion, is making highlight better at browser-side?

Unknwon:
Thank you both!

But there is a confusion here for you @denghongcai :

The hilightjs is now only used in the new Semantic UI pages(right now, only one page is using it: the webhook history). So currently change its configuration does not help on anything.

> I also want to change the current way of rendering markdown to front-end way.

This is not likely going to happen unless the JS lib support context-related rendering, like auto-render issue, @ mention, commit ID, etc.

I agree with @manfer about using https://github.com/gogits/gogs/pull/1442/files#diff-88e948dbecd728e2513ba9bafe876faaR30 for allowing `code` class attribute.

manfer:
If the regex filter is added I would suggest merging this pull request as probably current prettyprint highlighter would be enhanced too if the code tags include the language class. I have not tested but it should work better. Not sure if something else would be needed but at most minor changes to the code that handles prettyprint.

denghongcai:
A huge confusion :sob: 

We can just use js lib to highlight code, because it's easy and easy to add more language support.

Thanks for @manfer , I will do some tests tomorrow and update progress here. Have a nice night.

Unknwon:
@manfer yes, for current stage, merging this PR is 100% OK with me, actually.

Unknwon:
@denghongcai we have not yet use highlightjs for code blocks, keep this PR as it is the the best thing right now.

Unknwon:
@denghongcai hmm... my bad, actually you use chain operation:

```go
bluemonday.UGCPolicy().Allowxxx...
```

denghongcai:
cleaned my code :smile: 

```
Gogs.renderMarkdown = function() {
    var $md = $('.markdown');
    var $pre = $md.find('pre > code').parent();
    $pre.addClass('prettyprint');
    prettyPrint();

    // Set anchor.
    var headers = {};
    $md.find('h1, h2, h3, h4, h5, h6').each(function() {
        var node = $(this);
        var val = encodeURIComponent(node.text().toLowerCase().replace(/[^\w\- ]/g, '').replace(/[ ]/g, '-'));
        var name = val;
        if (headers[val] > 0) {
            name = val + '-' + headers[val];
        }
        if (headers[val] == undefined) {
            headers[val] = 1;
        } else {
            headers[val] += 1;
        }
        node = node.wrap('<div id="' + name + '" class="anchor-wrap" ></div>');
        node.append('<a class="anchor" href="#' + name + '"><span class="octicon octicon-link"></span></a>');
    });
};
```

code-prettify was used here. I think it's easy to change it to highlight.js

another way is add some extensions to code-prettify, because the lib used here works on

>The comments in prettify.js are authoritative but the lexer should work on a number of languages including C and friends, Java, Python, Bash, SQL, HTML, XML, CSS, Javascript, Makefiles, and Rust.
It works passably on Ruby, PHP, VB, and Awk and a decent subset of Perl and Ruby, but, because of commenting conventions, doesn't work on Smalltalk, OCaml, etc. without a language extension.

https://github.com/google/code-prettify#for-which-languages-does-it-work

Unknwon:
Thanks again, merging...

PR #1618, opened 2015-09-11
Fix the deactive user can sign in bug.

ohko:
I find a bug, the user is deactived, but can sign in.
I fix it, may be the code don't right, you can modify it.

### DIFF
--- a/models/login.go
+++ b/models/login.go
@@ -193,6 +193,10 @@ func UserSignIn(uname, passwd string) (*User, error) {
 		if !u.ValidatePassword(passwd) {
 			return nil, ErrUserNotExist{u.Id, u.Name}
 		}
+		// If not active, return error
+		if !u.IsActive {
+			return nil, ErrUserNotExist{u.Id, u.Name}
+		}
 		return u, nil
 	}
 


Unknwon:
Thanks your PR!

1. What are the reproduce steps that can actually lead to a inactive user post anything to Gogs?
2. ... because it is designed to be able to login and always see inactive account notice at https://github.com/gogits/gogs/blob/master/modules/middleware/auth.go#L101
3. Please send all PRs to `develop` branch as documented here: https://github.com/gogits/gogs/blob/master/CONTRIBUTING.md#pull-request

ohko:
Sorry, I was a lot "github" novice, don't understand the rules.
I see what you mean, then you can add a "disabled" fields whether can sign in. Just advice.
The next step is I submit to the develop, or nothing to do?

Unknwon:
No worries.

Hmm... this is somehow became a bigger topic, if you simply disable login, then there is no way to resend activation e-mail by the user self.

ohko:
That user was disabled by administrator, so user don’t has access to gogs and active by email, only can enabled by administrator.

英文太差，还是中文吧，呵呵。
另外，你是怎么进行开发的呢？不会是修改部分程序然后build，测试有问题再修改，再build ，因为我发现build需要一点时间，这样岂不是很慢？

接着上面，我这个提交是关闭不管了，还是重新提交到develop?对github不熟悉。

Unknwon:
> That user was disabled by administrator, so user don’t has access to gogs and active by email, only can enabled by administrator.

Then there is a problem that how do you know the user is inactivate because he/she does not confirm with his/her e-mail or because he/she is disabled by admin?

I think we need another field in `user` table.

> 另外，你是怎么进行开发的呢？不会是修改部分程序然后build，测试有问题再修改，再build ，因为我发现build需要一点时间，这样岂不是很慢？

Yes, but I use automated recompile/reload tool: https://github.com/Unknwon/bra

ohko:
我恍然大悟，inactive在这里是用来标记email是否激活过，而我把这个字段理解为被禁用了。
看来是需要增加一个字段来标记账号是否限制登录。

Unknwon:
OK...

Now you can close this PR and open an issue for adding a new field about admin prohibits user login.

ohko:
add issue #1621.

PR #1629, opened 2015-09-12
PageIsExplore must be PageIsExploreRepositories

manfer:
https://github.com/gogits/gogs/commit/e43341f4c562800017a040926857aa9cad6e3f60#commitcomment-13195559

### DIFF
--- a/templates/base/head.tmpl
+++ b/templates/base/head.tmpl
@@ -50,7 +50,7 @@
 							<a class="item{{if .PageIsHome}} active{{end}}" href="{{AppSubUrl}}/">{{.i18n.Tr "home"}}</a>
 							{{end}}
 
-							<a class="item{{if .PageIsExplore}} active{{end}}" href="{{AppSubUrl}}/explore">{{.i18n.Tr "explore"}}</a>
+							<a class="item{{if .PageIsExploreRepositories}} active{{end}}" href="{{AppSubUrl}}/explore">{{.i18n.Tr "explore"}}</a>
 							<!-- <div class="item">
 								<div class="ui icon input">
 									<input class="searchbox" type="text" placeholder="{{.i18n.Tr "search_project"}}">


Unknwon:
Thanks your PR!

But the template is correct, the problem is missing this value in the router, have pushed fixed on it.

manfer:
The router was using PageIsExploreRepositories instead of PageIsExplore. So it could be fixed by changing it on the template or changing it on the router.

Not sure why you need two flags (PageIsExplore and PageIsExploreRepositories) to identify that page. So if you fix it by adding PageIsExplore to the router it looks like then the line https://github.com/gogits/gogs/blob/develop/routers/home.go#L54 is not needed anymore.

Unknwon:
> Not sure why you need two flags (PageIsExplore and PageIsExploreRepositories) to identify that page

There are two navbars, top and left...

manfer:
> There are two navbars, top and left...
Not sure if that makes a difference. If that function is executed as it is coded now both flags `PageIsExplore` and `PageIsExploreRepositories` are set to true. I can understand why you can't use only one.

Unknwon:
>  I can't understand why you can't use only one flag.

What if you have explore users feature now? What do you do?

manfer:
> What if you have explore users feature now? What do you do?
Then it is an option to be able to identify in the future the exact subsection of that Explore section. But as for now it doesn't exists because there is only one section. If I understand correctly. It is OK.

Unknwon:
> But as for now it doesn't exists because there is only one section

This is not a acceptable statement for features we already know to be added.

manfer:
> This is not a acceptable statement for features we already know to be added.

It is a totally acceptable statement because it just explains what's going on. :)

As already said in previous reply. It is OK.

I understand the logic now. And I agree about the need of both flags.

Unknwon:
Thanks for your understanding!

PR #1662, opened 2015-09-19
Commit messages now rendered with line breaks and prefix spacing.

SergioBenitez:
This commit improves the way that commit messages are rendered, especially in the diff view. First, it fixes the 'diff' template so that the 'browse source' button correctly floats to the right of the message. Second, it renders line breaks and prefix spacing in commit messages.

Previously, when a commit message appeared in the log as:

```
Change 1 did this. Change 2 did this.

These changes were very good. I'm pretty happy with them. Here they are:
    * Change 1
    * Change 2
```

Gogs would render it is as:

```
Change 1 did this. Change 2 did this. These changes were very good. I'm pretty happy with them. Here they are: * Change 1 * Change 2
```

With this commit, Gogs will render the message just as it was in the log.

### DIFF
--- a/modules/base/template.go
+++ b/modules/base/template.go
@@ -96,9 +96,42 @@ func ToUtf8(content string) string {
 	return res
 }
 
+// Replaces all prefixes 'old' in 's' with 'new'.
+func ReplaceLeft(s, old, new string) string {
+	old_len, new_len, i, n := len(old), len(new), 0, 0
+	for ; i < len(s) && strings.HasPrefix(s[i:], old); n += 1 {
+		i += old_len
+	}
+
+	// simple optimization
+	if n == 0 {
+		return s
+	}
+
+	// allocating space for the new string
+	newLen := n*new_len + len(s[i:])
+	replacement := make([]byte, newLen, newLen)
+
+	j := 0
+	for ; j < n*new_len; j += new_len {
+		copy(replacement[j:j+new_len], new)
+	}
+
+	copy(replacement[j:], s[i:])
+	return string(replacement)
+}
+
 // RenderCommitMessage renders commit message with XSS-safe and special links.
 func RenderCommitMessage(msg, urlPrefix string) template.HTML {
-	return template.HTML(string(RenderIssueIndexPattern([]byte(template.HTMLEscapeString(msg)), urlPrefix)))
+	cleanMsg := template.HTMLEscapeString(msg)
+	fullMessage := string(RenderIssueIndexPattern([]byte(cleanMsg), urlPrefix))
+	msgLines := strings.Split(strings.TrimSpace(fullMessage), "\n")
+	for i := range msgLines {
+		msgLines[i] = ReplaceLeft(msgLines[i], " ", "&nbsp;")
+	}
+
+	fullMessage = strings.Join(msgLines, "<br>")
+	return template.HTML(fullMessage)
 }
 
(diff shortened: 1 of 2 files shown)


modules/base/template.go
```diff
@@ -96,9 +96,42 @@ func ToUtf8(content string) string {
...
+
+	j := 0
+	for ; j < n*new_len; j += new_len {
+		copy(replacement[j:j+new_len], new)
+	}
+
+	copy(replacement[j:], s[i:])
+	return string(replacement)
+}
+
```
Unknwon: I simplified this function as follows:

```go
func ReplaceLeft2(s, old, new string) string {
	old_len, i, n := len(old), 0, 0
	for ; i < len(s) && strings.HasPrefix(s[i:], old); n += 1 {
		i += old_len
	}

	// simple optimization
	if n == 0 {
		return s
	}

	return strings.Repeat(new, n) + s[i:]
}
```

And did simple benchmark:

```go
func Benchmark_ReplaceLeft(b *testing.B) {
	for n := 0; n < b.N; n++ {
		ReplaceLeft(`Change 1 did this. Change 2 did this.

These changes were very good. I'm pretty happy with them. Here they are:
    * Change 1
    * Change 2`, " ", "&nbsp;")
	}
}

func Benchmark_ReplaceLeft2(b *testing.B) {
	for n := 0; n < b.N; n++ {
		ReplaceLeft2(`Change 1 did this. Change 2 did this.

These changes were very good. I'm pretty happy with them. Here they are:
    * Change 1
    * Change 2`, " ", "&nbsp;")
	}
}
```

The performance are pretty close but mine one has less code:

```
Benchmark_ReplaceLeft-8 	200000000	         8.37 ns/op
Benchmark_ReplaceLeft2-8	200000000	         8.08 ns/op
```

How do you think?

Unknwon:
Thanks your PR!

I've made some comments to help you improve it. :blush:

SergioBenitez:
Stylistically, I like your change, but your code is quite a bit slower. Unfortunately, your benchmark doesn't measure the functions at all, which is why each call takes only 8ns. This is because the string you've passed in to the call doesn't contain any `" "` characters as prefix, and so both implementations will simply hit the "simple optimization" case and do almost nothing. The .37 difference is likely an artifact.

I've concocted the following benchmark that shows that the original function is in fact 1.3x faster:

```go
func Benchmark_ReplaceLeft(b *testing.B) {
	testStrings := strings.Split(`                      dsfsdfds
         * Change 1
    * Change 2 and a half`, "\n")
	b.ResetTimer()

	for n := 0; n < b.N; n++ {
	  for i := 0; i < len(testStrings); i++ {
		ReplaceLeft(testStrings[i], " ", "&nbsp;")
	  }
	}
}

func Benchmark_ReplaceLeft2(b *testing.B) {
	testStrings := strings.Split(`                      dsfsdfds
         * Change 1
    * Change 2 and a half`, "\n")
	b.ResetTimer()

	for n := 0; n < b.N; n++ {
	  for i := 0; i < len(testStrings); i++ {
		ReplaceLeft2(testStrings[i], " ", "&nbsp;")
	  }
	}
}
```

```
Benchmark_ReplaceLeft	  500000	      2422 ns/op
Benchmark_ReplaceLeft2	  500000	      3162 ns/op
```

Unknwon:
Yes, you are right, let me think again!

Unknwon:
OK, merging for now, thanks again!

PR #1676, opened 2015-09-23
Fix #1669

### LINKED ISSUE #1669: service restart packager.io
HI,

Is it possible to do an automatic service restart at the end of the update on packager.io?

Have a nice day

preovaleo:
Restart gogs at the end of the update on Debian


Unknwon:
Thanks!

PR #1678, opened 2015-09-23
Fix `go get` not working when there is a username or repository name with some uppercase letters.

TheHowl:
Already described in the title. At the moment, if you do `go get` to a repo hosted on a gogs instance, you must use a lowercase username and repo name, otherwise it won't work, because gogs returns in the meta tag the username and the repo name in lowercase, and so `go get` finds two different names, and will just output to the user `package repo.url/username/reponame: unrecognized import path "repo.url/username/reponame"`.


Unknwon:
Thank you very much!

PR #1686, opened 2015-09-25
Fix #1669 on develop

### LINKED ISSUE #1669: service restart packager.io

preovaleo:
Fix #1669 in the develop branch
Restart the gogs service at the end of an update

### DIFF
--- a/packager/debian/postinst
+++ b/packager/debian/postinst
@@ -33,5 +33,5 @@ case "$1" in
     service gogs restart || true
     
   ;;
-  
+
 esac
--- a/packager/debian/postinst
+++ b/packager/debian/postinst
@@ -30,4 +30,6 @@ case "$1" in
     ${CLI} scale web=1 || true
   ;;
 
+service gogs restart || true
+
 esac
--- a/packager/debian/postinst
+++ b/packager/debian/postinst
@@ -28,8 +28,12 @@ case "$1" in
 
     # scale
     ${CLI} scale web=1 || true
+    
+    #restart the service
+    service gogs restart || true
+    
   ;;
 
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks again!

Based on https://github.com/gogits/gogs/issues/1669#issuecomment-142007106

@crohr suggests add this line to end of file?

packager/debian/postinst
```diff
@@ -30,4 +30,6 @@ case "$1" in
     ${CLI} scale web=1 || true
   ;;
 
+service gogs restart || true
+
```
TheHowl: I'm no bash expert, but why did you write this line out of all the other cases? I mean, shouldn't it go before the `;;` at line 31?

preovaleo:
Yes you're right solved

Unknwon:
cc @TheHowl @crohr please help review!

packager/debian/postinst
```diff
@@ -28,6 +28,12 @@ case "$1" in
 
     # scale
     ${CLI} scale web=1 || true
+    
+    #restart the service
```
TheHowl: I don't want to be picky here, but all the other comments are in the format `# <content>`. Why make this one `#<content>`?

packager/debian/postinst
```diff
@@ -28,6 +28,12 @@ case "$1" in
...
+    service gogs restart || true
+    
   ;;
 
+
+
```
TheHowl: And also, why unnecessary newlines?

Unknwon:
Thanks!

PR #1689, opened 2015-09-25
New admin notice UI based on user list example

sapk:
capture d ecran_2015-09-25_18-28-06

### DIFF
--- a/routers/admin/notice.go
+++ b/routers/admin/notice.go
@@ -6,11 +6,13 @@ package admin
 
 import (
 	"github.com/Unknwon/com"
+        "github.com/Unknwon/paginater"
 
 	"github.com/gogits/gogs/models"
 	"github.com/gogits/gogs/modules/base"
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/middleware"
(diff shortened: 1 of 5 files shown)


Unknwon:
Thanks and one little thing: total number can be showed after `system notice` title, like in users page.

sapk:
Ok i will correct it

routers/admin/notice.go
```diff
@@ -6,11 +6,13 @@ package admin
 
 import (
 	"github.com/Unknwon/com"
+        "github.com/Unknwon/paginater"
```
TheHowl: Indentation seems a bit fucked up here. Run `go fmt`.

Unknwon:
It is interesting that I keep getting build fail notices where here is passed: https://travis-ci.org/gogits/gogs/builds/82186221 :joy:

Unknwon:
Looks good, merging!

sapk:
Maybe you receive fail from the travis running on my fork (since it show in gitter) : https://travis-ci.org/sapk/gogs/jobs/82186216 since it pull (go get) from my master that is outdated. Maybe ?

Unknwon:
@sapk possible.

PR #1699, opened 2015-09-27
Set USER env variable so installation can complete successfully.

geerlingguy:
In testing startup on Ubuntu/Debian, it seems the `USER` environment variable is not present when Gogs is running using the default init script.

There may be a better way to fix this and make sure start-stop-daemon preserves the preconfigured environment, but I was at least able to complete installation (instead of getting an annoying `Run user isn't the current user: git ->` message (note the empty string for the 'CurrentUsername()' as picked up by the environment).

### DIFF
--- a/scripts/init/debian/gogs
+++ b/scripts/init/debian/gogs
@@ -48,11 +48,11 @@ do_start()
 	#   0 if daemon has been started
 	#   1 if daemon was already running
 	#   2 if daemon could not be started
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR!

But please re-PR to `develop` branch.

geerlingguy:
Created https://github.com/gogits/gogs/pull/1710 — sorry about that! I wish GitHub had a way to change branches in the same PR :/

Issue #1499, opened 2015-08-18
Use ctrl+c when clone through SSH makes Gogs docker exit unexpectedly


masiqi:
client:
 git clone ssh://git@git.xxx.xxxx:10022/xxxx/abcd.git
Cloning into 'abcd'...
remote: Counting objects: 12138, done.
remote: Compressing objects: 100% (5245/5245), done.
^CKilled by signal 2.3% (365/12138)   
fatal: index-pack failed
fatal: The remote end hung up unexpectedly

docker gogs:
root@iZ23j3nxjzkZ:/data1/gogs/gogs# docker run  -p 192.168.42.1:3000:3000 -p 10022:22 -v /data1/gogs:/data gogs/gogs
Starting OpenBSD Secure Shell server: sshdCould not load host key: /data/ssh/ssh_host_ed25519_key
.
sending incremental file list

sent 2750 bytes  received 40 bytes  5580.00 bytes/sec
total size is 353972  speedup is 126.87
2015/08/18 04:38:16 [T] Custom path: /data/gogs
2015/08/18 04:38:16 [T] Log path: /gopath/src/github.com/gogits/gogs/log
2015/08/18 04:38:16 [I] Gogs: Go Git Service 0.6.5.0817 Beta
2015/08/18 04:38:16 [I] Log Mode: File(Info)
2015/08/18 04:38:16 [I] Redis Supported
2015/08/18 04:38:16 [I] Memcache Supported
2015/08/18 04:38:16 [I] Cache Service Enabled
2015/08/18 04:38:16 [I] Session Service Enabled
2015/08/18 04:38:16 [I] Mail Service Enabled

2015/08/18 04:38:17 [I] SQLite3 Supported
2015/08/18 04:38:17 [I] Run Mode: Production
2015/08/18 04:38:18 [I] Listen: http://0.0.0.0:3000

root@iZ23j3nxjzkZ:/data1/gogs/gogs# 

类似这样，自己就退出了。。我怎么才能知道是什么问题？怎样才能避免？谢谢

masiqi:
补充一下，直接docker pull gogs/gogs的

Unknwon:
Thanks your feedback!

Check `/gopath/src/github.com/gogits/gogs/log/gogs.log` see if any error presents there.

masiqi:
没有什么信息。。。
root@iZ23j3nxjzkZ:/tmp/ttt# tail -f gogs.log 
[...]

Issue #1708, opened 2015-09-28
Add VOLUME to Dockerfile


freemountain:
Please add ```VOLUME /data``` into the Dockerfile.
So you could use ```--volumes-from``` to link containers.

Thank you in advance

Unknwon:
Thanks your feedback!

But I think this is basically same as https://github.com/gogits/gogs/issues/1631 and please read the docs here: https://github.com/gogits/gogs/tree/master/docker

freemountain:
Thank your for the quick response.
I read the docs but i didnt see the other issue.
The docs describe how to mount a host directory as a data volume.
This is possible but i think using data volume containers is a better (and more 'dockerish') approach.
Adding ```VOLUME /data``` doesent prevent someone from adding other volumes using the ```-v```flag.

Unknwon:
> Adding VOLUME /data doesent prevent someone from adding other volumes using the -vflag.

But why would you need two volumes for same purpose, as explained in #1631, the Dockerfile tries to leave open choices for volume.

freemountain:
I just wanted to clarify, that even if you add the volume directive to the dockerfile, you leave open choices for volumes.
It is a way to tell docker where you need persistence in your container, so you could use ```--volumes-from```like this:
```
run --name gogs-data --entrypoint /bin/true gogs/gogs 
run --name=gogs --volumes-from gogs-data gogs/gogs
```

Unknwon:
Thanks your explanation! 

I'm kind of confused by dockerfile command `VOLUME /data`, what does it supposed to mean? Does it only for giving the ability to mount to another data container?

freemountain:
http://container-solutions.com/understanding-volumes-docker:
>In order to understand what a Docker volume is, we first need to be clear about how the filesystem normally works in Docker. Docker images are stored as series of read-only layers. When we start a container, Docker takes the read-only image and adds a read-write layer on top. If the running container modifies an existing file, the file is copied out of the underlying read-only layer and into the top-most read-write layer where the changes are applied. The version in the read-write layer hides the underlying file, but does not destroy it — it still exists in the underlying image. When a Docker container is deleted, relaunching the image will start a fresh container without any of the changes made in the previously running container — those changes are lost. Docker calls this combination of read-only layers with a read-write layer on top a Union File System.

https://docs.docker.com/reference/builder/#volume:
> The VOLUME instruction creates a mount point with the specified name and marks it as holding externally mounted volumes from native host or other containers.

[...]

Issue #1313, opened 2015-06-05
error adding ed25519 key


kyrias:
I get the following error when trying to add ed25519 key to try.gogs.io:

``Sorry, we're not able to verify your SSH key: ssh-keygen -l -f: line 1 too long: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIFa2...``

Unknwon:
How long is a ed25519 key? :scream:

kyrias:
@Unknwon An ed25519 public key is around 80 chars + the comment,

Unknwon:
Well... that shouldn't be long...

But what do you mean by the 

> the comment

kyrias:
The same comment that every SSH key has, most often being the hostname or username + hostname.

Unknwon:
Thanks your info!

Unknwon:
I've did some researching that I can't find out why Gogs gives you this error.

So... would you mind save your key content into a file, and do `ssh-keygen -l -f filename` see what system returns you?

kyrias:
```
256 SHA256:KImNN87f1lzuLGHkm4hXrlIgJwxjF9OqzY8T8uDywjg kyrias@zorg.kyriasis.com (ED25519)
```

like it should.

Unknwon:
Well... thanks your info. Let me generate a ED25519 key and add to Gogs.

Unknwon:
Hmm...facing another problem myself with `unknown key type ed25519`.

:joy:

arnehormann:
@Unknwon you need at least [OpenSSH version 6.5](http://www.openssh.com/txt/release-6.5), ed25519 is not supported in older versions.

Unknwon:
@arnehormann thanks for the tip!

cc @kyrias can you check on that?

arnehormann:
I should probably add that I use gogs with a ed25519 key. Works for me! :grinning:

Unknwon:
@arnehormann :eyes:

arnehormann:
@Unknwon does that mean you'd like a screenshot or you want to see the configuration?
I think you only need OpenSSH 6.5+ for the (superior) ed25519 keys, that's it. I'm running Gogs 0.6.9, it worked on 0.6.3, too.

Unknwon:
@arnehormann no ... :sweat_smile: that's just a emoji .. no special meaning..

arnehormann:
@Unknwon no problem, my bad :-)

[...]

PR #1719, opened 2015-09-29
Improve Docker

### LINKED ISSUE #1717: Gogs docker don't support ED25519 keys
If you try to add an ed25519 user ssh key the web responds:
> Sorry, we're not able to verify your SSH key: ssh-keygen -l -f: line 1 too long: ssh-ed25519 AAAAC3NsdaP5lJHI1NTA5BBBBIHSi...

Could be cause a very old SSH server?

ogarcia:
- Updated to use Debian Jessie base image.
- Less intermediate containers.
- Uses supervisor to start process.
- Fix #1717, #1708 and probably #1313.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,54 +1,48 @@
-FROM google/debian:wheezy
-MAINTAINER u@gogs.io
+FROM debian:jessie
+MAINTAINER ogarcia@connectical.com
 
-RUN echo "deb http://ftp.debian.org/debian/ wheezy-backports main" >> /etc/apt/sources.list && \
-	apt-get update -qqy && \
-	apt-get install --no-install-recommends -qqy \
-	curl build-essential ca-certificates git \ 
-	openssh-server libpam-dev && \
-	apt-get autoclean && \
-    apt-get autoremove && \
-    rm -rf /var/lib/apt/lists/*
-
-ENV GOROOT /goroot
-ENV GOPATH /gopath
-ENV PATH $PATH:$GOROOT/bin:$GOPATH/bin
-
-COPY . /gopath/src/github.com/gogits/gogs/
-WORKDIR /gopath/src/github.com/gogits/gogs/
+# Update and install required packages
+RUN apt-get update -qqy && \
+  apt-get install --no-install-recommends -qqy \
+    build-essential ca-certificates curl git libpam-dev \
+    openssh-server supervisor && \
+  apt-get autoclean && \
+  apt-get autoremove && \
+  rm -rf /var/lib/apt/lists/*
 
 # Build binary and clean up useless files
-RUN mkdir /goroot && \
-	curl https://storage.googleapis.com/golang/go1.5.linux-amd64.tar.gz | tar xzf - -C /goroot --strip-components=1 && \
-	go get -v -tags "sqlite redis memcache cert pam" && \
-	go build -tags "sqlite redis memcache cert pam" && \
-	mkdir /app/ && \
-	mv /gopath/src/github.com/gogits/gogs/ /app/gogs/ && \
-	rm -r $GOROOT $GOPATH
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks!

Comments are made inline.

Dockerfile
```diff
@@ -1,54 +1,48 @@
...
-    apt-get autoremove && \
-    rm -rf /var/lib/apt/lists/*
-
-ENV GOROOT /goroot
-ENV GOPATH /gopath
-ENV PATH $PATH:$GOROOT/bin:$GOPATH/bin
-
-COPY . /gopath/src/github.com/gogits/gogs/
-WORKDIR /gopath/src/github.com/gogits/gogs/
+FROM debian:jessie
```
Unknwon: Any reason changed to this?

docker/start.sh
```diff
@@ -40,4 +37,4 @@ then
 fi
 
 chown -R git:git /data .
-exec su git -c "./gogs web"
+exec supervisord -c /app/gogs/docker/supervisord.conf
```
Unknwon: Why use supervisor simply ./gogs web would work?

Dockerfile
ogarcia: Yes. The ENV variables still remain in docker although are unneeded cause GOROOT and GOPATH will be deleted.

I think that could keep COPY line (`COPY . /gopath/src/github.com/gogits/gogs/`), cause the Docker image are build with DockerHub, in this case we can remove the git clone lines (16 and 17):
> git clone https://github.com/gogits/gogs.git \
>    /gopath/src/github.com/gogits/gogs && \

And can remove the `mkdir /gopath/src/github.com/gogits` in 14 cause the copy do the magic.

ogarcia: Mmmm. I will do this changes and update te PR. ;)

ogarcia:
Is better with the second change c92515ae5d221e9a4ceff1663233bb630701a88f, cause the git clone not is neccesary inside docker if you have a local copy of repo.

docker/start.sh
ogarcia: Cause supervisor runs sshd too.

In docker you must run only one master process. In this case the master process is `supervisord` and it have `sshd` and `gogs` as child processes.

Ah, and this must fix #1499. :wink:

Dockerfile
ogarcia: Ah! If you say the FROM line:
`FROM debian:jessie`

This base distro is officially supported by docker team, and is most recent than `google/debian:wheezy`.

It fixes the issues reated with SSH and ED25519 keys. :smile:

tboerger:
I would suggest to use alpine as a base image, that way the container gets really tiny. And instead of supervisor I would use s6 as this works really correct as PID1 within docker, and it doesn't rely on python which increases the image size as well. I have already done something like that for [gitea](https://github.com/dockhippie/gitea), my image is still pretty large as I have installed logstash-forwarder and other tools in my base image.

tboerger:
Beside that you are not rotating the logs and you are not printing the logs to stdout which sucks.

ogarcia:
@tboerger we could use Alpine, but the main target is that the docker can be *easy to use and modify*, on the other hand, supervisor is only one package cause python is installed in base system.

I don't understand the problem with log rotation? Supervisor rotates the log automagically and in stdout you see the main proccess log (supervisord).

tboerger:
It is easy to use and modify with alpine. Just python bloats the image already massively... Just as an example, a full alpine image takes ~5mb. And where rotates supervisor all the logs? Beside that got supervisor issues as pid1 in docker.

Unknwon:
Thanks your work again!

And close by https://github.com/gogits/gogs/pull/1730 

You can continue improving based on https://github.com/gogits/gogs/pull/1730 .

Issue #1721, opened 2015-09-30
See users who starred/watch/forked repositories


soudy:
As of `Gogs version 0.6.15.0926 Beta` you are able to see the amount of people who have followed/starred/forked a repository, but not who. I think it would be a nice addition to also see the people's names.

Unknwon:
Thanks your feedback!

soudy:
implemented -> #1725

Unknwon:
:smile:

PR #1725, opened 2015-10-01
Implemented #1721: see users who forked/starred/watched a repository

soudy:
The logic and templates are there and are accessible through their paths (/repo/stars, /repo/forks, /repo/watchers). It is however not implemented in the repo's UI.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -514,6 +514,9 @@ func runWeb(ctx *cli.Context) {
 		m.Get("/labels/", repo.RetrieveLabels, repo.Labels)
 		m.Get("/milestones", repo.Milestones)
 		m.Get("/branches", repo.Branches)
+		m.Get("/stars/?:index", middleware.RepoRef(), repo.Stars)
+		m.Get("/watchers/?:index", middleware.RepoRef(), repo.Watchers)
+		m.Get("/forks", middleware.RepoRef(), repo.Forks)
 		m.Get("/archive/*", repo.Download)
 
 		m.Group("/pulls/:index", func() {
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -514,8 +514,8 @@ func runWeb(ctx *cli.Context) {
 		m.Get("/labels/", repo.RetrieveLabels, repo.Labels)
 		m.Get("/milestones", repo.Milestones)
 		m.Get("/branches", repo.Branches)
-		m.Get("/stars/?:index", middleware.RepoRef(), repo.Stars)
-		m.Get("/watchers/?:index", middleware.RepoRef(), repo.Watchers)
+		m.Get("/stars", middleware.RepoRef(), repo.Stars)
+		m.Get("/watchers", middleware.RepoRef(), repo.Watchers)
 		m.Get("/forks", middleware.RepoRef(), repo.Forks)
 		m.Get("/archive/*", repo.Download)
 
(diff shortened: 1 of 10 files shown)


cmd/web.go
```diff
@@ -514,6 +514,9 @@ func runWeb(ctx *cli.Context) {
 		m.Get("/labels/", repo.RetrieveLabels, repo.Labels)
 		m.Get("/milestones", repo.Milestones)
 		m.Get("/branches", repo.Branches)
+		m.Get("/stars/?:index", middleware.RepoRef(), repo.Stars)
+		m.Get("/watchers/?:index", middleware.RepoRef(), repo.Watchers)
+		m.Get("/forks", middleware.RepoRef(), repo.Forks)
```
tboerger: Why did you not add an index here as well?

public/ng/less/gogs/repository.less
```diff
@@ -795,3 +795,84 @@
         padding-top: 15px;
     }
 }
+
+#stars, #watchers, #forks {
```
tboerger: Does it really make sense to add that much custom CSS for this simple kind of pages?

cmd/web.go
soudy: Forks show less data than stars and watchers, so it all fits pretty well on one page.

tboerger:
The templates got in general pretty bad indention.

soudy:
@tboerger could you elaborate on what's wrong with the indention?

cmd/web.go
tboerger: I think that toally depends on the size of the gogs install. I would suggest to add paging but display it only for a bigger number of rows.

tboerger:
> @tboerger could you elaborate on what's wrong with the indention?

Just look at the diff? Maybe https://github.com/gogits/gogs/pull/1725/files#diff-6e6e73dad0326f79bf8064a4851cad92R35

Fastidious:
Looks sloppy to me, I agree with @tboerger, and a lot of new CSS.

soudy:
@tboerger strangely enough it looked fine in my text editor, thanks
@Fastidious these are 3 new pages, some new CSS was needed

tboerger:
@soudy thankfully to the CSS framework I'm pretty sure we don't need that much heavily customized CSS.

tboerger:
I still dislike the indention of that, but at the end it's the decision of @Unknwon.

I'm absolutely sure that this CSS is not required, especially the pagination CSS must be already somewhere present because it's not the first page including pagination.

You are still referring to some old templates?

Beside of that you should create proper commit messages. Maybe sonsider some git rebase instead of git merge commands, that way you can avoid commit messages like `	Merge branch 'develop' of https://github.com/soudy/gogs into develop`. At the end of a pull request it makes sense to squash the commits.

soudy:
@tboerger thanks for the advice, I'll see what I can do with the CSS

Unknwon:
@soudy very much thank to your PR!

However, you should really discuss on [Gitter.im](https://gitter.im/gogits/gogs) as big red notice here: https://github.com/gogits/gogs#notices

Because some details are not shown in the surface. For example, the page style you have in the PR is based on old CSS theme which is going to removed at some point.

Always ask and confirm and save very much of your time. :stuck_out_tongue_closed_eyes:

PR #1730, opened 2015-10-02
New approach to Gogs Docker Container

### LINKED ISSUE #1499: Use ctrl+c when clone through SSH makes Gogs docker exit unexpectedly

0rax:
This is a complete rewrite of the current Gogs docker container.

Some stuff was bothering me like the size of the container or the fact that we could not use localhost to connect to linked database.

I've rewrite it for this purpose while implementing other stuff that people wanted

This PR provide the following:
- New container base on `alpine:3.2` making the base container lighter
- S6 as system init and process supervisor (does not require a scripting runtime and pretty straightforward to configure)
- A build script for the application, this help with the size of the container as only one layer is created when the application is compiled. This script install building dependencies, compile the software and then remove those dependencies. The resulting layer then only contains the app folder without the needed building dependencies.
- This container implement a `VOLUME` directive for the `/data` folder as requested in #1708 
- OpenSSH is configured to handle `rsa`,`dsa`,`ecdsa` and `ed25519` key fixing #1717
- Fixes  #1499

The resulting size of this docker container is 77mo.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,7 @@
-.git/*
-conf/*
-packager/*
-scripts/*
+.git
+conf
+packager
+scripts
 *.yml
 *.md
 .bra.toml
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# Install build deps
+apk -U --no-progress add linux-pam-dev go@community gcc musl-dev
+
+# Init go environment to build Gogs
+mkdir -p ${GOPATH}/src/github.com/gogits/
+ln -s /app/gogs/ ${GOPATH}/src/github.com/gogits/gogs
+cd ${GOPATH}/src/github.com/gogits/gogs
+go get -v -tags "sqlite redis memcache cert pam"
+go build -tags "sqlite redis memcache cert pam"
+
+# Cleanup GOPATH
+rm -r $GOPATH
+
+# Remove build deps
+apk --no-progress del linux-pam-dev go gcc musl-dev
--- a/docker/s6/.s6-svscan/finish
+++ b/docker/s6/.s6-svscan/finish
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec /bin/true
--- a/docker/s6/gogs/run
+++ b/docker/s6/gogs/run
@@ -1,0 +1,28 @@
+#!/bin/sh
+USER=git
+USERNAME=$USER
+
+if ! test -d /data/gogs; then
+	mkdir -p /data/gogs/data /data/gogs/conf /data/gogs/log /data/git
+fi
+
+if ! test -d ~git/.ssh; then
+    mkdir ~git/.ssh
+    chmod 700 ~git/.ssh
+fi
+
+if ! test -f ~git/.ssh/environment; then
+    echo "GOGS_CUSTOM=/data/gogs" > ~git/.ssh/environment
+    chown git:git ~git/.ssh/environment
+    chown 600 ~git/.ssh/environment
+fi
+
+ln -sf /data/gogs/log  /app/gogs/log
+ln -sf /data/gogs/data /app/gogs/data
+ln -sf /data/gogs/conf /app/gogs/conf
+
+chown -R git:git /data /app/gogs ~git/
+
+export USER
+export USERNAME
+exec gosu $USER /app/gogs/gogs web
--- a/docker/s6/openssh/run
+++ b/docker/s6/openssh/run
@@ -12,4 +12,4 @@ then
 	chmod 600 /data/ssh/*
 fi
 
-exec gosu root /usr/sbin/sshd -D -f /etc/ssh/sshd_config
+exec gosu root /usr/sbin/sshd -D -f /app/gogs/docker/sshd_config
--- a/docker/s6/openssh/run
+++ b/docker/s6/openssh/run
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+if ! test -d /data/ssh
+then
+	mkdir -p /data/ssh
+	ssh-keygen -q -f /data/ssh/ssh_host_key -N '' -t rsa1
+	ssh-keygen -q -f /data/ssh/ssh_host_rsa_key -N '' -t rsa
+	ssh-keygen -q -f /data/ssh/ssh_host_dsa_key -N '' -t dsa
+	ssh-keygen -q -f /data/ssh/ssh_host_ecdsa_key -N '' -t ecdsa
+	ssh-keygen -q -f /data/ssh/ssh_host_ed25519_key -N '' -t ed25519
+	chown -R root:root /data/ssh/*
+	chmod 600 /data/ssh/*
+fi
+
+exec gosu root /usr/sbin/sshd -D -f /etc/ssh/sshd_config
--- a/docker/sshd_config
+++ b/docker/sshd_config
@@ -1,0 +1,17 @@
+Port 22
+AddressFamily any
+ListenAddress 0.0.0.0
+ListenAddress ::
+Protocol 2
+LogLevel INFO
+HostKey /data/ssh/ssh_host_key
+HostKey /data/ssh/ssh_host_rsa_key
+HostKey /data/ssh/ssh_host_dsa_key
+HostKey /data/ssh/ssh_host_ecdsa_key
+HostKey /data/ssh/ssh_host_ed25519_key
+PermitRootLogin no
+AuthorizedKeysFile	.ssh/authorized_keys
+PasswordAuthentication no
+UsePrivilegeSeparation no
+PermitUserEnvironment yes
+AllowUsers git
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -1,43 +1,12 @@
-#!/bin/bash -
-#
+#!/bin/sh
 
-if ! test -d /data/gogs
-then
-	mkdir -p /var/run/sshd
-	mkdir -p /data/gogs/data /data/gogs/conf /data/gogs/log /data/git
-fi
+# Bind linked docker container to localhost socket using socat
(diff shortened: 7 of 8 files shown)


.dockerignore
```diff
@@ -1,7 +1,7 @@
-.git/*
-conf/*
-packager/*
-scripts/*
+.git
+conf
+packager
+scripts
```
Unknwon: Based on my experience, somehow without `/*` does not work...

Unknwon:
Um... what is s6?

.dockerignore
0rax: While testing using the `conf/*` form, the `conf` folder was still sent to the build context, making the symlink of `/data/gogs/conf` to `/app/gogs/conf/conf` as an empty `conf` folder was created.

This way the `conf` folder is not sent to the build context.

I admit that this is pretty weird, as `dockerignore` and `gitignore` does not really work exactly in the same way.

`gitignore` works great with the `conf/*` form as an empty folder is not sent to remote, but this is not the case with the docker build context.

Unknwon:
And can you give more details one each file what it does and why it is necessary?

.dockerignore
Unknwon: I see people are using `.git` in stead of `.git/*` but when I do `docker build .`, it still sends whole `.git` directory to container... maybe it just a bug on my machine.. :joy:

0rax:
Yes of course. Will add comment on each file.

For `s6` it is a small process supervision system that can be used as an init system. As `docker` suffer from the [PID 1 zombie reaping problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/) using a proper init system that solve this issue is a good way to build your container.

Using s6 instead of another process supervisor like `supervisord` or `runit` for this docker containers comes from the fact that `supervisord` is not made to be use as an init system (run as PID 1, see [supervisord documentation](http://supervisord.org/index.html?highlight=init)) and requires a heavy runtime depencies (python).

For the case of `runit` it comes more from a personal point of view but `s6` doesnt not requires other libraries as the one ship with it making it very small and the way process script are made are really just simple start script making maintenance easier while multiple person are working on the same container. 

If you want more information about the intricate feature of `s6`, you will found everything you need on [s6 website](http://www.skarnet.org/software/s6/).

.dockerignore
0rax: What's your docker version ?

In fact if this is version dependent, we can put them both in the `.dockerignore` as this will not break anything !

docker/build.sh
```diff
@@ -0,0 +1,17 @@
...
+ln -s /app/gogs/ ${GOPATH}/src/github.com/gogits/gogs
+cd ${GOPATH}/src/github.com/gogits/gogs
+go get -v -tags "sqlite redis memcache cert pam"
+go build -tags "sqlite redis memcache cert pam"
+
+# Cleanup GOPATH
+rm -r $GOPATH
+
+# Remove build deps
+apk --no-progress del linux-pam-dev go gcc musl-dev
```
0rax: ## Gogs Building script

This script externalize the building process of gogs from the `Dockerfile`.

The purpose of this is to reduce the size of the container by having the build dependencies only installed in the container during building. We could just create a full block inside the `Dockerfile` but having a specific script seems way cleaner.

This little trick helps us save more that 200mo on the final container.

docker/s6/.s6-svscan/finish
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec /bin/true
```
0rax: ## Init 3 Script

This script is run when the docker container is stopped (shutdown process), this is required by s6 an has no purpose in our case that's why it only calls `/bin/true`

docker/s6/gogs/run
```diff
@@ -0,0 +1,28 @@
...
+
+ln -sf /data/gogs/log  /app/gogs/log
+ln -sf /data/gogs/data /app/gogs/data
+ln -sf /data/gogs/conf /app/gogs/conf
+
+chown -R git:git /data /app/gogs ~git/
+
+export USER
+export USERNAME
+exec gosu $USER /app/gogs/gogs web
```
0rax: ## Gogs Init Script

This is the init script for the `gogs` process, the last line call the software itself, what is before is just the various check and operation required to run `gogs web`.

Another solution, instead of having initialization in the software init script is to create a `init 1` script that will do all the initialization before launching `s6`. If prefered, the change will not take long.

docker/s6/openssh/run
```diff
@@ -0,0 +1,15 @@
...
+	ssh-keygen -q -f /data/ssh/ssh_host_key -N '' -t rsa1
+	ssh-keygen -q -f /data/ssh/ssh_host_rsa_key -N '' -t rsa
+	ssh-keygen -q -f /data/ssh/ssh_host_dsa_key -N '' -t dsa
+	ssh-keygen -q -f /data/ssh/ssh_host_ecdsa_key -N '' -t ecdsa
+	ssh-keygen -q -f /data/ssh/ssh_host_ed25519_key -N '' -t ed25519
+	chown -R root:root /data/ssh/*
+	chmod 600 /data/ssh/*
+fi
+
+exec gosu root /usr/sbin/sshd -D -f /app/gogs/docker/sshd_config
```
0rax: ## OpenSSH Init Script

This is the init script for the `sshd` process. As for `docker/s6/gogs/run` it just verifies that the ssh-keys are present before launching the process itself

docker/sshd_config
```diff
@@ -0,0 +1,17 @@
...
+HostKey /data/ssh/ssh_host_rsa_key
+HostKey /data/ssh/ssh_host_dsa_key
+HostKey /data/ssh/ssh_host_ecdsa_key
+HostKey /data/ssh/ssh_host_ed25519_key
+PermitRootLogin no
+AuthorizedKeysFile	.ssh/authorized_keys
+PasswordAuthentication no
+UsePrivilegeSeparation no
+PermitUserEnvironment yes
+AllowUsers git
```
0rax: ## OpenSSH Server Config File

Found it cleaner to put the configuration in a specific file instead of creating different layers to update the system configuration. Also updating the system configuration is not really something safe, as if the package is updated and the system configuration change, the modification done to it using `sed` could fail.

docker/start.sh
```diff
@@ -1,43 +1,12 @@
...
+# Bind linked docker container to localhost socket using socat
+env | sed -En 's|(.*)_PORT_([0-9]*)_TCP=tcp://(.*):(.*)|\1_\2 socat -ls TCP4-LISTEN:\2,fork,reuseaddr TCP4:\3:\4|p' | \
+while read NAME CMD; do
+    mkdir -p /app/gogs/docker/s6/$NAME
+    echo -e "#!/bin/sh\nexec $CMD" > /app/gogs/docker/s6/$NAME/run
+    chmod +x /app/gogs/docker/s6/$NAME/run
+done
+
+# Exec S6 as process manager for gogs and dropbear ssh
+exec /usr/bin/s6-svscan /app/gogs/docker/s6/
```
0rax: ## Init 1 Script

This script is launched when the container starts, it will create init files for the different link made to this container and then start `s6` to launch `openssh`, `gogs` and a certain number of `socat` instance that will bind those link to localhost.

Unknwon:
Thanks! Merging...

tboerger:
I would suggest a setup script for each daemon (common is /etc/s6/$daemon/setup) which gets sourced by the run script instead of your current way. With your current way bash gets executed as pid 1 instead of s6 ;)

0rax:
@tboerger, dont really see what you mean ? My PID1 is an `sh` script just for the container initalization. The exec part will then replace this by s6 when everything is set up !

tboerger:
But it's still recommended to start s6 directly through cmd. I'm on mobile so I've got no access to my boolmarks related to that :)

0rax:
It might be, where did you find this as in the `s6` documentation it is stated that if needed an `init 1` script can be provided and that it just requires to exec s6 when done.

We might just dont talk about the same things, are you talking about launching `s6-svscan` or for the `setup` needed by the application ?

PR #1731, opened 2015-10-02
Several bugfixes in Docker build

ogarcia:
- Removed unnecessary variables
- Fixed symbolic links creation
- Fixed enter point
- Less intermediate containers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,21 +9,13 @@ RUN echo "@edge http://dl-4.alpinelinux.org/alpine/edge/main" | tee -a /etc/apk/
  && apk -U --no-progress add ca-certificates bash git linux-pam s6@edge curl openssh socat \
  && chmod +x /usr/sbin/gosu
 
-# Configure Go and build Gogs
-ENV GOPATH /tmp/go
-ENV PATH $PATH:$GOPATH/bin
+ENV GOGS_CUSTOM /data/gogs
 
 COPY . /app/gogs/
 WORKDIR /app/gogs/
 RUN ./docker/build.sh
 
-ENV GOGS_CUSTOM /data/gogs
-
-# Create git user for Gogs
-RUN adduser -D -g 'Gogs Git User' git -h /data/git/ -s /bin/sh && passwd -u git
-RUN echo "export GOGS_CUSTOM=/data/gogs" >> /etc/profile
-
 # Configure Docker Container
 VOLUME ["/data"]
 EXPOSE 22 3000
-CMD ["./docker/start.sh"]
+CMD ["docker/start.sh"]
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks @ogarcia !

Have you tested those changes?

ogarcia:
Yes!

If you want you can test it, i have an autobuild in https://hub.docker.com/r/ogarcia/gogs/
`docker pull ogarcia/gogs` do the magic :wink:

Dockerfile
```diff
@@ -9,21 +9,13 @@ RUN echo "@edge http://dl-4.alpinelinux.org/alpine/edge/main" | tee -a /etc/apk/
...
-
-# Create git user for Gogs
-RUN adduser -D -g 'Gogs Git User' git -h /data/git/ -s /bin/sh && passwd -u git
-RUN echo "export GOGS_CUSTOM=/data/gogs" >> /etc/profile
-
 # Configure Docker Container
 VOLUME ["/data"]
 EXPOSE 22 3000
-CMD ["./docker/start.sh"]
+CMD ["docker/start.sh"]
```
Unknwon: Only thing unsure is this, can you confirm `"docker/start.sh"` does exactly the same thing as `./docker/start.sh` does?

ogarcia: Yes, cause you are in the workdir `/app/gogs/` and run `./docker/start.sh` or `docker/start.sh` is the same.

Unknwon: Thanks your confirmation!

Unknwon:
Thanks, merging...

Dockerfile
ogarcia: :+1:

docker/s6/gogs/run
```diff
@@ -12,17 +11,16 @@ if ! test -d ~git/.ssh; then
...
-ln -sf /data/gogs/conf /app/gogs/conf
+cd /app/gogs
+
+ln -sf /data/gogs/log  ./log
+ln -sf /data/gogs/data ./data
 
 chown -R git:git /data /app/gogs ~git/
 
 export USER
-export USERNAME
```
0rax: Not sure if this line needs to be removed as this ENV var is checked during installation. I had a problem without having it. Will try to test the install process.

Thanks for the modification !

Unknwon: @0rax FYI, https://github.com/gogits/gogs/blob/master/modules/user/user.go#L11 either one of `USER` or `USERNAME` would work.

0rax: Thanks, was weird that it did not happened when USERNAME was not set :/ Will test it

ogarcia: @0rax I tested it in new install and running install and works perfect only with `USER`. As @Unknwon say, the `func CurrentUsername()` do the magic.

PR #1732, opened 2015-10-02
Docker Container: Init 1 & Initialisation

0rax:
- Now using a setup script before starting the app. The separation of the run script and the setup script will make service initialization a little bit clearer
- Now calling start.sh script as ENTRYPOINT and S6 as CMD. This way when running the container with just a shell script, the start.sh script will be launched before, making debugging easier
- Added note about `.dockerignore` ignored during Docker Hub Automated Build

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -17,7 +17,7 @@ $ mkdir -p /var/gogs
 $ docker run --name=gogs -p 10022:22 -p 10080:3000 -v /var/gogs:/data gogs/gogs
 
 # Use `docker start` if you have stopped it.
-$ docker start gogs 
+$ docker start gogs
 ```
 
 Files will be store in local path `/var/gogs` in my case.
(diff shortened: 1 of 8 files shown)


docker/README.md
```diff
@@ -68,4 +68,9 @@ Steps to upgrade Gogs with Docker:
...
 - `docker stop gogs`
 - `docker rm gogs`
-- Finally, create container as the first time and don't forget to do same volume and port mapping.
\ No newline at end of file
+- Finally, create container as the first time and don't forget to do same volume and port mapping.
+
+## Known Issues
+
+- [Use ctrl+c when clone through SSH makes Docker exit unexpectedly](https://github.com/gogits/gogs/issues/1499)
+- `.dockerignore` seems to be ignored during Docker Hub Automated build
```
Unknwon: you may want to resync.. because I just removed this section.

0rax: Should I keep the section for the `.dockerignore` issue ?

Unknwon: Hmm, yes.

Unknwon:
Thanks, merging...

PR #1736, opened 2015-10-04
Added repository remove option to API

soudy:
I wanted to implement some more features like update repo, remove user and update user but that would cause some difficulties having to work on [go-gogs-client](https://github.com/gogits/go-gogs-client).

usage is as following:

## Remove repository
    DELETE /repos/:owner/:reponame

#### Example
     $ curl -X DELETE -H 'Authorization: token TOKEN_ID' https://try.gogs.io/api/v1/repos/soud/MyRepo

#### Response
```
Status: 204 No Content
Content-Type: application/json
```

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -226,6 +226,7 @@ func runWeb(ctx *cli.Context) {
 
 				m.Group("", func() {
 					m.Post("/migrate", bindIgnErr(auth.MigrateRepoForm{}), v1.MigrateRepo)
+					m.Delete("/:owner/:reponame", v1.RemoveRepo)
 				}, middleware.ApiReqToken())
 
 				m.Group("/:username/:reponame", func() {
--- a/routers/api/v1/repo.go
+++ b/routers/api/v1/repo.go
@@ -253,3 +253,39 @@ func MigrateRepo(ctx *middleware.Context, form auth.MigrateRepoForm) {
 	log.Trace("Repository migrated: %s/%s", ctxUser.Name, form.RepoName)
 	ctx.JSON(201, ToApiRepository(ctxUser, repo, api.Permission{true, true, true}))
 }
+
+func RemoveRepo(ctx *middleware.Context) {
+	user, err := models.GetUserByName(ctx.Params(":owner"))
+	if err != nil {
+		if models.IsErrUserNotExist(err) {
+			ctx.HandleAPI(404, err)
+		} else {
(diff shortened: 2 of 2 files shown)


tboerger:
I personally prefer restful URLs, your approach gets aeay from that even more :(

soudy:
@tboerger what do you mean with 'restful url'? I believe the url is pretty restful

tboerger:
No, it's absolutely not restful. If it's a route to delete a resource it should be DELETE /repos/:reponame. Even another approach that is more restful than yours, but not the best for restful perspective, should be GET /repos/:reponame/remove.

tboerger:
For the reference, there it gets explained pretty well http://guides.rubyonrails.org/routing.html#resources-on-the-web

soudy:
@tboerger ah yes I agree that's better, I'll get onto that. thanks for the feedback

Unknwon:
Thanks your PR!

Gogs's API design mostly follow the GitHub style, see reference: https://developer.github.com/v3/repos/#delete-a-repository

soudy:
@Unknwon ok, so you'd prefer `/repos/:owner/:repo` over `/repos/:repo`?

Unknwon:
@soudy yes, otherwise, how would you delete a organization repository?

soudy:
@Unknwon good point - I forgot about organizations

Unknwon:
GitHub has pretty good documentation for APIs :smile:

soudy:
@Unknwon I've updated it to work with organisations

Unknwon:
> perhaps I should call GetOwner on the repo so Owner isn't null, what do you think?

No, `IsOwnedBy` only compare ID which can/should never be null.

soudy:
@Unknwon I was talking about the Owner: section in the JSON response

Unknwon:
@soudy 

1. See https://github.com/gogits/gogs/blob/master/routers/api/v1/repo_file.go#L23-L27 for an example of response. (`ok` is only been used for UI-related APIs)
2. Based on reference, https://developer.github.com/v3/repos/#response-11 , status 204 should be presented and no content.

Unknwon:
PS: response a status without content can use `ctx.Status(204)`.

soudy:
@Unknwon thanks for the pointers, I've updated it to Github's standars now

Unknwon:
Thanks! Looks better now, will merge soon.

Unknwon:
Thanks again, merging.

PR #1758, opened 2015-10-10
Docker Container Restart Fix

0rax:
- Fix s6 fifodir error on container restart, cleanup is done twice
    - On start so that the issue does not happen when container is killed  via `docker kill`
    - On stop so the resulting stopped container is clean if accessed by other means
- Add .tags* to .gitignore (Atom auto ctags generation)


Unknwon:
Thanks!

PR #1764, opened 2015-10-12
Volumed data setup changed to allow #1759

### LINKED ISSUE #1759: Volume directive in Dockerfile prevents folders in /data to be created
Steps to reproduce:

Build the following image:

```Dockerfile
FROM gogs/gogs
RUN /bin/bash -c 'set -exuo pipefail && mkdir -p /data/gogs/data /data/gogs/conf /data/gogs/log /data/git && ls -l /data && ls -l /data/gogs'
```

execute ``docker run $imageid ls -l /data``

Expected result: subfolders

Actual result: ``total 0``

This issue was introduced in e63e0b3105124bd8ec3028a39dc71c8d8ca103e3 and is caused by the the [volume directive](https://github.com/gogits/gogs/blob/master/Dockerfile#L19).

This behaviour forces users to mount the whole ``/data```folder as volume even if e.g. the logs are not wanted.

0rax:
- Fixes #1759 
- Volumed subfolder now created up in the ENTRYPOINT script, this way
  they are created before S6 even starts making VOLUME.
- The subfolder will be created during VOLUME creation too as ENTRYPOINT
  script will be run before /bin/true
- SSH Keys will now be created on a single key basis not replying on the
  existence of /data/ssh folder

### DIFF
--- a/docker/s6/gogs/setup
+++ b/docker/s6/gogs/setup
@@ -1,11 +1,7 @@
 #!/bin/sh
 
-if ! test -d /data/gogs; then
-    mkdir -p /data/gogs/data /data/gogs/conf /data/gogs/log /data/git
-fi
-
 if ! test -d ~git/.ssh; then
-    mkdir ~git/.ssh
+    mkdir -p ~git/.ssh
     chmod 700 ~git/.ssh
 fi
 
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks!

PR #1767, opened 2015-10-12
Fix bug introduced in #1759 preventing ssh server to launch

0rax:
- ssh/setup: test directive now check if a file exist in key path instead of a folder
- ssh/setup script was hanging waiting for an input about rewriting the key as the test case was returning true, when it should have been false (check if file is a folder instead of a file)


Unknwon:
:smiley:

PR #1799, opened 2015-10-18
Simplify disable Gravatar preference wording

legoktm:
"Enable this to disable..." is needlessly confusing, simplify it by being
upfront that this setting disables the Gravatar fetching.


Unknwon:
Thanks!

PR #1803, opened 2015-10-19
Fix import path

kenno:
Hello,
After re-compiling gogs with Redis from the master branch today, I wasn't able to run gogs. The error message was "panic: cache: unknown adapter 'redis'(forgot to import?)"
It's a minor issue with new go-macaron path.
Kenno


Unknwon:
Thanks!

kenno:
No, thank you.

PR #1836, opened 2015-10-24
workaround for reverse proxy, ssh, submodule

likeit92:
If someone use gogs with reverse proxy and try to push a ssh submodule to the server the reference url was wrong.

This workaround works only for the local server!!!

### DIFF
--- a/modules/git/submodule.go
+++ b/modules/git/submodule.go
@@ -6,6 +6,7 @@ package git
 
 import (
 	"strings"
+	"github.com/gogits/gogs/modules/setting"
 )
 
 type SubModule struct {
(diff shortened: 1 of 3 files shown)


modules/git/submodule.go
```diff
@@ -51,8 +51,14 @@ func (sf *SubModuleFile) RefUrl() string {
 	i := strings.Index(url, "@")
 	j := strings.LastIndex(url, ":")
 	if i > -1 && j > -1 {
-		return "http://" + url[i+1:j] + "/" + url[j+1:]
+		// fix problem with reverse proxy works only with local server
+		if strings.Contains(appUrl,url[i+1:j]) {
```
Unknwon: As a quick work around, why not just use `setting.AppUrl` here so you need to change anything in template file.

Unknwon:
You don't need to close and reopen a pull request, GitHub detects changes automatically.

likeit92:
okay thank you for your feedback

Unknwon:
Travis is slow today...

Unknwon:
OK, good, thanks!

PR #1845, opened 2015-10-27
Sanitize input to LDAP authentication module

### LINKED ISSUE #1841: Missing input sanitization in LDAP filters /  LDAP injection
gogs does not sanitize the username input, allowing the user to alter the filter string itself and/or construct invalid filter strings. For more information on LDAP injection see https://www.owasp.org/index.php/LDAP_injection

Example: Login with a username that contains a right parenthesis ')' character, watch the log output.

The culprit is that gogs uses simple string substitution without sanitization, i.e.
```go
userFilter := fmt.Sprintf(ls.Filter, name)
```
Source: https://github.com/gogits/gogs/blob/master/modules/auth/ldap/ldap.go#L58

[RFC 4515](http://tools.ietf.org/search/rfc4515) defines how the values should be encoded:
```
      valueencoding  = 0*(normal / escaped)
      normal         = UTF1SUBSET / UTFMB
      escaped        = ESC HEX HEX
      UTF1SUBSET     = %x01-27 / %x2B-5B / %x5D-7F
                          ; UTF1SUBSET excludes 0x00 (NUL), LPAREN,
                          ; RPAREN, ASTERISK, and ESC.
```

SergioBenitez:
This fixes #1841 and a related bug with the UserDN by disallowing usernames with characters that could lead to LDAP injection attacks.

### DIFF
--- a/modules/auth/ldap/ldap.go
+++ b/modules/auth/ldap/ldap.go
@@ -9,6 +9,7 @@ package ldap
 import (
 	"crypto/tls"
 	"fmt"
+	"strings"
 
 	"github.com/gogits/gogs/modules/ldap"
 	"github.com/gogits/gogs/modules/log"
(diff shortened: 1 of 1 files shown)


tboerger:
LGTM

Unknwon:
Thanks!

PR #1861, opened 2015-10-30
Fix the JSON field names for MigrateRepoForm

nathan7:
For some reason, the field names for Private and Mirror got swapped.


Unknwon:
Hmm.. indeed a good catch...

But can you re-PR to `develop` branch?

nathan7:
Moved to #1862.

PR #1862, opened 2015-10-30
Fix the JSON field names for MigrateRepoForm

nathan7:
For some reason, the field names for Private and Mirror got swapped.
This causes private non-mirror migrations created through the API to end up public, and public mirrors to end up private.


tboerger:
:+1:

Unknwon:
Thanks again!

PR #1863, opened 2015-10-30
Add go-source directive to repo pages

nathan7:
This makes Gogs repos work with godoc.org: https://github.com/golang/gddo/wiki/Source-Code-Links


Unknwon:
Thanks!

PR #1865, opened 2015-10-30
Translation for invalid milestone due date format ISO 8601 fix.

mattarnster:
Hello,

Just something I have noticed when going through the translation files...

As per https://en.wikipedia.org/wiki/ISO_8601 format, I have changed the translation strings (taking into account each country's ISO 8601 standard format (except for a few which haven't been translated).

So 'year-mm-dd' becomes 'yyyy-mm-dd' or the country's equivalent.

If I need to do anything else or if I have missed something, please let me know.

Thanks,
Matt


tboerger:
This won't get merged because the translations are managed through crowdin: http://gogs.io/docs/features/i18n

Unknwon:
Thanks! But the only file will take effect is `locale_en-US.ini`, because as @tboerger mentioned, all other language local files are generated from Crowdin.

Unknwon:
But, I'll do that for you on Crowdin :eyes:

PR #1871, opened 2015-11-01
move minimum key sizes to config

Gibheer:
This patch moves the, currently hard coded, minimum key sizes into the configuration file. This way it is possible to enforce company security standards in gogs. For example it is now possible to disable DSA support or enforce RSA keys with a higher minimum length.

This also makes it possible to easily support new key formats as the only place which has to be adjusted is the configuration file.


tboerger:
:+1:

Unknwon:
Thanks a lot!

PR #1872, opened 2015-11-01
[Translators Wanted!] Change "Close and comment" to "Comment and close"

### LINKED ISSUE #1831: "close and comment" should be "comment and close"
I think it would make more sense (as a single action) to comment on an issue then close it. At the moment, the "close and comment" button does exactly that - it closes, then comments. This can be hard to tell if the comment was made AFTER the close, and is suggesting a re-open without actually reading the comment. Whereas, if the comment followed by a close is the last comment, it's easy to see there's no commentary after the close! (0.6.9)

JohnMaguire:
This would close #1831 if we could get other translations in this pull request.

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -455,7 +455,7 @@ issues.num_comments = %d comments
 issues.commented_at = `commented <a id="%[1]s" href="#%[1]s">%[2]s</a>`
 issues.no_content = There is no content yet.
 issues.close_issue = Close
-issues.close_comment_issue = Close and comment
+issues.close_comment_issue = Comment and close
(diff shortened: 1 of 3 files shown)


tboerger:
Gogs uses crowdin for handling transpations.

JohnMaguire:
@tboerger Is that something I can handle / does that make this pull request void?

tboerger:
Yes, you can read how it works at http://gogs.io/docs/features/i18n

This makes this pull request obsolete.

JohnMaguire:
Thank you. :)

ChangePlaces:
Thanks anyway :)

Unknwon:
All changes are made and pushed to `develop` branch.

PR #1889, opened 2015-11-04
adding Makefile to auotomate the gogs build process

tjyang:
Hi Unknown

I like this Makefile , not sure about you. This should work in Windows "nmake" when building gogs on window also. I had not try it on Windows yet.


tjyang:
I did not check your master for awhile, there is a Makefile already. cancelling this PR.

PR #1890, opened 2015-11-04
Develop

tjyang:
adding "make check"  target into Makefile.


Unknwon:
Thanks, but I think check and goimports will never be used.

Issue #1893, opened 2015-11-04
How troubleshoot SSHd on Docker ?


ygbillet:
Hi,

I use GOGS as Docker appliance.
I just update to the last version.

I added my public key to my account.
Try to connect with `ssh -vT git myhostname` and get `Permission denied (publickey,keyboard-interactive).`

I check `/home/git/.ssh/.authorized_keys` and seems OK.
Is there a way to troubleshoot OpenSSH in GOGS running in Docker (where are the logs ?) ?

Unknwon:
> Permission denied (publickey,keyboard-interactive)

Get into Docker and check SSH auth.log.

Unknwon:
cc @0rax

ygbillet:
Did not find "auth.log" file in my container.

Am i missing something ?

Unknwon:
Check if `/home/git/.ssh/.authorized_keys`'s permission is `600`.

0rax:
Can you try to connect to your serveur via ssh ?

```
ssh git@my-server.ext
```

It should output
```
Hi there, You've successfully authenticated, but Gogs does not provide shell access.
If this is unexpected, please log in with password and setup Gogs under another user.
Connection to my-server.ext closed.
```

ygbillet:
permission is `600`.

Cannot connect to ssh.

ygbillet:
- Service SSH is up and running (get Permission denied and OK from `ps`)
- `authorized_key`'s permission is 600 and public key is in file.

    1 root       0:00 /usr/bin/s6-svscan /app/gogs/docker/s6/
    13 root       0:00 s6-supervise gogs
    14 root       0:00 s6-supervise openssh
    15 root       0:00 /usr/sbin/sshd -D -f /app/gogs/docker/sshd_config
    16 git        0:00 /app/gogs/gogs web
    61 root       0:00 /bin/bash
    73 root       0:00 ps

Bun cannot find auth.log or any log from openssh.

0rax:
To get access to sshd log, you will need to activate syslogd (not launched by default).

Just do a `docker exec gogs-container-name syslogd`, you will then be able to get sshd log inside `/var/log/messages`.

Currently working on a way to print sshd logs inside `docker logs`, that why syslogd was not launched by default. I should had it for now. my bad.

Unknwon:
@ygbillet can you dump your `authorized_keys` file here?

ygbillet:
Actually with logging enabled it's better 

[...]

Issue #1815, opened 2015-10-21
gogs docker fails on Kubernetes


kai0:
Kubernetes (http://kubernetes.io/) gogs config files:

- gogs-rc.yaml

```yaml
apiVersion: v1
kind: ReplicationController
metadata:
  name: gogs
spec:
  replicas: 1
  selector:
    name: gogs
  template:
    metadata:
      labels:
        name: gogs
    spec:
      containers:
        - name: gogs
          image: gogs/gogs
          ports:
            - containerPort: 3000
            - containerPort: 22
          volumeMounts:
            - mountPath: /data
              name: data
        - name: postgres
          image: postgres
          ports:
            - containerPort: 5432
          volumeMounts:
            - mountPath: /var/lib/postgresql/data
              name: postgres
          env:
            - name: POSTGRES_USER
              value: "gogs"
            - name: POSTGRES_PASSWORD
              value: "gogs"
      volumes:
        - name: data
          hostPath:
            path: /var/gogs
        - name: postgres
          hostPath:
            path: /var/gogs-postgres
```

- gogs-service.yaml

```yaml
apiVersion: v1
kind: Service
metadata:
  name: gogs
spec:
  type: NodePort
  ports:
    - port: 3000
      protocol: TCP
      nodePort: 33000
      name: a
    - port: 22
      protocol: TCP
      nodePort: 30022
      name: b
  selector:
    name: gogs
```

Steps to reproduce:
1. kubectl create -f gogs-rc.yaml
2. kubectl create -f gogs-service.yaml
3. kubectl stop rc gogs
4. kubectl create -f gogs-rc.yaml

Expected result:
- Gogs accessible from host:33000.

Actual result:
- web.go:587 runWeb()] [E] Fail to start server: listen tcp 0.0.0.0:3000: bind: address already in use

kai0:
When I `ps -ef` inside the container, I see tons of:
```
socat -ls TCP4-LISTEN:3000,fork,reuseaddr TCP4:10.254.175.191:3000
```

10.254.175.191 is kubernetes' gogs service virtual IP:

```bash
# kubectl get services
NAME         LABELS                                    SELECTOR    IP(S)            PORT(S)
gogs         <none>                                    name=gogs   10.254.175.191   3000/TCP
                                                                                    22/TCP
kubernetes   component=apiserver,provider=kubernetes   <none>      10.254.0.1       443/TCP
nginx        run=nginx                                 run=nginx   10.254.63.3      80/TC
```

[...]

PR #1898, opened 2015-11-05
Docker Improvmement and small fix

0rax:
- Added `SOCAT_LINK` environment variable to prevent SOCAT link creation as dicussed in #1815 
- Added right & ownership directory rigths to `/data/git`, `/data/gogs` & `/data/ssh` preventing weird bug like #1893
- Added `syslogd` service
  - gather `openssh` log
  - print them to stdout to let us track it via `docker logs`
- Docker image for testing availlable at [orax/gogs](https://hub.docker.com/r/orax/gogs/tags/) tag `latest`


Unknwon:
Thanks!

PR #1927, opened 2015-11-11
Fix #1824 add tabindex

### LINKED ISSUE #1824: Tabbing when editing a comment / issue
When editing a comment or issue, if I tab, I'd like to be taken to the save button so I can then press spacebar and have the issue / comment update. However this doesn't happen. Tabbing takes me to the next comment save button instead! (0.6.9)

makhov:
Fix #1824 issue


Unknwon:
Thanks!

PR #1928, opened 2015-11-11
#1917 add tabindex

### LINKED ISSUE #1917: Can't tab into "close and comment" button
I think related to #1824 it seems impossible to tab into the close and comment button after typing the message.

makhov:
Fix #1917


Unknwon:
Oops... I didn't notice this PR is coming... and I have fixed that one based your previous PR.

Unknwon:
PS: as a result of testing, the new comment form tag indexes should be 4, 5, 6...

PR #1933, opened 2015-11-12
HOME fix into FreeBSD init script

renatoaquino:
The gogs_directory is not necessarily the same as the HOME directory.
WIthout this gogs ends up writing the authorized_keys file at  $gogs_directory/.ssh/authorized_keys which is not always the right place for it.


Unknwon:
Thanks!

PR #1935, opened 2015-11-12
Show issue title at dashboard

### LINKED ISSUE #1854: Show issue title for comment news feeds
it could go just under the header line:

**username commented on issue organisation/project-name#221**
*Need to process the input using sql queries directly*
"Comment text goes here"

This is because with so many issues, it can become a little confusing what issue number refers to what issue!

makhov:
Resolve #1854

### DIFF
--- a/models/action.go
+++ b/models/action.go
@@ -138,8 +138,8 @@ func (a Action) GetIssueInfos() []string {
 }
 
 func (a Action) GetIssueTitle() string {
-	issueID := com.StrTo(a.GetIssueInfos()[0]).MustInt64()
-	issue, err := GetIssueByID(issueID)
+	issueIndex := com.StrTo(a.GetIssueInfos()[0]).MustInt64()
+	issue, err := GetIssueByIndex(a.RepoID, issueIndex)
 	if err != nil {
 		log.Error(4, "GetIssueByID: %v", err)
 		return "500 when get title"
--- a/models/action.go
+++ b/models/action.go
@@ -10,12 +10,12 @@ import (
 	"fmt"
 	"path"
 	"regexp"
-	"strconv"
 	"strings"
 	"time"
 	"unicode"
 
 	"github.com/go-xorm/xorm"
+	"github.com/Unknwon/com"
 
 	api "github.com/gogits/go-gogs-client"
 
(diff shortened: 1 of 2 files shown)


models/action.go
```diff
@@ -136,6 +137,12 @@ func (a Action) GetIssueInfos() []string {
 	return strings.SplitN(a.Content, "|", 2)
 }
 
+func (a Action) GetIssueTitle() string {
+	issueID, _ := strconv.Atoi(strings.SplitN(a.Content, "|", 2)[0])
```
Unknwon: You can import `github.com/Unknwon/com` and use `com.StrTo().MustInt64`.

Also,

```go
issueID := com.StrTo(a.GetIssueInfos()[0]).MustInt64()
```

makhov:
update pr

models/action.go
```diff
@@ -136,6 +137,12 @@ func (a Action) GetIssueInfos() []string {
 	return strings.SplitN(a.Content, "|", 2)
 }
 
+func (a Action) GetIssueTitle() string {
+	issueID := com.StrTo(a.GetIssueInfos()[0]).MustInt64()
+	issue, _ := GetIssueByID(issueID)
```
Unknwon: Should we check error here?

`if err !=nil` then `log.Error(4, "GetIssueByID: %v", err)` and returns somethings like `500 when get title` as title?

makhov: Fix it again

models/action.go
```diff
@@ -136,6 +137,16 @@ func (a Action) GetIssueInfos() []string {
 	return strings.SplitN(a.Content, "|", 2)
 }
 
+func (a Action) GetIssueTitle() string {
+	issueID := com.StrTo(a.GetIssueInfos()[0]).MustInt64()
```
Unknwon: Thanks. Now the question is index `0` is actually the index of issue, not the title...

makhov: As I understood index ```0``` is the issue id. At the next line I call ```GetIssueById```.
What's wrong?

Unknwon: Oops, my bad, I really mean it is the issue index, not issue ID.

Unknwon: But I know how to fix it, you can leave it to me or I can indicate where changes should be made. Which way you prefer?

makhov: I see. Please, look at last commit.

makhov:
Change Id to Issue

Unknwon:
It should work now, thanks!

PR #1949, opened 2015-11-15
Fix bad issue links in non-readme markdown files

TheHowl:


Unknwon:
Thanks!

Can you given an example of the bug this PR fixes?

TheHowl:
I've already written it in the comments of the source, but oh well.

* Assume we have the documentation files in a sub folder of the root repo. Say, like in my case, in the `rfc/` folder.
* In order to access them, we would navigate in the repository, find the files, and open them.
* When opening a file, the URL looks like this: 
* If there is a link to an issue of the repository (e.g. ), it should become a link to the issue in the repository.
* What it actually does, though, is this:

Unknwon:
Thanks, I got it now.

But directly slice `[:3]` may not be safe, may be should first use `i:=strings.Index(urlPrefix, "src")`, if found then `urlPrefix[:i]`?

TheHowl:
Genious! Gonna do it asap.

TheHowl:
done

Unknwon:
Looks good, thanks!

Unknwon:
OK.. I think it's not the safest way... what if your user name or repo name is `src`...

TheHowl:
Hmmm. then perhaps we should check whether the slash is present at least two times in `urlPrefix[:i]`?

Unknwon:
There are two ways I can think of:

1. Does not allow user or repo name starts with `src`.
2. Use a helper function to check `/` in `urlPrefix`, once it occurs the third, make a `[:i]` and early return, otherwise let it finish the loop.

```go
count:=0
for i:= 0; i<len(urlprefix);i++{
    if urlprefix[i]=='/'{
        count++
    }
    if count >= 3 {
    return urlprefix[:i]
}
}
return urlprefix
```

TheHowl:
The second option would be actually quite the same as `strings.Join(strings.Split(urlPrefix, "/")[:3], "/")`. That is because even if we were using the function in, say, `/:owner/:repo`, `len(strings.Split(urlPrefix, "/"))` must at the very least be 3 (`[]string{ "", ":owner", ":repo" }`). [(See this for proof it actually works)](http://ideone.com/BtkSOo)

Unknwon:
Hmm.. Not sure why Go does not throw `index out of bound`, but the second one certainly has much better performance.

TheHowl:
Well yeah, yours is better performant... by ~1.4µs

http://ideone.com/BtkSOo

Unknwon:
No, it is actually ~15x faster.

```go
package test

import (
	"strings"
	"testing"
)

func cutoutSuffix(urlprefix string) string {
	count := 0
	for i := 0; i < len(urlprefix); i++ {
		if urlprefix[i] == '/' {
			count++
		}
		if count >= 3 {
			return urlprefix[:i]
		}
	}
	return urlprefix
}

func Benchmark_CorrectPrefix(b *testing.B) {
	for i := 0; i < b.N; i++ { //use b.N for looping
		strings.Join(strings.Split("/owner/repo", "/")[:3], "/")
	}
}

func Benchmark_CorrectPrefix2(b *testing.B) {
	for i := 0; i < b.N; i++ { //use b.N for looping
		cutoutSuffix("/owner/repo")
	}
}

func Benchmark_VerbosePrefix(b *testing.B) {
	for i := 0; i < b.N; i++ { //use b.N for looping
		strings.Join(strings.Split("/owner/repo/src/master/hello", "/")[:3], "/")
	}
}

func Benchmark_VerbosePrefix2(b *testing.B) {
	for i := 0; i < b.N; i++ { //use b.N for looping
		cutoutSuffix("/owner/repo/src/master/hello")
	}
}
```

```
Benchmark_CorrectPrefix-8 	 5000000	       286 ns/op	      80 B/op	       3 allocs/op
Benchmark_CorrectPrefix2-8	100000000	        22.3 ns/op	       0 B/op	       0 allocs/op
Benchmark_VerbosePrefix-8 	 5000000	       355 ns/op	     128 B/op	       3 allocs/op
Benchmark_VerbosePrefix2-8	50000000	        23.5 ns/op	       0 B/op	       0 allocs/op
```

TheHowl:
Interesting. Gonna make new PR tomorrow to implement that.

Unknwon:
I can fix it real quick, if you don't mind :D

because i'm going to forget it real quick...

Unknwon:
Pushed :D

TheHowl:
:+1: :+1:

Issue #1850, opened 2015-10-28
Problem with creating a pull request


bioven:
Hey guys,

i have a problem with creating a pull request. 
I don't know how i create one. Maybe i'm just dumb and/or did something wrong in the installation part. 

But there is no button to create a pull request even when i'm in the menu for the pull request. There is only a button to create an issue. 

I assume this could be a bug, but i'm no really sure.
Here is picture of my case: http://imgur.com/iyQoyDY

I'm using the following version of gogs.
Gogs-Version: 0.6.16.1022 Beta

I hope you can help me and tell me how to do it.

Greetings, 
Ben

andreynering:
To create a pull request you have to fork the repository and click in this button:

![image](https://cloud.githubusercontent.com/assets/7011819/10805070/5afc7562-7db2-11e5-9b99-a1f9cce9fdda.png)

Unknwon:
@andreynering said what I'm going to say :D

bioven:
Thanks guys,

i feel pretty dumb now for asking this question. xP

But i get it now.

Unknwon:
:trollface:

Unknwon:
Close as fixed...

PR #1950, opened 2015-11-15
Change regex for /compare/

TheHowl:
For full reference about the git rules on branch names, see `man git check-ref-format`.

I've changed the regex to allow many more branches names in the url, including branch names that include a slash. I've not implemented all the rules as defined by `man git check-ref-format`, however I don't think this is required because git should error from the very moment it notices there's something odd in the branch name.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -536,7 +536,7 @@ func runWeb(ctx *cli.Context) {
 			m.Get("/forks", repo.Forks)
 		}, middleware.RepoRef())
 
-		m.Get("/compare/:before([a-z0-9]{40})...:after([a-z0-9]{40})", repo.CompareDiff)
+		m.Get("/compare/:before(.*[^.])...:after(.*[^.])", repo.CompareDiff)
 	}, ignSignIn, middleware.RepoAssignment(true))
 
 	m.Group("/:username", func() {


Unknwon:
Thanks!

The regexp is done by purpose in the current way.

Compare between branches normally means you want to make a PR, so https://github.com/TheHowl/gogs/blob/fix-compare-branch-names/cmd/web.go#L510-L511 allows all names.

TheHowl:
Oh - that explains it. I was (and still am) not quite sure on how to make a pull request (as the button to make it seems to have been removed with the new design, and pressing the 'New Issue' button from the pull requests page creates, in fact, an issue), and I thought that it wasn't working because of the wrong regex there :P

Unknwon:
See https://github.com/gogits/gogs/issues/1850 :D

TheHowl:
That's the point of my comment - it's not there anymore!

Unknwon:
You have to fork it first...

TheHowl:
oh! - that's a bit of a bummer. I find it useful when, e.g., I'm working in a collaborative project under my username (such as that I can't fork it) and I want others to be able to review the code.

Unknwon:
>  a collaborative project under my username (such as that I can't fork it) and I want others to be able to review the code.

PR between same repository will be implemented, there is a issue for this.

PR #1962, opened 2015-11-16
Update S6 Path after Alpine Package update

0rax:
- Alpine Package has been update changing the path to S6 binaries
- Updated path & added some debug to build.sh


Unknwon:
Thanks!

Issue #1975, opened 2015-11-18
[i18n] Elapsed time formatting


wqweto:
There is a problem with elapsed time formatting for most languages that don't follow original english word order as in "%d seconds %s" -> "5 seconds ago". Most of the languages use wording like "before 5 seconds" so translators attempt to fix this by swapping %s and %d as in "%s %d segundos" from Spanish translation.

Unfortunately this "hack" does not mix well with go's positional formatting which first passes 5 and then "ago" as parameters to format and the whole translation collapses with final result "%!s(int64=5) %!d(string=hace) segundos" indicating format params type mismatch.

There has to be special care taken for this translation "hack" to work (swap formatting params) or include every variation of %s in locale files. Second option is more flexible as there could be languages where concatenating a number, a prеposition and a direct translations of "ago", "before", etc. might not produce anything meaningful but this option will introduce lots of similar new strings for translators to work on.

I'll try to further investigate how the "hack" can be impl in current codebase.

wqweto:
Irrelevant.

PR #1983, opened 2015-11-18
Fixes elapsed time formatting issue #1975

wqweto:


wqweto:
Major embarrassment: I have to close this PR as irrelevant as `sprintf` supports `Explicit argument indexes` like `%[2]s` and `%[1]d` so this whole translation thread can be resolved by putting these in the original English strings so that translators can shuffle numeric and string params as pleased.

Sorry again for my ignorance. Will now go read a book on golang or something :-))

Issue #1965, opened 2015-11-17
The hyperlink to the branch is wrong if the branch is in a folder or the branch name has '#'


arthuroy:
Suppose I have a branch named ```'feature/#1965'```, and the following history will be listed on the page of the dashboard feed after I push something to the repository.

***[john.doe](/john.doe) pushed to [#1965](/x_corp/project_x/src/#1965) at [x_corp/project_x](/x_corp/project_x)***

```html
<p class="text-bold">
    <a href="/john.doe">john.doe</a>
    pushed to <a href="/x_corp/project_x/src/#1965" rel="nofollow">#1965</a> at <a href="/x_corp/project_x" rel="nofollow">x_corp/project_x</a>
</p>
```

When I click the hyperlink, ```/x_corp/project_x/src/#1965```, the page show 'Not Found'.

The hyperlink of the branch, ```'feature/#1965'```, shall be:

 /x_corp/project_x/src/***feature/%231965***

and the branch name to display shall be ```'feature/#1965'```, not ```'#1965'```

Unknwon:
Thanks your feedback!

tboerger:
You are really including a hash within the branch name? Sounds pretty error proven on the terminal

arthuroy:
@tboerger Well, it's not my decision to define the branch name ;) But I had told the management team about the hash thing.

arthuroy:
@Unknwon 
About the [pull request 1986](https://github.com/gogits/gogs/pull/1986), do I need to provide you the bug example? I thought this issue had provided it. Please enlighten me, thanks.

Unknwon:
> About the pull request 1986, do I need to provide you the bug example? I thought this issue had provided it. Please enlighten me, thanks.

Yes, please.

arthuroy:
@Unknwon 
Ok, I setup a branch like the issue said at https://try.gogs.io/john.doe/project_x
And I had add your account to the collaborator, you can see what listed below on the dashboard afte you logined:

```
john.doe pushed to #1965 at john.doe/project_x

 ceb578ba4e Initial commit
 bfd5a61610 Initial commit
```

Unknwon:
Thanks, I can see it now on https://try.gogs.io/john.doe?tab=activity

arthuroy:
@Unknwon 
Ahhh, I didn't know it can be viewed from personal activity page. Thanks.

Unknwon:
I have updated everything to `master`.

[...]

PR #1986, opened 2015-11-18
Fix the hyperlink and the name of the branch #1965

arthuroy:
The hyperlink and the display name of the branch if the branch is in a folder or the branch name has '#'.

I didn't commit modules/bindata/bindata.go, you need to re-generate and commit it.


wqweto:
Locale files modification is an overkill. These get pulled from https://crowdin.com/project/gogs so will definately get replaced if not translated there. You can leave the fix to translators if PR gets accepted.

arthuroy:
@wqweto Ahhhh, thanks so much. I didn't know I can leave the fix to translators ;)

Unknwon:
Hmm... sorry I didn't get the bug example... And as @wqweto mentioned, all translation except English should be done via Crowdin.

PR #1992, opened 2015-11-18
fix #1957 - disable version display

### LINKED ISSUE #1957: Disable version display
For security reasons, there should be an option to disable the display of the version numbers (for gogs and also for go). I prefer not to expose it to everyone (specially on BETA software).

Gibheer:
This allows the admin to disable the version information about gogs and
go in use in the footer.

This fixes #1957.


Unknwon:
Thanks!

SoftCreatR:
Thx

PR #2010, opened 2015-11-20
[WIP] Started to integrate DroneCI

tboerger:


tboerger:
As an addition I will add/update tasks within the Makefile. And maybe we should discuss the usage of `gox` for simplified cross compilation :)

Unknwon:
If no CGO, the easiest way to cross compile is upgrade to Go 1.5

tboerger:
But with sqlite support CGO should be enabled

Unknwon:
Yes, so in CI, we can use gox, but the bootstrap time I think is quite long.

vtolstov:
Does gogs support status api to allow drone notify gogs about successful build ?

Unknwon:
Currently not support

tboerger:
@vtolstov that's a totally different topic. This PR ist just for adding .drone.yml to gogs :)

tboerger:
Waiting for the successful build but it looks promising. When everything works fine I will remove the WIP flag, @Unknwon needs to add the .drone.sec and change how the docker images are getting built.

With that change we will have autobuilt binaries/archives for windows, linux, mac, freebsd, openbsd and netbsd in all available architectures.

Maybe the build from source section of the docs need some update as well as I have updated the Makefile.

tboerger:
Damn, still one issue with `make release`, it generates only binaries for openbsd/amd64, linux/amd64, darwin/amd64 and freebsd/amd64. Got to figure out why...

Unknwon:
:+1:

tboerger:
I think I will remove the gox stuff and will use the native go 1.5 crosscompile functionality.

Unknwon:
@tboerger Haha, welcome back.

tboerger:
Closed in favor of https://github.com/gogits/gogs/pull/2372

PR #2012, opened 2015-11-21
Open and close issue count always 0

andreynering:
(This pull request is not supposed to be merged, is just to show a issue).

I've noticed that the issue count is always 0 (at least with PostgreSQL, not sure about others).

I did a bit of investigation and found that the PG driver is returning an error on these 2 queries:

```
PANIC: pq: operator does not exist: boolean = integer
```

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -702,6 +702,12 @@ type IssueStatsOptions struct {
 	IsPull      bool
 }
 
+func checkErr(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
+
 // GetIssueStats returns issue statistic information by given conditions.
 func GetIssueStats(opts *IssueStatsOptions) *IssueStats {
 	stats := &IssueStats{}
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

So in PostgreSQL, should we use `is_xxx=true`?

andreynering:
I tried to fix it but couldn't.

Both `column = true` and `column = 't'` should work in PostgreSQL. I don't know why, but the driver seems to be not woking correctly with boolean columns. I tried `'t'` and `'f'` but it didn't work.

Unknwon:
Does the list is showing the correct results?

andreynering:
Yes, everything else is correct, including the total issue count in the sidebar:

Unknwon:
OK... so I think xorm handles it correctly.

Unknwon:
Can you check on `develop` again?

New release is coming today, good to fix this before it. :D

andreynering:
It's perfect. I didn't notice that the problem was in the other condition.

I would check for the error values. Don't need to panic, just print in the log. It would make debug easier.

Unknwon:
> I would check for the error values. Don't need to panic, just print in the log. It would make debug easier.

Make sense. Not sure which way is better, currently all `count`-related queries discard error check

PR #2024, opened 2015-11-23
Fixing Dropzone should accept all files when config is "*/*".

andreynering:


Unknwon:
Thanks! Can you confirm it fixes the problem you met?

andreynering:
Yes, I just tried.

PR #2039, opened 2015-11-26
API extensions

kiliit:
Some extensions on the API


kiliit:
Notice that the dependency from gogits/go-gogs-client is changed to kiliit/go-gogs-client

tboerger:
To avoid merge commits you should use rebase and force push your pull request ;)

cmd/web.go
```diff
@@ -29,7 +29,7 @@ import (
 	"gopkg.in/ini.v1"
 	"gopkg.in/macaron.v1"
 
-	api "github.com/gogits/go-gogs-client"
+	api "github.com/kiliit/go-gogs-client"
```
Unknwon: Maybe you want re-PR this after api client is merged. :sweat:

kiliit: I can make one more PR, rebase and force push if it makes you :sweat_smile:

Unknwon: Yes, please do that.

PR #2053, opened 2015-11-28
various fixes in gogs dump command

kakwa:
fixes includes:

* adding error handling on the target dump creation
* using a temporary directory to create the db and repositories dumps
* using the proper setting.<dir> instead of setting.workDir() for log/ and custom/
* adding few logs for tmpdir handling
* adding the name of the generated archive in  the final log

It fixes some weird behaviors where if the log/ and custom/ are not in a path relative to the gogs binary directory and if the command is not run from this directory, **gogs dump** will fail silently and create an empty archive.

### DIFF
--- a/cmd/dump.go
+++ b/cmd/dump.go
@@ -60,10 +60,18 @@ func runDump(ctx *cli.Context) {
 	}
 
 	workDir, _ := setting.WorkDir()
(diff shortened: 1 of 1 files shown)


kakwa:
Travis-ci seems to have failed on **go get** because of github timeouts. I would relaunch the job, but I can't.

Unknwon:
Thanks!

PR #2054, opened 2015-11-28
Admin should be able to see all organizations

nanoant:
This is follow-up for 56c66ee486b4b8d544201662de62a23f36e6a069 allowing admin
to see private repositories, even when not being member of them.

Without this patch, admin is effectively unable to edit any organization, so is not really an admin.

### DIFF
--- a/modules/middleware/org.go
+++ b/modules/middleware/org.go
@@ -48,7 +48,12 @@ func HandleOrgAssignment(ctx *Context, args ...bool) {
 		return
 	}
 
-	if ctx.IsSigned {
+	// Admin has super access.
+	if ctx.IsSigned && ctx.User.IsAdmin {
+		ctx.Org.IsOwner = true
+		ctx.Org.IsMember = true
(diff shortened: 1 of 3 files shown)


nanoant:
Closing to rebase to `develop` :cry:

tboerger:
Please don't open new pull requests just to rebase. For that usecase you should always force push the updated history to avoid useless noise.

nanoant:
@tboerger While your remark is generally true, but unfortunately does not work for GitHub, see very long and emotional discussion isaacs/github#18. Even if you rebase onto `develop` and force push, GitHub still want to merge the branch you created your PR, so `master` in our case.

The problem was few days ago Gog's `develop` == `master`, so even I was always making feature branched based on `develop`, this time it was based on `master` as well. Then I just pushed and didn't notice that GitHub defaulted `master` as base branch (because `develop` == `master`). So I had no other choice than close and make new PR. So again sorry for the noise.

PR #2064, opened 2015-11-30
Repo file list layout & misc fixes

nanoant:
This patch improves experience, making Octicons and repo file list look better. 

* Ensure Octicons are used with 16px font size
* UI: Mark top menu icons blue only when non-zero
* UI: Gray SHA1 next to message, file list collapse

I can understand that changing SHA1 label color to gray may be arguable, but I believe that this make the whole UI more balanced and less aggressive-looking. I believe that colorful labels should be reserved either for tags or issue types.

##### Before changes (current `devel`)

repo-file-list-layout-before

##### After changes

repo-file-list-layout-after

PR #2068, opened 2015-11-30
Repo file list layout & misc fixes

nanoant:
This is rebased to develop version of #2064 (thanks to GitHub for not allowing to simply changing PR base :imp:)

This patch improves experience, making Octicons and repo file list look better. 

* Ensure Octicons are used with 16px font size
* UI: Mark top menu icons blue only when non-zero
* UI: Gray SHA1 next to message, file list collapse

I can understand that changing SHA1 label color to gray may be arguable, but I believe that this make the whole UI more balanced and less aggressive-looking. I believe that colorful labels should be reserved either for tags or issue types.

See below for latest screenshots.

### DIFF
--- a/public/less/_octicons.less
+++ b/public/less/_octicons.less
@@ -14,7 +14,9 @@
 
 // .octicon is optimized for 16px.
 // .mega-octicon is optimized for 32px but can be used larger.
-.octicon, .mega-octicon {
+.octicon, .mega-octicon,
+// ensure Semantic UI .icon of 14px does not apply to .icon.octicon:
+.icon.octicon, .icon.mega-octicon {
   font: normal normal normal 16px/1 octicons;
   display: inline-block;
   text-decoration: none;
--- a/public/css/gogs.css
+++ b/public/css/gogs.css
@@ -5,7 +5,9 @@
   font-style: normal;
 }
 .octicon,
-.mega-octicon {
+.mega-octicon,
+.icon.octicon,
+.icon.mega-octicon {
   font: normal normal normal 16px/1 octicons;
   display: inline-block;
   text-decoration: none;
--- a/public/css/gogs.css
+++ b/public/css/gogs.css
@@ -1857,40 +1857,16 @@ footer .container .links > *:first-child {
 .repository.file.list .head.meta li .ui.breadcrumb a {
   font-size: 16px;
 }
-.repository.file.list #repo-files-table .table.list {
-  width: 80% !important;
-}
 .repository.file.list #repo-files-table thead th {
   padding-top: 8px;
   padding-bottom: 5px;
   font-weight: normal;
 }
-.repository.file.list #repo-files-table thead th #last-commit-message {
-  margin-left: 5px;
-  margin-bottom: -4px;
-  width: 400px;
-}
-.repository.file.list #repo-files-table thead th .age {
-  margin-top: 2px;
-}
 .repository.file.list #repo-files-table thead .ui.avatar {
   margin-bottom: 5px;
 }
 .repository.file.list #repo-files-table tbody .icon {
-  margin-left: 5px;
-}
-.repository.file.list #repo-files-table tbody .name {
-  max-width: 120px;
-}
-.repository.file.list #repo-files-table tbody .message {
-  max-width: 300px;
-}
-.repository.file.list #repo-files-table tbody .age {
-  min-width: 150px;
-}
-.repository.file.list #repo-files-table tbody .text.truncate {
-  margin-bottom: -5px;
-  max-width: 100%;
+  margin-right: 5px;
 }
 .repository.file.list #repo-files-table td {
   padding-top: 8px;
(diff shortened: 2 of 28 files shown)


nanoant:
@Unknwon I reworked this PR to rely completely on `fixed single line table` from SUI which adds ellipsis when necessary, and now the file list looks following and quite narrow screen (note the ellipsis)

zrzut ekranu 2015-12-01 o 00 45 21

Effectively many CSS entries were not needed anymore since SUI handles is just well.

Unknwon:
Looks awesome!

How about change SHA background color to green?

nanoant:
@Unknwon As I noted in c6ae28a (UI: Use grey SHA1 labels rather distracting green) I find the green SHA labels very distracting. But of course I can remove this commit from this PR and move to some other if you are willing to accept the rest of the changes.

Unknwon:
You're right, hmm, kind of used to that green, maybe that's reason. 

Ok, let me review the changes.

Unknwon:
Found one thing Gogs does not change .css files directly, but .less files.

Unknwon:
PS1: there are conflicts.
PS2: you can use `scripts/less.sh` to generate less to css.

Depado:
I also find the green SHA1 quite distracting. It makes the SHA1 something really important compared to other information present on the page.

nanoant:
@Unknwon Conflicts resolved, also aside of improvements and simplification of ellipsis handling I revamped a bit latest design of repo main menu bar, I think you will like it.

NOTE: I removed commit changing tags to grey, I think we can make it as separate PR and have discussion with community.

#### New design & layout fixes in this PR

repo-file-list-layout-new

#### Current `develop` branch design

repo-file-list-layout-develop

Unknwon:
Thanks! Looks awesome, one little thing, I don't think we need to align table header with its body.

Unknwon:
The old way looks more nature, what do you think?

image

nanoant:
> The old way looks more nature, what do you think?

There are two issues with old approach:

1. It uses colspan in 1st row, that doesn't play well with table fixed layout that's necessary for automatic ellipsis (this is denoted in the commit message)

2. It uses non-labeled SHA1 which is inconsistent with rest of the UI, IMHO we should stick to single way of displaying SHA1s

But I generally agree that the "last commit" header needs some love. I can promise to give 2nd look and find some better approach to show it. Either as separate PR or reworking this one if you don't want to merge it yet.

Unknwon:
I think I can accept to make it non-green (the default grey) , so in first look we can tell it's not part of the table body?

nanoant:
@Unknwon Okay, I think I found a workaround, see last commit and sshot:

zrzut ekranu 2015-12-03 o 01 18 54

Unknwon:
Looks awesome. Let me review everything again just in case.

nanoant:
@Unknwon Sure thing, take your time.

public/less/_octicons.less
```diff
@@ -14,7 +14,9 @@
 
 // .octicon is optimized for 16px.
 // .mega-octicon is optimized for 32px but can be used larger.
-.octicon, .mega-octicon {
+.octicon, .mega-octicon,
+// ensure Semantic UI .icon of 14px does not apply to .icon.octicon:
+.icon.octicon, .icon.mega-octicon {
```
Unknwon: You may want to reenforce this section in `_base.less` because this file will be updated when new octicon is released.

Unknwon:
I think the clone section should be put in the old location, because repo description can be easily fit into multiple lines, but the breadcrumb does not.

image

nanoant:
> I think the clone section should be put in the old location, because repo description can be easily fit into multiple lines, but the breadcrumb does not.

@Unknwon I disagree. 1st it takes space for the description that can be longer, 2nd there's an empty space there and feels more balanced, 3rd when you are viewing a file or subdirectory the clone section isn't shown with my PR.

Also the change fixes a problem where clone section was going off the screen on narrow window.

Anyway this is how GitHub looks. And I think it is way it should.

Unknwon:
So how do you handle `root/level1/level2/level3/level4/level5/level6`?

Unknwon:
image

As you can tell, the new location can't even handle `level3`, that's too short.

nanoant:
> So how do you handle root/level1/level2/level3/level4/level5/level6?

@Unknwon I told you, clone section is shown only on root. It is written in commit message too.

Unknwon:
> @Unknwon I told you, clone section is shown only on root. It is written in commit message too.

That's what I thought. :wink:

LGTM.

public/less/_octicons.less
nanoant: PR Updated.

Unknwon:
Found a small Ui problem: https://try.gogs.io/gogs/gogs/commit/dcb391d3415a84ec893d11b788f776ea89c301c3

Need some padding:

image

PR #2069, opened 2015-11-30
Admin should be able to see all organizations

nanoant:
This is follow-up for 56c66ee486b4b8d544201662de62a23f36e6a069 allowing admin
to see private repositories, even when not being member of them.

Without this patch, admin is effectively unable to edit any organization, so is not really an admin.

(this is rebased to `develop` PR #2054)


Unknwon:
Thank you very much!

PR #2067, opened 2015-11-30
[WIP] Import command (GitLab, Redmine, etc.)

nanoant:
This is work-in-progress stuff. It is not intended to be merged now, or ever. This is experimental code, but I already managed to migrate medium GitLab installation to Gogs using this. Comments welcome.

### GitLab import
[patch]: https://gist.github.com/nanoant/f6f3e2f2fe465f371c6b

- [x] Users import (including password hash import - requires [small patch to GitLab][patch])
- [x] User SSH keys import
- [x] Organizations import (including organization members and permissions)
- [x] Projects import (including public/private visibility and project members aka collaborators)
- [ ] Issues import
- [ ] Wiki import

### Redmine import

- [ ] Users import
- [ ] Groups import
- [ ] Projects import
- [ ] Issues import
- [ ] Wiki import

PR #2070, opened 2015-11-30
[WIP] Import command (GitLab, Redmine, etc.)

nanoant:
This is work-in-progress stuff. It is not intended to be merged now, or ever. This is experimental code, but I already managed to migrate medium GitLab installation to Gogs using this. Comments welcome.

### GitLab import
[patch]: https://gist.github.com/nanoant/f6f3e2f2fe465f371c6b

- [x] Users import (including password hash import - requires [small patch to GitLab][patch])
- [x] User SSH keys import
- [x] Organizations import (including organization members and permissions)
- [x] Projects import (including public/private visibility and project members aka collaborators)
- [ ] Issues import
- [ ] Wiki import

### Redmine import

- [ ] Users import
- [ ] Groups import
- [ ] Projects import
- [ ] Issues import
- [ ] Wiki import

(this is rebased to `develop` PR #2067)

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -30,6 +30,15 @@ This can be used to migrate from other service to Gogs.`,
 				stringFlag("token", "", "private authentication token"),
 			},
 		},
+		{
+			Name:   "redmine",
+			Usage:  "import from Redmine",
+			Action: importGitLab,
+			Flags: []cli.Flag{
+				stringFlag("url", "", "Redmine service base URL"),
+				stringFlag("token", "", "private authentication token"),
+			},
+		},
 	},
 	Flags: []cli.Flag{
 		stringFlag("config, c", "custom/conf/app.ini", "Custom configuration file path"),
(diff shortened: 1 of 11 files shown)


Unknwon:
Thanks your PR!

However, I really believe import/migrate thing should be a separate repository other than in the main repo.

nanoant:
@Unknwon I agree that the proper solution should live in its own project and not play around with Gogs internals. The reason I did it as is was:

1. I needed to make it working ASAP to migrate existing site
2. Doing it a simple client for 2 REST services GitLab <-> Gogs adds extra layer of complication and errors handling
3. We need (as suggested) to have some extra APIs present

Overall if I had to do it the right way it would took me 3⨉ longer. But until we get a proper solution please keep this open (and do not merge of course), because anyways there's no other way to import GitLab currently :smiling_imp:

Unknwon:
Sure, make sense. But this will eventually become an independent project and PR will be closed.

PR #2078, opened 2015-12-01
Makefile improvements

nanoant:
Long story short - compile time improvement, automatic build of CSS & bindata reducing the risk of forgetting to recompile them. See the commits for details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,10 @@
 LDFLAGS += -X "github.com/gogits/gogs/modules/setting.BuildTime=$(shell date -u '+%Y-%m-%d %I:%M:%S %Z')"
 LDFLAGS += -X "github.com/gogits/gogs/modules/setting.BuildGitHash=$(shell git rev-parse HEAD)"
 
+DATA_FILES := $(shell find conf | sed 's/ /\\ /g')
+LESS_FILES := $(wildcard public/less/gogs.less public/less/_*.less)
+GENERATED  := modules/bindata/bindata.go public/css/gogs.css
+
 TAGS = ""
 
 RELEASE_ROOT = "release"
@@ -9,7 +13,7 @@ NOW = $(shell date -u '+%Y%m%d%I%M%S')
 
 .PHONY: build pack release bindata clean 
 
-build:
+build: $(GENERATED)
 	go install -ldflags '$(LDFLAGS)' -tags '$(TAGS)'
 	cp '$(GOPATH)/bin/gogs' .
 
@@ -25,8 +29,15 @@ pack:
 
 release: build pack
 
-bindata: 
-	go-bindata -o=modules/bindata/bindata.go -ignore="\\.DS_Store|README.md" -pkg=bindata conf/...
+bindata: modules/bindata/bindata.go
+
+modules/bindata/bindata.go: $(DATA_FILES)
+	go-bindata -o=$@ -ignore="\\.DS_Store|README.md" -pkg=bindata conf/...
+
+less: public/css/gogs.css
+
+public/css/gogs.css: $(LESS_FILES)
+	lessc $< $@
 
 clean:
 	go clean -i ./...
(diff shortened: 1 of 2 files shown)


Unknwon:
No, generated content should always be part of main repo.

tboerger:
I already did a massive makefile refactoring within the droneci pr :)

nanoant:
@Unknwon okay, updated, this time without last commit removing bindata from repo

scripts/less.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
```
Unknwon: Why remove this?

Makefile
```diff
@@ -25,11 +29,18 @@ pack:
...
-	go-bindata -o=modules/bindata/bindata.go -ignore="\\.DS_Store|README.md" -pkg=bindata conf/...
+bindata: modules/bindata/bindata.go
+
+modules/bindata/bindata.go: $(DATA_FILES)
+	go-bindata -o=$@ -ignore="\\.DS_Store|README.md" -pkg=bindata conf/...
+
+less: public/css/gogs.css
+
+public/css/gogs.css: $(LESS_FILES)
+	lessc $< $@
```
Unknwon: Does this really work without giving `../public/less/gogs.less`? .. just to confirm.

Makefile
```diff
@@ -25,11 +29,18 @@ pack:
 
 release: build pack
 
-bindata: 
-	go-bindata -o=modules/bindata/bindata.go -ignore="\\.DS_Store|README.md" -pkg=bindata conf/...
+bindata: modules/bindata/bindata.go
+
+modules/bindata/bindata.go: $(DATA_FILES)
```
Unknwon: Why add this command?

Makefile
nanoant: First element of `$(LESS_FILES)` is `public/less/gogs.less` so it doesn't need to be given explicitly.

Makefile
```diff
@@ -1,17 +1,21 @@
 LDFLAGS += -X "github.com/gogits/gogs/modules/setting.BuildTime=$(shell date -u '+%Y-%m-%d %I:%M:%S %Z')"
 LDFLAGS += -X "github.com/gogits/gogs/modules/setting.BuildGitHash=$(shell git rev-parse HEAD)"
 
+DATA_FILES := $(shell find conf | sed 's/ /\\ /g')
+LESS_FILES := $(wildcard public/less/gogs.less public/less/_*.less)
```
Unknwon: gogs.less already imports other .less, is `public/less/_*.less` part really needed?

Makefile
Unknwon: Thanks, I updated my original comment.

nanoant: > Why use two commands other than combine to single one?

1st one is just an convenience alias to use `make less` via command line. 2nd is doing actual work and necessary to track dependencies and trigger rebuild.

Makefile
```diff
@@ -1,17 +1,21 @@
...
+GENERATED  := modules/bindata/bindata.go public/css/gogs.css
```
Unknwon: Is this needed?

Makefile
Unknwon: How does it trigger rebuild?

Unknwon:
I think all we need for this PR is the improvement of use `make less` to replace `less.sh`.

Makefile
```diff
@@ -25,11 +29,18 @@ pack:
...
 release: build pack
 
-bindata: 
-	go-bindata -o=modules/bindata/bindata.go -ignore="\\.DS_Store|README.md" -pkg=bindata conf/...
+bindata: modules/bindata/bindata.go
+
+modules/bindata/bindata.go: $(DATA_FILES)
+	go-bindata -o=$@ -ignore="\\.DS_Store|README.md" -pkg=bindata conf/...
+
+less: public/css/gogs.css
```
Unknwon: Why not just `lessc public/less/gogs.less public/css/gogs.css`?

Makefile
Unknwon: Same as below, why not just use one? Is this some high-tech on make command (very little knowledge on makefile)?

Makefile
Unknwon: Follow comments below, should this be changed to `bindata less`？

Unknwon:
LGTM.

PR #2079, opened 2015-12-01
LDAP: Optional user name attribute specification

nanoant:
Consider following LDAP search query example:

    (&(objectClass=Person)(|(uid=%s)(mail=%s)))

Right now on first login attempt Gogs will use the text supplied on login form
as the newly created user name. In example query above the text matches against
both e-mail or user name. So if user puts the e-mail then the new Gogs user
name will be e-mail which may be undesired.

Using optional user name attribute setting we can explicitly say we want Gogs
user name to be certain LDAP attribute eg. `uid`, so even user will use e-mail
to login 1st time, the new account will receive correct user name.


models/login.go
```diff
@@ -242,17 +242,30 @@ func LoginUserLDAPSource(u *User, name, passwd string, source *LoginSource, auto
 	}
 
 	// Fallback.
+	if len(name) == 0 {
+		name = loginName
+	}
 	if len(mail) == 0 {
 		mail = fmt.Sprintf("%s@localhost", name)
 	}
+	var fullName string
```
Unknwon: I think it would be better to put into a function like `composeFullName(fn,sn) string`

nanoant: @Unknwon agreed, I like you approach of having clean reusable code! Fixing that.

models/login.go
```diff
@@ -260,6 +263,18 @@ func LoginUserLDAPSource(u *User, name, passwd string, source *LoginSource, auto
 	return u, CreateUser(u)
 }
 
+func composeFullName(firstName, surename, userName string) string {
```
Unknwon: As you return in every case branch, you do not need `else if` anymore.

Unknwon:
LGTM.

wqweto:
@nanoant FYI, search query has to use fmt's `explicit argument indexes`  to make it work
```
(&(objectClass=Person)(|(uid=%[1]s)(mail=%[1]s)))
```

nanoant:
@wqweto Oh.. thanks, completely forgot about that one.

PR #2086, opened 2015-12-02
Translation: Add missing entry for new "code" tab

nanoant:
We have new tab, but we had no entry. That's why it was showing "code"
(lowercase) as this is text id, where we were expecting properly title cased
"Code" to be shown in English version.

code-missing-translation

Also add Polish translation "code=Kod".

@Unknwon Btw. I am not sure whether I should add `bindata.go` together with the PR or not. Suggestions?


Unknwon:
Thanks, just a note that in the future, you only need to PR English version, others will be overwritten anyway.

PR #2094, opened 2015-12-03
UI: Use more subtle grey SHA1 labels

nanoant:
Current green SHA1 labels are more pronounced than other UI elements attracting attention as if they were most important thing in the UI, while they are not as important, especially without real Git client.

Using grey SHA1 labels makes the UI more balanced, less aggressive and lets user to focus on other content elements.

##### Current

current
current

##### Proposed

proposed
proposed

NOTE: Neither GitHub or Bitbucket uses so heavy pronunciation as Gogs. (they are also not using SHAs on file list, that's why showing commit lists below)

##### Git

github

##### Bitbucket

bitbucket

This was previously proposed by #2068 and supported by @Depado.

## Discussion

I know there are supporters of current green SHA1 label look, and many users that are used to them. But they are also ppl that may find them distracting and actually repellent, that is why I humbly ask for concrete arguments behind having green labels and for keeping current design. Thank you.

### DIFF
--- a/public/less/_repository.less
+++ b/public/less/_repository.less
@@ -602,6 +602,9 @@
 		    padding: 6px 40px 4px 35px;
 			}
 		}
+		&.ui.basic.striped.table tbody tr:nth-child(2n) {
+			background-color: rgba(0, 0, 0, .02)!important;
+		}
 	}
 
 	.diff-detail-box {
--- a/public/css/gogs.css
+++ b/public/css/gogs.css
@@ -2243,6 +2243,9 @@ footer .container .links > *:first-child {
   font-size: 13px;
   padding: 6px 40px 4px 35px;
 }
+.repository #commits-table.ui.basic.striped.table tbody tr:nth-child(2n) {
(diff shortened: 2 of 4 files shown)


rnuske:
I strongly support the subtle grey SHA1 labels. The green is extremely distracting.

public/less/_repository.less
```diff
@@ -602,6 +602,9 @@
 		    padding: 6px 40px 4px 35px;
 			}
 		}
+		&.ui.basic.striped.table tbody tr:nth-child(2n) {
```
Unknwon: Any reason why need this rule? I thought remove `green` class is enough?

nanoant: @Unknwon please see reasoning in 70a1cbaf90aa6b13b163d07aeb962858e2bd4903 message, the problem is that grey labels mix with grey strips background, that's why. Alternatively I'd remove the strips altogether.

Unknwon: Make sense.

Unknwon:
LGTM. Thank you!

PR #2110, opened 2015-12-06
Fix misspelled words

msoedov:
Few words found misspelled throughout the codebase by my strict IDE settings. There are should be no functional changes.


Unknwon:
haha, thank you!

PR #2112, opened 2015-12-06
Render commit msg as header + verbatim description

nanoant:
Most commit in Git are expected to follow standard of single header line,
followed by description paragraphs, separated by empty line from previous block.

Previously Gogs were treating everything as single header. Now we are trying to
render only first line as header, but following lines (description chunks) as a
verbatim.

#### Before (current `develop`)

before

#### After (this PR)

after

### DIFF
--- a/modules/template/template.go
+++ b/modules/template/template.go
@@ -183,15 +183,27 @@ func ReplaceLeft(s, old, new string) string {
 }
 
 // RenderCommitMessage renders commit message with XSS-safe and special links.
-func RenderCommitMessage(msg, urlPrefix string, metas map[string]string) template.HTML {
+func RenderCommitMessage(full bool, msg, urlPrefix string, metas map[string]string) template.HTML {
 	cleanMsg := template.HTMLEscapeString(msg)
 	fullMessage := string(base.RenderIssueIndexPattern([]byte(cleanMsg), urlPrefix, metas))
 	msgLines := strings.Split(strings.TrimSpace(fullMessage), "\n")
-	for i := range msgLines {
-		msgLines[i] = ReplaceLeft(msgLines[i], " ", "&nbsp;")
+	numLines := len(msgLines)
+	if numLines == 0 {
+		return template.HTML("")
+	} else if !full {
+		return template.HTML(msgLines[0])
+	} else if numLines == 1 || (numLines >= 2 && len(msgLines[1]) == 0) {
+		// First line is a header, standalone or followed by empty line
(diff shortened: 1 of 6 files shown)


modules/template/template.go
```diff
@@ -183,15 +183,30 @@ func ReplaceLeft(s, old, new string) string {
...
 // RenderCommitMessage renders commit message with XSS-safe and special links.
-func RenderCommitMessage(msg, urlPrefix string, metas map[string]string) template.HTML {
+func RenderCommitMessage(full bool, msg, urlPrefix string, metas map[string]string) template.HTML {
 	cleanMsg := template.HTMLEscapeString(msg)
 	fullMessage := string(base.RenderIssueIndexPattern([]byte(cleanMsg), urlPrefix, metas))
 	msgLines := strings.Split(strings.TrimSpace(fullMessage), "\n")
-	for i := range msgLines {
-		msgLines[i] = ReplaceLeft(msgLines[i], " ", "&nbsp;")
+	header := ""
+	noLines := len(msgLines)
```
Unknwon: Maybe use `numLines`? `noLines` first thought was no line...

nanoant: Makes sense, will change that.

modules/template/template.go
```diff
@@ -183,15 +183,30 @@ func ReplaceLeft(s, old, new string) string {
...
+	if noLines == 0 {
+		return template.HTML("")
+	} else if !full {
+		return template.HTML(msgLines[0])
+	} else if noLines == 1 || (noLines >= 2 && msgLines[1] == "") {
```
Unknwon: Convention in Gogs for `msgLines[1] == ""` is `len(msgLines[1])==0`.

Unknwon:
Can you show a screenshot for two-line commit without having an empty in between?

nanoant:
> Can you show a screenshot for two-line commit without having an empty in between?

@Unknwon This is how it looks - it could be improved, but this is not the way anyone should format Git commit.

zrzut ekranu 2015-12-07 o 01 13 55

Unknwon:
Thanks! So in this case, I think the full commit should be rendered as header.

nanoant:
> Thanks! So in this case, I think the full commit message should be rendered as header.

Done, there are now 3 cases handled:

#### One-line

one-line

#### Multi-line (Git standard)

multi-line

#### Non-standard (no empty line after 1st line)

non-standard

Unknwon:
LGTM.

PR #2120, opened 2015-12-07
TextMate project file

nanoant:
TextMate 2 project properties file

1. Defining code style
2. Describing which files should not be searched (binaries)
3. Automatically sets the spelling language for translations

Also adds makefile "run" phony target required to be able to simply <kbd>⌘R</kbd> to run "gogs web" within TextMate 2.

### DIFF
--- a/.tm_properties
+++ b/.tm_properties
@@ -1,0 +1,41 @@
+# TextMate 2 project file
+#
+# Copyright 2015 The Gogs Authors. All rights reserved.
+# Use of this source code is governed by a MIT-style
+# license that can be found in the LICENSE file.
+
+# build with sqlite
+TM_MAKE_FLAGS = 'TAGS=sqlite'
+
+# exclude binary data from search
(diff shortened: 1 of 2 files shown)


tboerger:
I would prefer a editorconfig which is much more generic instead of a textmate config.

nanoant:
@tboerger The textmate project file configures far more than editorconfig, eg. spelling language for translations, binary files, hidden files that shouldn't be hidden. That's why I am adding this PR.
Moreover EditorConfig bundle is not installed in TM by default.

tboerger:
But this will result in a big mess if every contributor adds his editor config. In my opinions this is totally developer specific and not really project repository related. Only a pretty small amount of developers is using textmate compared to vi, emacs, atom or sublime.

nanoant:
@tboerger Okay my intention isn't to make anyone happy whether they like it or not. Therefore I am closing that one.

PR #2121, opened 2015-12-07
Consistent tab indentation for all source files

nanoant:
This is hygiene PR. I have changed semi-automatically indent in all `.less` and `.tmpl` files to use tabs, same as for `.go` files. Also I have indented all `.tmpl` files consistently:

1. Indents both HTML containers such as `<div>`, `<p>` and Go HTML template blocks such as `{{if}}` `{{with}}`
2. Cleans all trailing white-space
3. Adds trailing last line-break to each file

This change only changes indention not content, so `git blame -w ...` still works well (where `-w` means ignore whitespace).

This may help Gogs opening for more contributors.


Unknwon:
Hi, thanks your work!

But the diff is too large and GitHub does not allow me to review it and I've spotted some incorrections.

And `.go` files are not needed for this process as you should always use `gofmt`.

So why don't we separate it into two steps, first `.less` files and them `.tmpl` files?

nanoant:
> But the diff is too large and GitHub does not allow me to review it and I've spotted some incorrections.

Click on 2 individual commits, 1st for `.less`, 2nd for `.tmpl`.

> And .go files are not needed for this process as you should always use gofmt.

I didn't touch `.go` files.

> So why don't we separate it into two steps (PRs), first .less files and them .tmpl files?

Both are separated into 2 commits, so changes are independent.

Unknwon:
OK... Just to make sure that everything is changed to `1 tab`?

nanoant:
> OK... Just to make sure that everything is changed to 1 tab?

Yes, single indent == 1 tab for all `.less` & `.tmpl` files.

(I simply followed Go indent that uses tabs, also `.editorconfig` file suggest that tabs should be used exclusively.)

Unknwon:
Thank you very much!

PR #2122, opened 2015-12-08
Allow pre-receive hook customization

nanoant:
This hook can be used for example to reject too large commits and it is
executed before "update" hook, used exclusively by Gogs to update its state.

https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks


Unknwon:
Thanks!

PR #2125, opened 2015-12-09
Only show comparison link for >2 commits, fixes #1110

### LINKED ISSUE #1110: "View comparison for these 2 commits"
Dashboard feed always says "2 commits" regardless of many actual commits there are when you push.

There might be 1 commit, there might be 6, it will still say "2 commits".

2015-03-21-1353q5dqy

angus-g:
We can look at the PushCommits object to see how many commits were included in a commit, and add some template logic to only show the comparison link when there are at least 2 commits in a push. We also correct the link to display the number of commits.


Unknwon:
Thanks your PR!

Can you give corresponding screenshots?

angus-g:
1. Only 1 commit: 1commit
2. 2 commits: 2commit
3. 3 commits: 3commit

Unknwon:
LGTM.

Issue #1846, opened 2015-10-27
Select branch in commits page


hhenkel:
![gogs_ go git service](https://cloud.githubusercontent.com/assets/2813782/10754071/7d907a4c-7c95-11e5-9daa-7e693e7c7577.png)

Would be good to show the current branch that the commits belong to. Maybe a simple line under the repository name or like in github a dropdown box?

ChangePlaces:
I think this is a good idea - but for now, you can always look in the url ;)

ChangePlaces:
#1153 duplicate?

Unknwon:
@ChangePlaces not exactly.

hhenkel:
@ChangePlaces saw that issue but had the impression that it is a bit of a different thing.

Unknwon:
I think it has been fixed?

Losing idea of the main purpose of this issue.

hhenkel:
![gogs_commits](https://cloud.githubusercontent.com/assets/2813782/11208510/e82f48ce-8d1b-11e5-8ff2-d91f115e42c2.png)

It was mostly a request to have something similar dropdown as github offers or a simple text field showing which branch is shown.

hhenkel:
Okay, seems like #1572 is more or less what I was looking for. Nonetheless I would vote for showing the dropdown that is visible in the repo overview in gogs as it is more consistent and would also allow switching the branch that is shown.

Unknwon:
I got it!

So how about the dropdown here:

![image](https://cloud.githubusercontent.com/assets/2946214/11220492/4e715c46-8d2e-11e5-99d3-36e31810b308.png)

The one in the repository home page.

hhenkel:
@Unknwon yes, that was what I had in mind.

Unknwon:
This should be implemented by merging https://github.com/gogits/gogs/pull/2126

hhenkel:
@Unknwon Works as expected on try.gogs.io, going to close the issue.

Unknwon:
Awesome!

PR #2126, opened 2015-12-09
Dropdown on commits page to choose branch #1846

angus-g:
We break the branch-chooser dropdown from the repository homepage into its own template and also call it from the commits page to allow for the selection of a branch/tag to view.


Unknwon:
LGTM. Thanks!

PR #2137, opened 2015-12-09
LOCAL_ROOT_URL for workers accessing web service 

nanoant:
Local (DMZ) URL for gogs workers (such as ssh update) accessing web service. In
most cases you do not need to change default http://localhost:HTTP_PORT/. You
may need to alter it only if your ssh server node is not the same as http node,
eg. running behind proxy on different node than web server.

                     --- 80 public port -> 8080 -- web server node
                    /
    public proxy --<
                    \
                     --- 22 public port -> 10022 -- ssh server node

This option is not intended to be accessible via web GUI settings, since it is
unlikely someone needs to change it to somethings else than default
http://localhost:HTTP_PORT/ which should work for most of the cases.

But this should land into the documentation somewhere.

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -46,6 +46,10 @@ DOMAIN = localhost
 ROOT_URL = %(PROTOCOL)s://%(DOMAIN)s:%(HTTP_PORT)s/
 HTTP_ADDR =
 HTTP_PORT = 3000
+; Local (DMZ) URL for Gogs workers (such as SSH update) accessing web service.
+; In most cases you do not need to change the default value.
+; Alter it only if your SSH server node is not the same as HTTP node.
+LOCAL_ROOT_URL = http://localhost:%(HTTP_PORT)s/
 ; Disable SSH feature when not available
 DISABLE_SSH = false
 ; Whether use builtin SSH server or not.
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -109,7 +109,7 @@ func handleUpdateTask(uuid string, user *models.User, username, reponame string,
 	}
 
 	// Ask for running deliver hook and test pull request tasks.
-	reqURL := setting.AppUrl + username + "/" + reponame + "/tasks/trigger?branch=" +
+	reqURL := setting.LocalUrl + username + "/" + reponame + "/tasks/trigger?branch=" +
 		strings.TrimPrefix(task.RefName, "refs/heads/")
 	log.GitLogger.Trace("Trigger task: %s", reqURL)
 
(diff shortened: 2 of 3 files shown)


conf/app.ini
```diff
@@ -46,6 +46,10 @@ DOMAIN = localhost
 ROOT_URL = %(PROTOCOL)s://%(DOMAIN)s:%(HTTP_PORT)s/
 HTTP_ADDR =
 HTTP_PORT = 3000
+; Local (DMZ) URL for gogs workers (such as ssh update) accessing web service.
+; In most cases you may leave it empty. Defaults: http://localhost:HTTP_PORT/
+; You may need to set it only if ssh server node is not the same as http node.
+LOCAL_ROOT_URL =
```
Unknwon: I believe you can use `LOCAL_ROOT_URL  = http://localhost:%(HTTP_PORT)s/` here.

nanoant: I didn't want to on purpose. It is better to not expose LOCAL_ROOT_URL in `custom/conf/app.ini` at all as for most users this would be irrelevant.
Doing this as you suggest we risk some rookie user changing `HTTP_PORT` but forgetting about port in `LOCAL_ROOT_URL` :)

Unknwon: INI parser saves as `LOCAL_ROOT_URL = http://localhost:%(HTTP_PORT)s/`, so rookie user does not need to care `LOCAL_ROOT_URL` for changing `HTTP_PORT`.

nanoant: Okay, I need some clarification what's the purpose of this template? `LOCAL_ROOT_URL = http://localhost:%(HTTP_PORT)s/` does it land into `custom/conf/app.ini`? Who and what actually parses `http://localhost:%(HTTP_PORT)s/` and replaces `%(HTTP_PORT)s` with the value. Finally doesn't the `MustString("....")` serve the same purpose?

Unknwon: It just a INI syntax for recursion values, the value you get for `LOCAL_ROOT_URL` is changing with the change of `HTTP_PORT`. 

It does not show in `custom/conf/app.ini` if user does not touch it.

nanoant: So the order of loading values is:

`custom/conf/app.ini` --if no value--> builtin `conf/app.ini` (evaluated template) --if no value--> `MustString default string` ?

If so then all values in MustString are completely ignored, because anyway Gogs falls back to defaults from built-in `conf/app.ini`. Right?

Unknwon: It is not like `if no value`, it overwrites, the loading order is, `conf/app.ini`, `custom/conf/app.ini`, `Mustxxx`, but basically the result is same.

Unknwon:
LGTM.

PR #2138, opened 2015-12-09
LDAP parameters UI: bind_dn and bind_password are not required

SlavikZ:
LDAP via BindDN bind_dn and bind_password parameters are optional. They may be left blank to perform an anonymous search.


Unknwon:
Thanks!

PR #2142, opened 2015-12-10
Fix postinstall to use GOGS_CUSTOM instead of symlinking

### LINKED ISSUE #1897: Install page comes up nearly every day
Nearly every day when I navigate to my gogs service, the install page comes up again, i think this happens after every update.
I installed gogs with the packager.io package.

crohr:
This should fix #1897 by settings `GOGS_CUSTOM` to `/etc/gogs`.


crohr:
Wrong branch, closing.

PR #2169, opened 2015-12-13
Ignore invalid env for SSH Server (OSX fix)

bclermont:
I was trying to use built-in ssh server from an OSX client and I couldn't push, it didn't even log an error. I added some debug code and found that it was unexpected ``env`` pushed.

```
2015/12/13 06:04:33 [I] [env]: payload LC_CTYPEUTF-8' replaced LC_CTYPEUTF-8' args: '[]string{"\bLC_CTYPE\x05UTF-8"}'
2015/12/13 06:04:33 [...s/modules/ssh/ssh.go:62 func·001()] [E] len(args) 1 != 2
```

I found the reason:

SSH: `SSH-2.0-OpenSSH_6.9`
OS: OSX El Capitan

`/etc/ssh/ssh_config` file from OSX (I guess, I didn't put that there myself):

```
 Host *
   SendEnv LANG LC_*
```

This pull request just log invalid env arguments instead of silently return.

### DIFF
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -53,7 +53,8 @@ func handleServerConn(keyID string, chans <-chan ssh.NewChannel) {
 				case "env":
 					args := strings.Split(strings.Replace(payload, "\x00", "", -1), "\v")
 					if len(args) != 2 {
-						return
+						log.Warn("Invalid env arguments: '%#v'", args)
+						continue
 					}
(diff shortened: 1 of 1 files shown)


bclermont:
this time target is `develop`

Unknwon:
Thanks, but does not seems like the way I commented?

Unknwon:
You can put `log.Warn` in early return, to avoid a big indention for the other case branch,

Unknwon:
LGTM.

PR #2257, opened 2015-12-20
Wiki pages containing question marks in their name weren't loading

Jofkos:
(untested)

### DIFF
--- a/models/wiki.go
+++ b/models/wiki.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"sync"
 
+	"net/url"
+
 	"github.com/Unknwon/com"
 
 	"github.com/gogits/git-module"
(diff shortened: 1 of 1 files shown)


models/wiki.go
```diff
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"sync"
```
Unknwon: Please remove this empty line.

Unknwon:
Thanks your PR.

But build is failling.

Jofkos:
Updated, but i still don't know if it is working.

Unknwon:
OK... you have to make sure it did fix the problem you have.

Jofkos:
I'm not able to build it, but if it's not the solution you'd like to have to fix that, it would be nice to somehow add support for page titles with question marks, or to forbid those

Unknwon:
Oh... that's too bad... Let me see...

Jofkos:
I was able to build it yesterday and i figured out that the solution works partially. With the changes you're able to open pages with names containing question marks, but you're not able to edit those.
I wasn't able to figure out why so i hope you will ^^

Unknwon:
OK, thank you!

PR #2262, opened 2015-12-22
Add default for NumFollowing field (fixes #2261)

### LINKED ISSUE #2261: Fail to initialize ORM engine
I did an update from gogs 8.10 to 8.13. After copying my custom and data folder in to the gogs folder I got an error message:

    [...s/routers/install.go:70 GlobalInit()] [E] Fail to initialize ORM engine: sync database struct error: Cannot add a NOT NULL column with default value NULL

I am using gogs with sqlite

angus-g:
We set the default value for the non-NULL field `NumFollowing` of the User model to 0, which stops an error when the ORM tries to sync.

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -91,7 +91,7 @@ type User struct {
 
 	// Counters
 	NumFollowers int
-	NumFollowing int `xorm:"NOT NULL"`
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

PR #2274, opened 2015-12-23
added docker-compose file with mariadb

rothgar:
I have a basic docker-compose file but I have two problems that I am not sure if they are bugs with gogs or something I missed.

1. Even though I specify environment variables for the database the initial setup screen is still shown and the variables I provide are not filled in. It's easy to add the values but probably should be done automatically based on the environment
2. I link the containers but it does not appear the application looks in /etc/hosts file where the link is stored. I verified the right values were in /etc/hosts but I would still get an error trying to link the db by name (tried db docker_db docker_db_1 etc) It works if I use 127.0.0.1 or the ip address of the db container but not hostname.

This could also be extended to support other databases and/or data containers but I figured that would be overkill for someone just wanting to try out gogs. I wanted to provide sane defaults with slightly better performance than a sqlite db.

Ideally all environment variables could easily be substituted with ${VAR:-default} but [that is not supported in compose files](https://github.com/docker/compose/issues/1377)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,22 @@
+gogs:
+  image: gogs/gogs:latest
+  links: 
+    - db
+  volumes:
+    - /var/gogs:/data
+  ports:
+    - "10080:3000"
+    - "10022:22"
+  environment:
+    DB_TYPE: mysql
+    HOST: db
+    NAME: gogs
+    USER: gogs
+    PASSWD: changeme
+db:
+  image: library/mariadb:latest
+  environment:
+    MYSQL_ROOT_PASSWORD: changeme
+    MYSQL_DATABASE: gogs
+    MYSQL_USER: gogs
+    MYSQL_PASSWORD: changeme


rothgar:
If this will be merged I will also update the README to let users know about the variables and compose option.

Unknwon:
Thanks your PR!

> Even though I specify environment variables for the database the initial setup screen is still shown and the variables I provide are not filled in. It's easy to add the values but probably should be done automatically based on the environment

I think Gogs does not support reading config values from environment variables.

> I link the containers but it does not appear the application looks in /etc/hosts file where the link is stored. I verified the right values were in /etc/hosts but I would still get an error trying to link the db by name (tried db docker_db docker_db_1 etc) It works if I use 127.0.0.1 or the ip address of the db container but not hostname.

There is an issue about this, but no one knows why yet...

Unknwon:
Also, please see https://github.com/gogits/gogs/wiki/Contributing-Code#final-checks

Gibheer:
@rothgar As for the /etc/hosts problem - that isn't an application problem, but might be a system configuration problem. Take a look into /etc/nsswitch.conf. What does it say there for hosts?

rothgar:
@Gibheer I'll have a look and see if I can resolve the issue.

@Unknwon I'll resubmit to the develop branch but I'm not sure the variable to set true in the app.ini. Line 95 was empty in the [admin] section

Unknwon:
@rothgar line 95 was `INSTALL_LOCK = false`.

Unknwon:
ping @rothgar

rothgar:
I was holding off on this until I could figure out why DNS resolution doesn't work in the application. I was trying to solve that problem but it's not necessary for a compose file to work (the db is available on 127.0.0.1). I'll close this for now so I can re-open on the right branch with an adjusted config.

PR #2277, opened 2015-12-24
Add Docker Volume from 1.9

jgsqware:


Unknwon:
Thanks your PR!

What makes this section to be different from `Volume with data container` section?

juliengarcia:
The volume is used by the database sqlite3 for example.

It's differing by the use of the new command line Docker volume create from
Docker 1.9 to create a persistent data container

Unknwon:
Thanks for following up!

I think this section should be a sub-section of `Volume with data container`.

And only following content is needed:

```
# Create docker volume.
$ docker volume create --name gogs-data

# Use `docker run` for the first time.
$ docker run --name=gogs -p 10022:22 -p 10080:3000 -v gogs-data:/data gogs/gogs
```

Unknwon:
ping @jgsqware

jgsqware:
I will update my commit, but I have a quesiotn on that, the amend and push is ok or I need to make a new pull request?

ByStones:
@jgsqware
I think you have to open a new pull request against the "develop" branch (or change the target of this one if possible) based on this: https://github.com/gogits/gogs/wiki/Contributing-Code

Unknwon:
Yes, @ByStones is right.

jgsqware:
Ok I'll close that one and create a new one on Develop branch. Sorry for that -> #2446

PR #2294, opened 2015-12-27
Fix the description on head.tmpl

joaopms:
Added a missing space, a missing "is" on the meta description tag and lowercased "Service".


Unknwon:
Thanks!

PR #2288, opened 2015-12-26
[Feature] Unified Diff-view

bkcsoft:
Implemented Unified Diff-view. Inline/Unified is selectable with a button (Fixes #1925)

2015-12-26-140919_1600x900_scrot
2015-12-26-114558_1600x900_scrot

PR #2296, opened 2015-12-28
Implement Split Diff-View

bkcsoft:
Opened up a new PR since the previous one ( #2288) was for the wrong branch.

Changed the name from Unified => Split and rebased on develop

~~This is still a **WIP** so don't merge it yet, it's just here to show that I'm still working on it...~~

Stuff that still needs fixing before considering merging:
- [x] move inline styles to .less-files
- [x] testing (see https://github.com/gogits/gogs/pull/2288#issuecomment-167347361 and https://github.com/gogits/gogs/pull/2288#issuecomment-167424391)
- [x] move a few hard-coded strings from .tmpl-files to i18n


bkcsoft:
@andreynering could you pull the latest and see if the problem is still there?
technically it was still side-by-side, just that the line was to wide and made it scroll (almost) forever ^.^ anyhow, I've made it line-wrap instead since it makes more sense for split-views :)

2016-01-01-233550_1600x900_scrot

2016-01-01-233558_1600x900_scrot

bkcsoft:
Now working as intended

@Unknwon thoughts?

Unknwon:
I think the button should be called `Unified` instead of `inline`.

andreynering:
@bkcsoft Did you commit this change? Because for me it is as before.

I think you should rebase (or squash?) your commits, since some commits from @Unknwon is showing in this pull requests.

But I'd say this is working as expected.

bkcsoft:
@Unknwon yeah, I think that makes most sense on second thought

@andreynering I've commited the changes, which web-browser are you testing in? (I'm currently only testing on firefox)
about the leaking commits, I rebased on develop and that seems to have borken something -. - I'll see if I can make it cleaner

andreynering:
> Which web-browser are you testing in?

Chrome

bkcsoft:
@andreynering yeah, not getting that on firefox or chrome. sure you checked it out cleanly and built it correctly? (`make build` and all that...)

@Unknwon Renamed and cleaned up a bit :)

andreynering:
@bkcsoft I tried in Firefox and it's OK. In Chrome it's still as before.

bkcsoft:
@andreynering which version if chrome? Since I canct reproduce it :/

andreynering:
47.0.2526.106

bkcsoft:
I'll have a look later today :)

bkcsoft:
@andreynering currently testing in "Google Chrome 47.0.2526.80" and it "Works On My Machine" ^.^
Can't really see how .80 and .106 breaks it, I'll try 48.x or 49.x and see if it's still working there :)

Unknwon:
Thanks @bkcsoft !

Just get back from trip. Will start reviewing today. (I'm on +8 CST right now)

bkcsoft:
@andreynering odd, I tried it in "Google Chrome 47.0.2526.80" and there it worked as expected, but in "Google Chrome 47.0.2526.106" I see what you're seeing...

Anyhow, the issue is with word-wrapping... I wrap on whitespace, but there's no whitespace to wrap on :/

andreynering:
I am playing a little here, trying to highlight what changed in each line, like GitHub and GitLab do:







There are some few bugs, but I hope I can get this working 100%.

bkcsoft:
@andreynering well, I can make it either break in the middle of words, OR break on whitespace... apparently not both -.-

about highlighting changes, that might be _really_ hard... for simple stuff it easy enough:
```
- foo()
+ bar()
```
but for a bit more complex stuff it is harder:
```
- foo()
- bar(1)
+ foo()
+ baz()
+ bar(2)
```
^ this one should highlight only the numbers (and perhaps `baz()`) :smile:

bkcsoft:
@andreynering well, I can make it either break in the middle of words, OR break on whitespace... apparently not both -.-

unless one uses an undocumented css-value... namely `word-break: break-word`...
I'll commit and push the fix so you can try it :smile:

andreynering:
I will try to get it right and maybe send a PR, but will wait this one be
merged first.

I am using a third party package that does the diff of the strings.

Em dom, 3 de jan de 2016 22:14, Kim Carlbäcker <notifications@github.com>
escreveu:

bkcsoft:
@andreynering yeah, I looked at diff-packages when I started this but it seemed like they wanted rad .patch-files (which I didn't wanna push to the frontend...) so I gave up on that...

bkcsoft:
Verified to work in the following browsers:
Chrome stable 47.0.2526.106
Firefox 43.0.1
Microsoft Edge 20.10240.16384.0 (that version-number xD
Internet Explorer 11.0.10240.16431
Opera 33.0.1990,115

andreynering:
@bkcsoft It's OK now. :+1:

Unknwon:
Oops! There are conflicts on bindata.go

Unknwon:
PS: A rebase is recommended to keep commit history clean.

routers/repo/commit.go
```diff
@@ -168,6 +168,7 @@ func Diff(ctx *middleware.Context) {
 		}
 	}
 
+	ctx.Data["Style"] = ctx.Query("style") == "split"
```
Unknwon: Shouldn't this be `IsSplitStyle` since it's a `bool` value?

routers/repo/commit.go
```diff
@@ -213,6 +214,7 @@ func CompareDiff(ctx *middleware.Context) {
 	}
 	commits = models.ValidateCommitsWithEmails(commits)
 
+	ctx.Data["Style"] = ctx.Query("style") == "split"
```
Unknwon: Same as above.

templates/repo/diff_box_unified.tmpl
```diff
@@ -0,0 +1,74 @@
...
+											<td class="lines-code">
+												<pre>{{$line.Content}}</pre>
+											</td>
+										</tr>
+									{{end}}
+								{{end}}
+							</tbody>
+						</table>
+					</div>
+					<script>
```
Unknwon: Should this part be removed?

templates/repo/diff_box_split.tmpl
```diff
@@ -0,0 +1,98 @@
...
+							var prev = $(this).prev();
+							if(prev.is('.del-code') && prev.children().eq(3).text().trim() === '') {
+								while(prev.prev().is('.del-code') && prev.prev().children().eq(3).text().trim() === '') {
+									prev = prev.prev();
+								}
+								prev.children().eq(3).html($(this).children().eq(3).html());
+								prev.children().eq(2).html($(this).children().eq(2).html());
+								prev.children().eq(3).addClass('add-code');
+								prev.children().eq(2).addClass('add-code');
+//								prev.children().eq(2).attr('style', 'background-color: #eaffea !important');
```
Unknwon: What's going on with commented out lines?

Unknwon:
Thanks agin for the PR!

I checked out locally and works great AFA I can tell.

Some comments are made, also I see a lot duplicated content between `diff_box_split.tmpl` and `diff_box_unified.tmpl`, is it worth doing it or should we reduce the duplicated part?

routers/repo/commit.go
bkcsoft: Indeed it should be IsSplitStyle

templates/repo/diff_box_split.tmpl
bkcsoft: Old code, I'll fix that

templates/repo/diff_box_unified.tmpl
bkcsoft: Yes. Most've missed that when I moved this part to it's own file

bkcsoft:
@Unknwon I've missed to remove to old code, I'll remove that and squash some commits to make it cleaner. Also resolve the bindata.go-issue

About the duplicate code, I made that decision because it'll only do 1 bool-check (and not 1 per line). My reasoning for that is that it should be faster to render, I can do some benchmarks if you'd like to see how much slower it'd be to do it inline. (just need to find a big commit first :smile:)

Unknwon:
> About the duplicate code, I made that decision because it'll only do 1 bool-check (and not 1 per line). My reasoning for that is that it should be faster to render, I can do some benchmarks if you'd like to see how much slower it'd be to do it inline. (just need to find a big commit first :smile:)

I don't mean to check bool per line, but reduce the check to a smaller range of content. As you can see, things are getting different inside `tbody` tag: https://github.com/gogits/gogs/pull/2296/files#diff-9d1f05e66c594694d453ec18a343782dR35

bkcsoft:
doing it on the `tbody`-tag would make it check `IsSplitStyle` on each listed file. but if that's acceptable I'll do that :)

bkcsoft:
merging them and having it check per-file (on `tbody`) didn't even make a dent in the render-time even for some of the larger commits I could find :smile:

bkcsoft:
Oh, I completely forgot to mention that this PR also includes the fold-icon (`opticon-fold`)
Which is what this is doing: https://github.com/gogits/gogs/pull/2296/files#diff-f0b73ac65b702903d90a117166a9b06bR75 (down to row 88)

Like this: 
fold-icon

Not sure if I should leave it there or have it in a separate PR?

Unknwon:
Never mind, I'll do that after merge.

And you need at least comment out fold icon, it's not working yet.

bkcsoft:
Already pushed the template-merging :)

The icon is more of a visual thing for me. but I'll remove it and see if I can make it work in a separate PR :)

Unknwon:
Oh, I see. I'll checkout again later today. :grinning:

Unknwon:
LGTM.

mastorm:
Is there a flag to enable split-view by default?

Unknwon:
@mastorm currently, no..

PR #2322, opened 2016-01-02
Updated systemd service file

noisykeyboard:


Unknwon:
Thanks your PR!

But why this change is needed?

noisykeyboard:
I'm running the newest version of gogs from your packager.io repository and it tells me that 'gogs web' is no longer an option and that i should launch it with 'gogs run web' instead, when i changed modified the systemd file like i did in the pull request i was able to run gogs again.

Unknwon:
Thanks!

But I think it's more packager.io thing, and that `gogs` is a command, which is not real `gogs` binary it compiles.

Gogs does not have `run` option at all.

PR #2323, opened 2016-01-02
Update s6 path in Dockerfile.rpi

ggramlich:
The Dockerfile.rpi (Raspberry / ARM Version) pointed to the old path for the s6-vscan command. As a consequence, docker run without providing a custom command fails.

This pull request corrects the path by applying the change in the Dockerfile from https://github.com/gogits/gogs/commit/0cbf56855aad82ff49952132287a0112f308c728 to the Dockerfile.rpi as well

I successfully built a container from the corrected version of Dockerfile.rpi (based on V0.8.10)


Unknwon:
Thank you very much!

PR #2325, opened 2016-01-02
Strike through issue ids in resolving commits

### LINKED ISSUE #2266: Strike through closed issue ids in commit list
On Bitbucket the ids of closed issues are displayed crossed out.
I would like to have the ids on Gogs displayed the same way, to make it more clear, if a commit closes an issue or just references it.

lukasdietrich:
This is a proposal to resolve #2266


Unknwon:
Thanks your PR!

But I can't figure out why it solves #2266?

lukasdietrich:
As I tried to explain in https://github.com/gogits/gogs/issues/2266#issuecomment-168028923, this just strikes out issue ids based on a closing keyword, rather than quering for the status of an issue in the database. This way it should be as performant as before.

Unknwon:
Commit message has `a closing keyword` does not mean it is closed...

lukasdietrich:
It does not mean it is closed, but it means, that this commit *should* close the issue, the same way the links do not check if an issue even exists.

Unknwon:
> It does not mean it is closed, but it means, that this commit should close the issue

Yes, but then what's the point of showing strike?

lukasdietrich:
To visually show, which commits *are* closing issues and which just reference. Its just to visually emphazise. Plus I am not sure, if the database even holds information about which commits closed an issue, just a list of referencing commits followed by a close action.

Unknwon:
> To visually show, which commits are closing issues and which just reference.

In this case, I think it is not worth doing. Strike through issues which are **really closed** is more useful.

PR #2328, opened 2016-01-03
Apply suggestions by github.com/mvdan/interfacer

mvdan:


Unknwon:
Thanks your PR!

However,

1. Please read https://github.com/gogits/gogs/wiki/Contributing-Code#final-checks
2. Places use `session` type are on purpose.

mvdan:
> Please read https://github.com/gogits/gogs/wiki/Contributing-Code#final-checks

Unintuitive, but okay.

> Places use session type are on purpose, so the changes of PR do not always apply.

Which places?

Unknwon:
Didn't look deep into every case, but most part of the change is required to use `session` because it may have possible transactions and need to be able to roll back.

PR #2336, opened 2016-01-04
Making scroll always visible.

andreynering:
Simple trick to prevent the page from "jumping" when navigating from/to a page not long enought to have the scroll.


Unknwon:
Thanks your PR!

Can you take a screenshot for the visual effect?

andreynering:
This is not something one will note unless they pay attention, actually. The elements of the page will move very slightly to right when navigating to a page without the scrollbar.

Not a big deal, but since the fix is easy, I made it.

Unknwon:
OK, make sense. Thanks again!

PR #2347, opened 2016-01-06
Remove RSA1 keys as only SSH version 2 is used

ivanmarban:
Since OpenSSH is configured for using protocol version 2, it's worthless the generation and use of RSA1 keys.

Building a Docker container using rpi version, OpenSSH 7.1 is used and due to a bug on this version RSA1 keys generation is broken and you'll get errors:

`Saving key "/data/ssh/ssh_host_key" failed: unknown or unsupported key type`
`Could not load host key: /data/ssh/ssh_host_key` 

Anyway, cloning via SSH seems to work fine.

Building a Docker container with normal version, OpenSSH 6.8 is used and there are no errors.

### DIFF
--- a/docker/s6/openssh/setup
+++ b/docker/s6/openssh/setup
@@ -1,10 +1,6 @@
 #!/bin/sh
 
 # Check if host keys are present, else create them
-if ! test -f /data/ssh/ssh_host_key; then
-    ssh-keygen -q -f /data/ssh/ssh_host_key -N '' -t rsa1
-fi
-
 if ! test -f /data/ssh/ssh_host_rsa_key; then
     ssh-keygen -q -f /data/ssh/ssh_host_rsa_key -N '' -t rsa
 fi
--- a/docker/sshd_config
+++ b/docker/sshd_config
@@ -4,7 +4,6 @@ ListenAddress 0.0.0.0
 ListenAddress ::
 Protocol 2
 LogLevel INFO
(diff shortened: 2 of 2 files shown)


Unknwon:
Thanks your PR!

cc @0rax

0rax:
This should not be a problem as RSA1 keys are pretty much no longer used. It was configured to be created as it was still compatible with the OpenSSH version present on Alpine Linux.

This type of key is only used with the old version of the SSH protocol (SSH1) and we are discouraged to use this version. (also this protocol was not enabled in our `sshd_config` so the key should not have been used at all, my bad for creating it in the first place).

Versions of OpenSSH `<7` were still implementing SSH1 using this `ssh_host_key` but going onward OpenSSH will not longer support SSH1 as part of its comunication protocols. And as we know, the next version coming to Alpine Linux of OpenSSH is version 7.1 so we will not be able to support SSH1.

So as we are discouraged to use this protocol and considering that OpenSSH version 7 will not support SSH1 and that this protocol is already deactivated in our `sshd_config`, it is totally okay to remove this key.

Unknwon:
Thanks @0rax ! Are you suggesting this PR can be merged?

0rax:
Yes, you can go on and merge it, I've tested it by removing the key &
changing the configuration manually and everything worked fine :)

Unknwon:
Thank you both!

PR #2356, opened 2016-01-07
make macaron setup public

koneu:
This allows embedding Gogs in a custom HTTP server, eliminating the need for proxying access to it.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -175,10 +175,7 @@ func newMacaron() *macaron.Macaron {
 	return m
 }
 
-func runWeb(ctx *cli.Context) {
-	if ctx.IsSet("config") {
-		setting.CustomConf = ctx.String("config")
-	}
+func Macaron() *macaron.Macaron {
 	routers.GlobalInit()
 	checkVersion()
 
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks, but I do not think `Macaron` is a good name.

Unknwon:
And, why is it a good idea to make it public?

koneu:
It's handy for serving Gogs while also serving other content without configuring a proxy. One simply takes the returned `*macaron.Macaron` and tosses it into the `http.Handler` multiplexer. This essentially makes Gogs usable as a plugin/library, rather than as a standalone executable.
For naming it I simply thought "it returns a `Macaron`, let's call it `Macaron`.

Unknwon:
Thanks your explanation!

But all code in Gogs are supposed to be internal use only, things could break because you suddenly to able to change public objects on sub-package which should only be used by Gogs itself.

koneu:
I see your point, and I won't mind maintaining this tiny patch out of tree.

I don't think exposing a single function to the public will make things overly unstable, though. While `cmd` might not quite be the proper place for exposing public API calls, it's where the concerned code currently lives. If you're worried about people taking the `*macaron.Macaron` and doing bad things with it, changing the function's return type to `http.Handler` might make you feel better.

Unknwon:
It's not just about a small thing, it's about people should never ever use internal code of an independent app, because Gogs is never considered to be used as a library. 

And of course, you can maintain this little patch in your own copy, which is one of the good things about open source.

PR #2369, opened 2016-01-09
Fix status command in FreeBSD init script

koenwtje:
If the init script is called with `status`, the rc.subr(8) routines check if
the first argument associated with the pid in the pidfile is equal to
`procname`. By default, `procname` is equal to the value of `command`. In our
case, `command` contains a space (i.e. has multiple arguments), so `procname`
can never be equal to the first argument of the command associated with the
pid.

Set `procname` to the first argument of `command` to fix the `status` command
of the init script.


Unknwon:
Thank you!

Issue #2349, opened 2016-01-06
PANIC: interface conversion: interface is []uint8, not int64


Ingramz:
After adding PAM as an authentication method, I am immediately greeted with a blank page and following panic.

Installed from source with pam tag using [this guide](https://gogs.io/docs/installation/install_from_source). Using Debian 8 if that matters.

```
[Macaron] PANIC: interface conversion: interface is []uint8, not int64
/usr/local/go/src/runtime/panic.go:423 (0x42ba09)
/usr/local/go/src/runtime/iface.go:221 (0x40d5f0)
/home/git/go/src/github.com/gogits/gogs/models/login.go:111 (0x548dd2)
        (*LoginSource).BeforeSet: switch LoginType((*val).(int64)) {
/home/git/go/src/github.com/go-xorm/xorm/session.go:1656 (0x6e8da5)
        (*Session)._row2Bean: b.BeforeSet(key, Cell(scanResults[ii].(*interface{})))
/home/git/go/src/github.com/go-xorm/xorm/session.go:1625 (0x6e8783)
        (*Session).rows2Beans: err := session._row2Bean(rows, fields, fieldsCount, bean, &dataStruct, table)
/home/git/go/src/github.com/go-xorm/xorm/session.go:1365 (0x6e4617)
        (*Session).Find: return session.rows2Beans(rawRows, fields, fieldsCount, table, newElemFunc, sliceValueSetFunc)
/home/git/go/src/github.com/go-xorm/xorm/engine.go:1347 (0x6b1389)
        (*Engine).Find: return session.Find(beans, condiBeans...)
/home/git/go/src/github.com/gogits/gogs/models/login.go:191 (0x549768)
        LoginSources: return auths, x.Find(&auths)
/home/git/go/src/github.com/gogits/gogs/routers/admin/auths.go:32 (0x7f69fc)
        Authentications: ctx.Data["Sources"], err = models.LoginSources()
/usr/local/go/src/runtime/asm_amd64.s:437 (0x45cbce)
/usr/local/go/src/reflect/value.go:432 (0x87fe4a)
/usr/local/go/src/reflect/value.go:300 (0x87eb11)
/home/git/go/src/github.com/go-macaron/inject/inject.go:117 (0xb28a42)
        (*injector).Invoke: return reflect.ValueOf(f).Call(in), nil
/home/git/go/src/gopkg.in/macaron.v1/context.go:113 (0x74b8d7)
[...]

Issue #1894, opened 2015-11-04
LDAP authentication with TLS


AceSlash:
Hello,

When I try to use TLS to connect to OpenLDAP, I have the following error:

```
Nov  4 18:25:18 gogs-test gogs-web-1.service[8722]: [Macaron] Started POST /user/login for 10.99.1.1
Nov  4 18:25:18 gogs-test gogs-web-1.service[8722]: 2015/11/04 18:25:18 #033[1;36m[T] LDAP will use BindDN.#033[0m
Nov  4 18:25:18 gogs-test gogs-web-1.service[8722]: 2015/11/04 18:25:18 #033[1;34m[D] Using TLS for LDAP without verifying: true#033[0m
Nov  4 18:25:18 gogs-test gogs-web-1.service[8722]: 2015/11/04 18:25:18 #033[1;31m[...es/auth/ldap/ldap.go:125 SearchEntry()] [E] LDAP Connect error, ldap1.example.com:LDAP Result Code 200 "": EOF#033[0m
Nov  4 18:25:18 gogs-test gogs-web-1.service[8722]: [Macaron] Completed /user/login 200 OK in 9.410884ms
```
I'm using the following versions:
gogs version 0.6.15-1446616151.1b5e1be.jessie
git version 2.1.4-2.1
postgresql 9.4.5
Debian Jessie (8.2)

I tried both STARTTLS and SSL port (389 and 636, which are both open). The certificate used by the LDAP server is signed by an internal private authority (not commercial), TLS is working with ldapsearch from the server with the credential configured on GOGS without any issue.

Both the boxes "Enable TLS Encryption" and "Skip TLS Verify" are checked.

The settings I changed from the default configuration are:

PROTOCOL = https
DOMAIN = scm.example.com
DISABLE_SSH = true
CERT_FILE = (path to a valid commercial certificate)
KEY_FILE = (path to a valid commercial certificate key)

And the settings to the database and debug level to trace.

If I deactivate TLS on the OpenLDAP server and on GOGS, everything works as expected.

Logs from OpenLDAP during the exchange available here: 

```
Nov  4 18:25:19 ldap1-test slapd[11954]: daemon: activity on 1 descriptor
Nov  4 18:25:19 ldap1-test slapd[11954]: daemon: activity on:
Nov  4 18:25:19 ldap1-test slapd[11954]: 
Nov  4 18:25:19 ldap1-test slapd[11954]: slap_listener_activate(8): 
[...]

PR #2376, opened 2016-01-10
Conditional StartTLS elevation based on UseSSL

### LINKED ISSUE #1666: LDAP binding fails with TLS enabled
LDAP binding without TLS is successful, but turning on *Enable TLS Encryption* fails with an error. I run into the same error if I also enable *Skip TLS Verify*

2015/09/21 03:49:19 [T] LDAP will use BindDN.
2015/09/21 03:49:19 [D] Using TLS for LDAP without verifying: false
2015/09/21 03:49:19 [...es/auth/ldap/ldap.go:93 SearchEntry()] [E] LDAP Connect error, example.net:LDAP Result Code 200 "": EOF

I have verified the LDAP server does support TLS by passing the *-Z* option used to "Start TLS request (-ZZ to require successful response)"

$ ldapwhoami -vvv -H ldap://ldap.example.net -x -ZZ
ldap_initialize( ldap://ldap.example.net:389/??base )
anonymous
Result: Success (0)

Gogs Version: 0.6.13.0918 Beta

vilppuvuorinen:
StartTLS workaround similar to Openshift. Would fix #1666 and #1894.

I have been able to test this locally using `LDAP (simple auth)` authentication type. #2349 made the testing an absolute pain though.

Other changes:

* Migration to prevent breaking existing installations. UseSSL==true `ldap` authentications will jump to using `StartTLS` if no actions are taken to prevent it.
* Documentation updates
* UI field name updates

### DIFF
--- a/modules/auth/ldap/ldap.go
+++ b/modules/auth/ldap/ldap.go
@@ -202,6 +202,22 @@ func ldapDial(ls *Source) (*ldap.Conn, error) {
 			InsecureSkipVerify: ls.SkipVerify,
 		})
 	} else {
-		return ldap.Dial("tcp", fmt.Sprintf("%s:%d", ls.Host, ls.Port))
+		conn, err := ldap.Dial("tcp", fmt.Sprintf("%s:%d", ls.Host, ls.Port))
+		if err != nil {
+			return nil, err
+		}
+
+		if ls.SkipVerify {
+			return conn, nil
+		}
+
+		if err := conn.StartTLS(&tls.Config{
+			InsecureSkipVerify: false,
+		}); err != nil {
+			conn.Close()
+			return nil, err
+		}
+
+		return conn, nil
 	}
 }


SergioBenitez:
This is a bit weird. If I'm reading this correctly, the following UI selections (on the left) translate to the following LDAP connection (after the =):

* SSL, SkipVerify = LDAPS
* SSL no SkipVerify = LDAPS
* No SSL, SkipVerify = Unencrypted
* No SSL, no SkipVerify = StartTLS

This doesn't make a lot of intuitive sense to me. The "SkipVerify" UI field has no effect if we select SSL, and when we _don't_ select SSL, _not_ selecting SkipVerify results in the connection actually using SSL.

What should just let the user select one of three options:

1. Use LDAP over TLS (LDAPS)
2. Use StartTLS
3. Use unencrypted LDAP

vilppuvuorinen:
You are absolutely right about. Having those choices would require changes to every component from UI to DB making it quite invasive.

The "SkipVerify" field should work as expected when using `ldaps`. The original choices were

* SSL, SkipVerify = LDAPS
* SSL no SkipVerify = LDAPS
* No SSL, SkipVerify = Unencrypted
* No SSL, no SkipVerify = Unencrypted (unused/illegal state)

It looks more like utilizing an unused field. Changing the checkbox labels could also clear things up. This is pretty much the only way to get `StartTLS` working without doing larger changes to the DB.

The other option would be to use port number to select the encryption scheme that the user probably wants to use and try the other one if neccessary. E.g. port 636 -> `ldaps` first, port 389 -> `StartTLS` first.

SergioBenitez:
Yeah, I didn't like the previous UI either, but at least there was one option, whether to use SSL/TLS, since the verify option did nothing. Using the port is a bad idea. There's nothing special about the port numbers other than being the defaults. Treating them specially is sure to be error prone.

Making the changes I suggest shouldn't be too big of an issue since everything goes through an ORM-ish layer.

vilppuvuorinen:
The port numbers in the second suggestion were there only to select the first method to try. Most of the installations are using default ports anyways and non-standard ports would only result in degraded performance due to multiple tries.

Having the extra fields in data model is the cleanest solution. However, it is not the simplest one either. @Unknwon: What do you think?

Unknwon:
From what I understand, does @SergioBenitez suggests that replace checkbox of `skip verify` to a dropdown list with three options?

- Use LDAP over TLS (LDAPS)
- Use StartTLS
- Use unencrypted LDAP

SergioBenitez:
@Unknwon: No, not necessarily. I think the "Skip Verify" option doesn't make a lot of sense here.

I think what I would _really_ like is the following:

1. Consolidate the two LDAP (simple auth, bind DN) options into *one* "LDAP" option.
2. Add a "Mechanism" (or whatever...) drop down selection with two options:
  1. Simple
  2. Bind DN
3. Add a "LDAP Protocol" drop down selection with three options:
  1. Unencrypted
  2. Secure (LDAPS)
  3. StartTLS
4. Remove the "Enable TLS encryption" and "Skip TLS verify" checkboxes.

The form should update automatically to show only the relevant fields when either "Simple" or "Bind DN" are selected.

This requires a bit more work, of course, but I think it's likely worth it. For now, we could simply do 3 and 4 above, which was my original suggestion.

Unknwon:
Thanks @SergioBenitez !

Agree to 3, but I don't think `Skip TLS verify` option could be removed?

Unknwon:
A quick summary of what I'm planning to do:

Add an "LDAP Protocol" drop down selection with three options as @SergioBenitez suggested:

1. Unencrypted (hide `Skip TLS verify` checkbox, no TLS)
2. LDAPS (show `Skip TLS verify` checkbox, TLS verify on choice)
3. StartTLS (show `Skip TLS verify` checkbox, TLS verify on choice)

Unknwon:
This approach requires auto-migration:

1. When `UseSSL` is `true`, treats it as `LDAPS`.
2. When `UseSSL` is `false`, treats it as `Unencrypted`.

Do not need to care `StartTLS` since Gogs does not support it before the point.

Thoughts?

SergioBenitez:
Why can't we simply not show the `Skip TLS verify` options for LDAP?

Unknwon:
`Unencrypted` will hide `Skip TLS verify` option as I listed.

SergioBenitez:
Right, I understood that. But, we don't need the `Skip TLS verify` option at all for LDAP if we have the three options (Unencrypted, LDAPS, StartTLS).

Unknwon:
- For LDAPS, we need it to support self-signed certificates.
- For StartTLS, we need it do determine whether to call `StartTLS` or not.

SergioBenitez:
Ah, I see. So you're saying the `Skip TLS verify` option _is_ currently being used to skip checking against known CAs? Any chance you could point me to the section of code that does this? If this is the case, then the option should be there for `StartTLS` as well. When a user selects `StartTLS`, we should always attempt to use `StartTLS` and error out if it doesn't work.

Unknwon:
Here you go: https://github.com/gogits/gogs/blob/master/modules/auth/ldap/ldap.go#L202

By default, Go does not allow self-signed CAs.

SergioBenitez:
Ah, excellent.

Okay then! Your comment:

1. Unencrypted (hide Skip TLS verify checkbox, no TLS)
2. LDAPS (show Skip TLS verify checkbox, TLS verify on choice)
3. StartTLS (show Skip TLS verify checkbox, TLS verify on choice)

looks perfect then!

Unknwon:
:+1:

vilppuvuorinen:
It looks like a decision has been reached. Some other points:

ldap.Source struct: LDAPS/StartTLS/Unencrypted selection looks enumeration based to me. `ConnectionType string` (`"ldaps"|"starttls"|"ldap"`) or int based ids? Judging by the general database model separate table for types is probably not going to happen.

Other things is the migration strategy.
* UseSSL == true -> ldaps, SkipVerify remains
* UseSSL == false -> ldap, optionally set SkipVerify to false to remain consistent.

Unknwon:
@vilppuvuorinen just `int` type should be good enough.

Unknwon:
PS: I'll merge your PR first then start working on other matters.

Issue #1270, opened 2015-05-12
An error is occurred : too many open files


chanyuenpang:
it shows 500 or 400, description is 
An error is occurred : open xxx.git too many open files

linquize:
What operation did you do? For example, the URL?

Unknwon:
Thanks your feedback!

But it is a common problem on system setting, and Gogs really can't help with it. :flushed:

For example: http://askubuntu.com/questions/181215/too-many-open-files-how-to-find-the-culprit

simonszu:
Well, maybe it should? I have set the ulimit to 300k files - gogs was greedy and snatched them all. So i increased to 1 million open files, gogs was greedy and snatched them all. All while just browsing one repository with 500 commits and around 1000 files in sum. Regarding the fact that the default value is around 65k files, i would assume that there is something weird happening with gogs's file management, maybe it doesn't close them or something. 
I could open these repostiorys with cgit and gitlab very well, with the default value of 65k max open files...

If you really say that there's nothing that gogs can do, what are good values to set the file limt to? 10 million? 100 million? A billion? 

And - according to my CPU usage and the page load time - is it possible that gogs is crawling the repository recursively and trying to read each file at its own? Then it _is_ some kind of gogs's own problem.

Unknwon:
@simonszu Currently Gogs only reads all files/directories of current browsing level.

If you want to dive into this problem more, you should provide more information about your Gogs version, system, etc. And where the error occurs.

simonszu:
Okay, sure. 
So, details to the affected repository: 8652 files, 492 commits. It is an /etc-Directory of one of my hosts, managed by etckeeper. 
The host which i run gogs on is a VPS with Debian Jessie. 6GB RAM, Dualcore with 2.2GHz each. I am using gogs from binary, version 0.8.10. 

So, i will startup gogs manually (usually via systemd) and check how many open files will appear there after each pageload. The hard- and soft limit is set to 1 million open files, system wide. 

Beginning: gogs is stopped, there are around 9000 files open by the dropbox daemon i have running on the same host, and around 600 files by other processes. I have checked it via the lsof-awk-combination found on the page you linked above. 

Starting gogs, but not accessed it via browser yet: gogs has 24 open files. 
Acessing it, logging in, visiting my dashboard: 24 open files.
Visiting the specified repository: Number of open files increases to 304, gogs is, after dropbox, the process with the second most open file handlers on my system. 
[...]

PR #2388, opened 2016-01-12
Fixed the 'too many open files' error when starting gogs with systemd

simonszu:
This commit addresses https://github.com/gogits/gogs/issues/1270. It seems that systemd has its own settings of max open file limits - i'm setting this limit to 65535 open files, which should succeed for most repositories.

### DIFF
--- a/scripts/systemd/gogs.service
+++ b/scripts/systemd/gogs.service
@@ -8,8 +8,12 @@ After=network.target
 #After=redis.service
 
 [Service]
-LimitMEMLOCK=infinity
-LimitNOFILE=65535
+# Modify these two values and uncomment them if you have
+# repos with lots of files and get an HTTP error 500 because
+# of that
+###
+#LimitMEMLOCK=infinity
+#LimitNOFILE=65535
 Type=simple
 User=git
 Group=git
--- a/scripts/systemd/gogs.service
+++ b/scripts/systemd/gogs.service
@@ -8,6 +8,8 @@ After=network.target
 #After=redis.service
 
 [Service]
+LimitMEMLOCK=infinity
+LimitNOFILE=65535
 Type=simple
 User=git
 Group=git


winsx:
@simonszu nice job,the problem solved,thans very much

Dasoren:
-1 please do not merge this change. If these limits need to be changed, the system admin should make them, not gogs. Opening so many files like that can have major issues if you do not know what the system limits are.

sapk:
@Dasoren the change is only a recommendation since in the script folder not in gogs logic. Futhermore is only make change for the gogs runtime not system-wide. It's like saying "gogs need to be able to open at least XXX files")

simonszu:
I have modified my pull request to better represent that these values are optional and when you should use them. It is like @sapk says a recommendation, not a fixed value. 
Since the modified service file has also recommendations for service dependencies, according to what database you use, i think, this is okay.

sapk:
@simonszu The only problem is your PR is on gogs master branch. It should be on develop (see https://github.com/gogits/gogs/wiki/Contributing-Code) but since it's a no effect change maybe it will be accepted by @Unknwon

simonszu:
Ah, damn. You're right. 
If that's a problem, i can close this PR and create another one, following the developer guidelines, but let's first see what @Unknwon says.

Dasoren:
@simonszu What is your reasoning for making it 65535? My issue with this is, most users that will just install this and use the start script will have no idea what that setting it or what a reasonable number is ether. Setting the number to 10,000 would be much reasonable, unless you know something I do not on this. 

I have done a lot of tests with how many files git needs to open for a diff or old commit and have not found any repo that goes over 15,000 (note that I have not tested everything on github). I have one repo on my own that has about 40-60K files that are changed from commit to commit, but that is because I am using git as a backup (I know this is wrong and I have my reasons).

Could you explain the 65535 limit change and what repo would have that many changes from commit to commit? I am just one that would rather leave system changes like that up to the system admin and not a standard setting unless it is really needed. ( I know this is not a change to "Gogs" but it is in that the start script is being sent out as part of Gogs).

Dasoren:
@simonszu As a side note, I do not like that in your update you have them commented out and a little note there for the user to read.

simonszu:
The reason is as simple as it can be: The default ulimit hard-limit on Debian Jessie is set to 65535 (check it with `ulimit -n`), and since i run gogs on Debian Jessie, i had no reason to let systemd limit its own open files to a value lower than the system's value. 

As you can see in the second commit i have attached to this PR, the setting of systemd's file limits via gogs's unit file is now just a recommendation, but i had to enter a numeric value, so that the user knows he should use numeric values.
I have commented them out, because the other recommendations, speak: The database dependencies, according to what underlying database service gogs should use are commented out as well. 

This issue that systemd has its own file limits is not widely known yet, and since other config files (not only gogs', but from other services as well) have comments inside them, explaining what each setting results in, i thought, it would be a good idea to make a comment about these settings in here as well, because systemd-unit-files are, in fact, config files.

Dasoren:
Much better thank you for explaining your thinking on this. As long as the values stay comment out by default I see no issue with adding them into the config file.

nanoant:
FYI I came across the same issue here on OS X server running Gogs, I have upgraded to latest version using concurrent commit info loading causing multiple pipes open at once. Default soft-limit for OS X is 256, so I had raised this to 8192 using launchd LaunchDaemons plist file entry:

		<dict>
			<key>NumberOfFiles</key>
			<integer>8192</integer>
		</dict>

bkcsoft:
I agree [mostly] with @Dasoren on this. This is something that shouldn't be set by default, and should be set by the sysadmin. Though I'm not against having it there as long as it's commented out and with sane values :smile: 

> Furthermore is only make change for the gogs runtime not system-wide

This isn't actually true, you're allowing gogs to potentally drain system-resources to the point where the system can't function normally anymore (bad idea)

> It's like saying "gogs need to be able to open at least XXX files"

This is plain wrong, it's actually saying "gogs should be *allowed* to have XXX files opened if the system is able to open them".

> This issue that systemd has its own file limits is not widely known yet

On this I agree, there isn't much talk about this since it's usually not an issue (except for file-intensive applications). And it's actually sort-of tricky to get the value out of systemd.

> [I] had no reason to let systemd limit its own open files to a value lower than the system's value

Well, systemd-devs most likely had their reasons :japanese_goblin:
Also, do keep in mind that file-intensive applications like bittorrent-clients and database-server (rememember that **a socket is a file** in a unix-system) usually limit to 8k or 16k depending on load... 65k is mosly regarded as a leaking application :) (unless you have github-level load (per-node obviosly, 1 node isn't gonna handle github by itself :laughing: )

Also, I can't really figure out why you're fiddling with MEMLOCK, since it doesn't relate to the issue...
From http://linux.die.net/man/2/getrlimit:
> The maximum number of bytes of memory that may be locked into RAM. In effect this limit is rounded down to the nearest multiple of the system page size. This limit affects mlock(2) and mlockall(2) and the mmap(2) MAP_LOCKED operation.

It only messes with how much memory it can _lock_ in RAM (i.e. memory it can't swap)... this is **not** related to the issue. And setting it to 'infinity' is a _really_ bad idea since gogs can potentially shove every other application running (that doesn't do the same) into swap :unamused: 

(WOO! longest comment I've written on github thusfar :smile: )

Dasoren:
thank you @bkcsoft Much better then I could have written up and to the point.

nanoant:
After thinking while I am coming to conclusion that current Gogs behavior is buggy, or at least it is wrong. The problem here is that for each connection or request Gogs may need theoretically to open 10-100 pipes to `git` as there is no sane limit set in Gogs itself. Dogs is just trying to run as many parallel git commands as possible, regardless of number of commands already running. So this PR is just a temporary workaround but #1270 still needs to be fixed.

Unknwon:
Thank you all :+1:

@simonszu yes this PR should be redone to `develop`.

@nanoant I've reopened the #1270

Unknwon:
Hijacked due to lack of feedback.

PR #2396, opened 2016-01-13
[Feature] Markdown Checklist-rendering

### LINKED ISSUE #1048: Support GitHub-style checklists/tasklists
https://github.com/blog/1375%0A-task-lists-in-gfm-issues-pulls-comments

I use a feature similar to this for larger issues when I don't want to file 20 smaller issues for various tasks. Would be nice to see something like it in Gogs for quickly specing out lots of work.

Thanks.

bkcsoft:
Implementing checklist-feature for the markdown-renderer (should work everywhere, not just for issues)

relates to #1048 :) (Though not as feature-complete as Github is, since github supports interactivly checking/unchecking without editing the issue)

### DIFF
--- a/modules/base/markdown.go
+++ b/modules/base/markdown.go
@@ -142,6 +142,16 @@ func (r *CustomRender) AutoLink(out *bytes.Buffer, link []byte, kind int) {
 	r.Renderer.AutoLink(out, link, kind)
 }
 
+func (options *CustomRender) ListItem(out *bytes.Buffer, text []byte, flags int) {
+	switch {
+	case bytes.HasPrefix(text, []byte("[ ] ")):
+		text = append([]byte(`<input type="checkbox" disabled="" />`), text[3:]...)
+	case bytes.HasPrefix(text, []byte("[x] ")):
+		text = append([]byte(`<input type="checkbox" disabled="" checked="" />`), text[3:]...)
+	}
+	options.Renderer.ListItem(out, text, flags)
+}
+
 var (
 	svgSuffix         = []byte(".svg")
 	svgSuffixWithMark = []byte(".svg?")
--- a/modules/base/tool.go
+++ b/modules/base/tool.go
@@ -31,7 +31,16 @@ import (
 	"github.com/gogits/gogs/modules/setting"
 )
 
-var Sanitizer = bluemonday.UGCPolicy().AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code").AllowElements("input").AllowAttrs("type", "checked", "disabled").OnElements("input")
+func BuildSanitizer() (p *bluemonday.Policy) {
+	p = bluemonday.UGCPolicy()
+	p.AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code")
+
+	p.AllowAttrs("type").Matching(regexp.MustCompile(`^checkbox$`)).OnElements("input")
+	p.AllowAttrs("checked", "disabled").OnElements("input")
+	return p
+}
+
+var Sanitizer = BuildSanitizer()
 
 // EncodeMD5 encodes string to md5 hex value.
 func EncodeMD5(str string) string {
--- a/modules/base/tool.go
+++ b/modules/base/tool.go
@@ -31,7 +31,7 @@ import (
 	"github.com/gogits/gogs/modules/setting"
 )
 
-var Sanitizer = bluemonday.UGCPolicy().AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code")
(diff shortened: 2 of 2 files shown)


bkcsoft:
actually something like this would be safer:
```
diff --git a/modules/base/markdown.go b/modules/base/markdown.go
index 0ef379b..d742041 100644
--- a/modules/base/markdown.go
+++ b/modules/base/markdown.go
@@ -346,7 +346,7 @@ OUTER_LOOP:
 func RenderMarkdown(rawBytes []byte, urlPrefix string, metas map[string]string) []byte {
 	result = PostProcessMarkdown(result, urlPrefix, metas)
-	result = Sanitizer.SanitizeBytes(result)
+	result = MarkdownSanitizer.SanitizeBytes(result)
 	return result
 }
 
diff --git a/modules/base/tool.go b/modules/base/tool.go
index c6522d1..d1ced27 100644
--- a/modules/base/tool.go
+++ b/modules/base/tool.go
@@ -31,4 +31,5 @@ import (
 	"github.com/gogits/gogs/modules/setting"
 )
 
-var Sanitizer = bluemonday.UGCPolicy().AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code").AllowElements("input").AllowAttrs("type", "checked", "disabled").OnElements("input")
+var Sanitizer = bluemonday.UGCPolicy().AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code")
+var MarkdownSaniziter = Sanitizer.AllowElements("input").AllowAttrs("type").Matching(regexp.MustCompile(`^checkbox$`)).OnElements("input").AllowAttrs("checked", "disabled").OnElements("input")
```

modules/base/markdown.go
```diff
@@ -142,6 +142,16 @@ func (r *CustomRender) AutoLink(out *bytes.Buffer, link []byte, kind int) {
 	r.Renderer.AutoLink(out, link, kind)
 }
 
+func (options *CustomRender) ListItem(out *bytes.Buffer, text []byte, flags int) {
+	switch {
+	case bytes.HasPrefix(text, []byte("[ ] ")):
+		text = append([]byte(`<input type="checkbox" disabled="" />`), text[3:]...)
+	case bytes.HasPrefix(text, []byte("[x] ")):
+		text = append([]byte(`<input type="checkbox" disabled="" checked="" />`), text[3:]...)
```
andreynering: I think the stardard style for attributes like `disabled` and `checked` is using the name as value:

```html
<input disabled="disabled" />
<input checked="checked" />
```

andreynering:
I didn't test it, but I'd like to have this feature, too. :+1:

modules/base/markdown.go
bkcsoft: Actually the standard allows for these:
```html
<input disabled />
<input disabled="" />
<input disabled="disabled" />
```
(and similar for checked ofc)
source: https://www.w3.org/TR/html-markup/input.checkbox.html#input.checkbox

Unknwon:
@bkcsoft agree that we should have two indepdent sanitizers.

bkcsoft:
Well, in the end it didn't really make any sense, since everything rendered by a template is passed through the Sanitizer, so for markdown it's actually being sanitized twice ^.^
And normally the markdown-sanitizer (user-content) would be more strict than the template-sanitizer (system-controlled-content), but in this case it was reversed (and in the end wrong)

modules/base/tool.go
```diff
@@ -31,7 +31,16 @@ import (
 	"github.com/gogits/gogs/modules/setting"
 )
 
-var Sanitizer = bluemonday.UGCPolicy().AllowAttrs("class").Matching(regexp.MustCompile(`[\p{L}\p{N}\s\-_',:\[\]!\./\\\(\)&]*`)).OnElements("code")
+func BuildSanitizer() (p *bluemonday.Policy) {
```
Unknwon: Why this to be a public function?

bkcsoft: Because I didn't know how golang worked with line-breaks, so I hade a function of it instead. I can revert that, though #2406 needs to do that anyhow to get the config-flag.

Unknwon: I mean, you can make it a private function...

bkcsoft: If you take a look at what I've done in #2406 you'll see why :) this is why I said that I could move this change (which isn't necessary for this PR) to #2406 which requires it

Unknwon: OK... why do you need build it every time you render?

bkcsoft: well, I don't, it's just that I have to wait for the config to be loaded before building the Sanitizer. That was the easiest way to do it :unamused:

Unknwon: This line? https://github.com/gogits/gogs/pull/2406/files#diff-45f69d540c7865155c0fea18b4e5050eR343

bkcsoft: No actually this line: https://github.com/gogits/gogs/pull/2406/files#diff-88e948dbecd728e2513ba9bafe876faaR40

Unknwon: Yes, so why would you read global config every time?

Just once on start time isn't enough?

bkcsoft: because "once on start time" is before the config is loaded, unless there's something I've missed :smile:

Unknwon: Then you could call this after config is loaded, instead of building same thing over and over again.

xiegeo: @bkcsoft  You can use [sync.Once](https://golang.org/pkg/sync/#Once) to initialization once on first use.

Unknwon:
Merging, thanks again!

bkcsoft:
Woho! \o/

PR #2398, opened 2016-01-13
commit.RefUrl expects AppUrl argument

nanoant:
This is fixup for ea375c0dcca118c8ac3c48ba569b025836ad5ccf. The bug was not
visible because commit.RefUrl was always returning empty url due regression
described in https://github.com/gogits/git-module/pull/4


Unknwon:
Thank you!

PR #2399, opened 2016-01-13
OS X launchd script

nanoant:
Using this script:

1. Copy scripts/launchd/io.gogs.web.plist into `/Library/LaunchDaemons`

2. The script assumes Gogs is running under `gogs` user and group, modify `/Library/LaunchDaemons/io.gogs.web.plist` if you want to user different user.

3. The script assumes Gogs is installed in `/Users/git/gogs`, modify `/Library/LaunchDaemons/io.gogs.web.plist` if you installed Gogs in different location.

4. Once you are sure that running Gogs manually via `gogs web` works fine, run it as a launchd service with:

         sudo launchctl load -F /Library/LaunchDaemons/io.gogs.web.plist

From now on launchd will ensure Gogs is running, eg. when system is restarted.


Unknwon:
Thank you!

Maybe you should make a PR to `docs` repository for these steps?

nanoant:
Yeah, will do.

Issue #2339, opened 2016-01-04
Release download file name doesn't include tag number


ddelpero:
When I create a new release, the zip or tar file that gets downloaded is project-commitnumber instead of tag name. For example at https://try.gogs.io/jh/gogs/releases, if you download the zip file for v0.8.10 you get a file named gogs-09c981846b.zip. However if you right click and copy link, you get the following link: https://try.gogs.io/jh/gogs/archive/v0.8.10.zip, which would seem to indicate that the file name should be named v0.8.10.zip and not gogs-09c981846b.zip.

ddelpero:
I'm fixing this in a fork, and I can submit a pull request to development. However, should the file name be projectname-releasename or only the releasename? For example, if project name is gogs and release is v0.8.10, should the file name be gogs-v0.8.10.zip or v0.8.10.zip?

Unknwon:
I think `gogs-v0.8.10.zip` makes more sense.

Be sure to read https://github.com/gogits/gogs/wiki/Contributing-Code before submitting the PR.

Unknwon:
This is claimed to be fixed by merging https://github.com/gogits/gogs/pull/2403

Unknwon:
Close as fixed.

PR #2403, opened 2016-01-13
Update repo.go

ddelpero:
Release download file name doesn't include tag number #2339
Download: Changed to use refName instead of commit.ID for downloaded file name


Unknwon:
Thank you!

PR #2406, opened 2016-01-14
Feature/markdown custom url scheme

bkcsoft:
~~This is a WIP-feature that needs rebasing after #2396 is merged~~

This adds the ability to define custom URL-Schemes (e.g. git,file,magnet) as acceptable links when rendering markdown

example `custom/conf/app.ini`:
```ini
[markdown]
CUSTOM_URL_SCHEMES = "git,file,magnet"
```

TODO:
- [x] implement it :laughing: 
- [x] wait for #2396 
- [x] add example to `conf/app.ini`

### DIFF
--- a/modules/base/markdown.go
+++ b/modules/base/markdown.go
@@ -340,7 +340,7 @@ OUTER_LOOP:
 func RenderMarkdown(rawBytes []byte, urlPrefix string, metas map[string]string) []byte {
 	result := RenderRawMarkdown(rawBytes, urlPrefix)
 	result = PostProcessMarkdown(result, urlPrefix, metas)
-	result = BuildSanitizer().SanitizeBytes(result)
+	result = Sanitizer.SanitizeBytes(result)
 	return result
 }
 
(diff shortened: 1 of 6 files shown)


Dasoren:
So much yes

bkcsoft:
@Unknwon any comments? :)

modules/base/markdown.go
```diff
@@ -346,7 +340,7 @@ OUTER_LOOP:
 func RenderMarkdown(rawBytes []byte, urlPrefix string, metas map[string]string) []byte {
 	result := RenderRawMarkdown(rawBytes, urlPrefix)
 	result = PostProcessMarkdown(result, urlPrefix, metas)
-	result = Sanitizer.SanitizeBytes(result)
+	result = BuildSanitizer().SanitizeBytes(result)
```
Unknwon: As mentioned in #2396, this should only build once (at start up) and reuse the same object `p`.

bkcsoft: Yes. And like I mentioned in #2396, there's no way to build it on start up, since the config isn't loaded then.
I'll try doing it another way and see if it works

Unknwon: There is a way by adding init in `NewConfigContext`.

Unknwon:
LGTM.

PR #2423, opened 2016-01-16
Fixed #2416

### LINKED ISSUE #2416: DisableMinimumKeySizeCheck is not a field of struct type interface {}
template: admin/config:82:38: executing "admin/config" at <.Service.DisableMini...>: DisableMinimumKeySizeCheck is not a field of struct type interface {}

I get that when I try to enter the configuration in "latest" docker tag. If I use 0.8.10 as docker tag, it works.
I chose blank ssh port and sqlite for db. Otherwise everything totally default.

exmex:


Unknwon:
Thanks your PR!

But commits and changes look very confusing?

Unknwon:
I think this PR was a mistake?

exmex:
Yes it was. Somehow the company intern files got pushed in here T_T

Unknwon:
Not the first time :joy:

PR #2424, opened 2016-01-16
Add command to run the test suite in Makefile.

andreynering:
`make test` will run the test files for all the packages.

`modules/httplib` is failing. But, if we can fix it, it would be a good idea to configure the CI to run the tests.


andreynering:
Wrong branch, sorry. :cry:

PR #2429, opened 2016-01-18
Expand split-diff view to full page width, as in GitHub.

cfallin:
I noticed when using the new split-diff view (from #2296 -- very useful, thanks) that lines in the diffs often wrapped, even with 80-column source. GitHub expands the diff container to (almost) full-screen width when in split-view mode. This is a simple template tweak to get the same behavior in Gogs.

There may be a better way to do this (adding a nice margin, also expanding the header/tabs to full width, maybe making this a config option) but I spend my day in C++/non-webdev stuff and couldn't work out how to tweak the LESS or the template contexts to do that -- my apologies. I hope this is at least a starting point!

### DIFF
--- a/templates/repo/diff.tmpl
+++ b/templates/repo/diff.tmpl
@@ -1,7 +1,7 @@
 {{template "base/head" .}}
 <div class="repository diff">
 	{{template "repo/header" .}}
-	<div class="ui container">
+	<div class="{{if $.IsSplitStyle}}ui{{else}}ui container{{end}}">
 		{{if .IsDiffCompare }}
 			{{template "repo/commits_table" .}}
 		{{else}}


Unknwon:
Thanks your PR!

@bkcsoft maybe can offer some help?

bkcsoft:
If you set it to `class="ui container fluid"` instead it goes full-width, but you get this instead :unamused:
2016-01-27-225251_1916x1041_scrot

bkcsoft:
so, setting it to `class="ui container fluid widen"` and define this in `public/less/gogs.less` would do:
```
.ui.container.fluid.widen {
  padding: 0 10px 0 10px;
}
```

bkcsoft:
@cfallin  If you want I can hijack this PR and make the necessary changes :)

cfallin:
@bkcsoft Feel free to hijack, as you likely know better what you're doing... thanks! :-)

Unknwon:
Closed by https://github.com/gogits/gogs/pull/2502

PR #2432, opened 2016-01-18
Fix #2431 - handle requests waiting for reply

### LINKED ISSUE #2431: TeamCity CI server fails fetch from gogs
TeamCity uses java git and ssh implementation (JGit & Jsch). When it tries to list refs of the remote repository hosted by Gogs it fails with timeout exception. Debugging shows that Jsch sets WantReply flag for its ssh requests and waits for reply from Gogs but never gets one. It seems like Gogs doesn't handle ssh requests with WantReply flag set to true.

Gogs Version: v0.8.10 @ 2015-12-18, as well as Gogs built from master branch
Git Version: not applicable since TeamCity uses JGit - java git implementation
System Type: all systems
Error Log: there is no errors in the Gogs logs

nd:
According to the docs [1], the Reply method must be called for all
requests where WantReply is true. This fixes a hanging java ssh
implementation (jsch) which sets WantReply flag and waits for reply from
the server.

[1] https://godoc.org/golang.org/x/crypto/ssh#Request.Reply

### DIFF
--- a/modules/ssh/ssh.go
+++ b/modules/ssh/ssh.go
@@ -91,6 +91,7 @@ func handleServerConn(keyID string, chans <-chan ssh.NewChannel) {
 						return
 					}
 
+					req.Reply(true, nil);
 					go io.Copy(input, ch)
 					io.Copy(ch, stdout)
 					io.Copy(ch.Stderr(), stderr)


Unknwon:
Thank you very much!

PR #2439, opened 2016-01-19
Formatting and update copyrights

farazfazli:
Fix formatting using ```gofmt -s -w .``` and update copyright notices.

### DIFF
--- a/cmd/cert.go
+++ b/cmd/cert.go
@@ -1,7 +1,7 @@
 // +build cert
 
 // Copyright 2009 The Go Authors. All rights reserved.
-// Copyright 2014 The Gogs Authors. All rights reserved.
+// Copyright 2016 The Gogs Authors. All rights reserved.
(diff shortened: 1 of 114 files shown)


bkcsoft:
You should probably read this ^.^ http://stackoverflow.com/a/2391555

andreynering:
Maybe it should be `2014-2016`.

farazfazli:
Yeah, I will change it to 2014-2016, that's better.

bkcsoft:
has there been substantial change in all those files over 2015/16? otherwise you should just leave them at 2014. The year is kinda m00t for something that's tracked on github though :smile:

farazfazli:
@Unknwon What are your thoughts?

Unknwon:
Hi @farazfazli , thanks your PR!

But Gogs uses the "Go"'s way, which only states the start year of the file.

Unknwon:
Close due to lack of feedback.

PR #2467, opened 2016-01-24
Fixed forgotten err variable assignment

pdan:


Unknwon:
Good catch! Thank you!

PR #2473, opened 2016-01-25
use another gravatar url to bypass firewall

MilesPong:
```secure.gravatar.com ``` should be more suitable

### DIFF
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -415,7 +415,7 @@ func NewContext() {
 	case "duoshuo":
 		GravatarSource = "http://gravatar.duoshuo.com/avatar/"
 	case "gravatar":
-		GravatarSource = "//1.gravatar.com/avatar/"
+		GravatarSource = "//secure.gravatar.com/avatar/"
 	default:
 		GravatarSource = source
 	}

### CODE REVIEW

**Comment by Unknwon** [comment]
Thanks your PR!

Any info of history between `1.gravatar.com` and `secure.gravatar.com`?

**Comment by MilesPong** [comment]
You could check the instructions of ```secure.gravatar.com``` from [Gravatar](https://en.gravatar.com/site/implement/images/)
In the Secure Requests section, ```secure.gravatar.com``` is recommended while the scheme is https, but it works the same when using http after my test.
For more infos, you may want to check this [website](http://blog.aofall.xyz/index.php/archives/2015/05/about-gravatar.html)
That's all.

**Comment by Unknwon** [comment]
Verified, thanks!

However, according to https://github.com/gogits/gogs/wiki/Contributing-Code#final-checks

Please re-PR to `develop` branch.

File: modules/setting/setting.go [hunk 1]
```diff
@@ -415,7 +415,7 @@ func NewContext() {
 	case "duoshuo":
 		GravatarSource = "http://gravatar.duoshuo.com/avatar/"
 	case "gravatar":
-		GravatarSource = "//1.gravatar.com/avatar/"
+		GravatarSource = "//secure.gravatar.com/avatar/"
```
cweagans: Protocol-agnostic URLs aren't needed anymore. Just use `https://` unconditionally if it's available. See the note at the top of http://www.paulirish.com/2010/the-protocol-relative-url/ for more detail.

Unknwon: Good catch! Welcome to make a PR for it. :wink:

MilesPong:
What do you mean by adding the `wontfix` label? Will not accept the suggestion? So there is no need for me to re-PR to `develop` branch ?

Unknwon:
You need to re-PR to `develop` branch as documented. :D

PR #2475, opened 2016-01-25
Update Dockerfile & build script and add /etc/nsswitch.conf

### LINKED ISSUE #2015: gogs in Docker container uses Host IP trying to access linked DB container
Hi, I'm trying to run gogs with this `docker-compose.yml`, linking it with a MariaDB container:

    gogs:
      image: gogs/gogs
      ports:
        - "10022:22"
        - "8080:3000"
      links:
        - db

    db:
      image: mariadb
      environment:
        - MYSQL_ROOT_PASSWORD=rootpass
        - MYSQL_DATABASE=gogs
        - MYSQL_USER=gogs
        - MYSQL_PASSWORD=gogs

On `localhost:8080` I configure gogs to use these database settings, but always get the error message shown:

screenshot from 2015-11-22 11-47-02

`77.37.17.149` is the IP of my router, which means that inside the gogs container a hostname is used for the DB which cannot be resolved (which then, via a catch-all, is resolved to the router). This is supported by the fact that the linked `db` container resolves (correctly) to a `172.*` address from inside the gogs container:

    $ docker exec -ti gogs_gogs_1 bash
    bash-4.3# ping db
    PING db (172.17.0.2): 56 data bytes
    64 bytes from 172.17.0.2: seq=0 ttl=64 time=0.404 ms
    64 bytes from 172.17.0.2: seq=1 ttl=64 time=0.185 ms
    ^C

Also, MariaDB is clearly reachable from within the gogs container:

    bash-4.3# telnet db 3306
    5.5.5-10.0.21-MariaDB-1~jessie1`(....)

So what hostname does the gogs container use trying to connect to the database? It's not `db`, or else it would work. Any ideas?

0rax:
- Add nsswitch.conf to configure LibC Name Service inside the container (Resolve #2015)
- Change my email in the Dockerfile
- Update build script to install software as a `build-deps` virtual package so that adding a package to it will be automatically removed at the end of the build script

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:3.2
-MAINTAINER roemer.jp@gmail.com
+MAINTAINER jp@roemer.im
 
 # Install system utils & Gogs runtime dependencies
 ADD https://github.com/tianon/gosu/releases/download/1.6/gosu-amd64 /usr/sbin/gosu
@@ -14,6 +14,9 @@ ENV GOGS_CUSTOM /data/gogs
 COPY . /app/gogs/
 WORKDIR /app/gogs/
 RUN ./docker/build.sh
+
+# Configure LibC Name Service
+COPY docker/nsswitch.conf /etc/nsswitch.conf
 
 # Configure Docker Container
 VOLUME ["/data"]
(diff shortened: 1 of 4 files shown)


Unknwon:
Thanks!

PR #2480, opened 2016-01-25
Compute inline diff for pull request view, too. Fix #2462

### LINKED ISSUE #2462: Comparing commits results in an empty changelog
A picture is worth a thousand words; I would expect that text would be present here. 
screenshot from 2016-01-22 15-48-08

Aditional information: 
Version: 0.8.22.0115

This only happens when clicking on the link 'View comparison for these x commits' from the dashboard. The differences from the individual commits does work as expected.

andreynering:


Unknwon:
Thanks!

PR #2493, opened 2016-01-27
Refactoring of inline diff computing to prevent empty diff box.

### LINKED ISSUE #2489: Pull request diff has no text
Gogs Version: 0.8.19.0111

When creating a pull request I don't see any text in the diff at the bottom of the page. The diff colouring appears correct however, were the text to be there.

s1

Whereas the actual commit diff shows:

s2

In my local gogs installation, the same problem happens after the pull request has been created and you view the *Files Changed* tab for the pull request. I've just tested this on https://try.gogs.io (which is slightly later version (0.8.23.0126) to my local install) and that shows the diff correctly **after** the pull request is created, but I can still reproduce the problem on https://try.gogs.io for the initial pull request create/Compare Changes.

andreynering:
Fix #2489

### DIFF
--- a/models/git_diff.go
+++ b/models/git_diff.go
@@ -51,7 +51,6 @@ type DiffLine struct {
 	RightIdx      int
 	Type          DiffLineType
 	Content       string
-	ParsedContent template.HTML
 }
 
 func (d *DiffLine) GetType() int {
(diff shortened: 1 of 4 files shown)


Unknwon:
Thanks a lot!

PR #2496, opened 2016-01-27
Split view expanded

bkcsoft:
Shamelessly highjacking #2429

@Unknwon not sure if I should expand more things?


Unknwon:
Thanks, but the wrong branch...

PR #2501, opened 2016-01-28
enable basic TLS for mysql connection without server CA cert validation

askannon:
warning: I am a total GO noob.
However I would love to see an options for the mysql driver to use TLS security - even if only without certificate verification. I found an option to enable it here: https://github.com/go-sql-driver/mysql#tls
and created a little patch reusing the SSLMode setting already used for Postgres. If set to "skip-verify" it should try to use TLS without verifying the remote CA cert.

Thanks

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -123,12 +123,16 @@ func getEngine() (*xorm.Engine, error) {
 	cnnstr := ""
 	switch DbCfg.Type {
 	case "mysql":
+		tls := ""
+		if DbCfg.SSLMode == "skip-verify" {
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR!

I got the idea, I would suggest you open up an issue for this feature request.

askannon:
Thanks, I created [Issue 2503](https://github.com/gogits/gogs/issues/2503) for this.

Unknwon:
Closed by #2503 as we're going to support full feature of https://github.com/go-sql-driver/mysql#tls .

PR #2517, opened 2016-01-30
Allow modification of a release if Content is empty (fix #2516)

### LINKED ISSUE #2516: Ability to edit a release if the content field is empty
I can create a release with an empty content but not edit an existing one, if content is empty. It should be possible to save a release, even if the content field is empty.

gogs-release-content-bug

Gogs Version: 0.8.10.1217 (stable)
Git Version: 2.6.2
System Type: prod

fnkr:


Unknwon:
Thanks a lot!

Issue #1655, opened 2015-09-17
Add possibility to checkout pull request locally 


hhenkel:
There are various reason why one might want to checkout a pull request locally.
- When there is conflicting pull request it should be possible to checkout the request itself to resolve the issue locally.
- It's enabling people to configure there CI solution to checkout and test the pull request (and comment the result back to gogs once the api is ready).
- The original author might have deleted his fork of the repo, but one might still want to accept the pull request.

This is also possible on other solutions like github - [merging a pull request](https://help.github.com/articles/merging-a-pull-request/#checking-out-a-pull-request-locally) / [checking out pull requests locally](https://help.github.com/articles/checking-out-pull-requests-locally/) or [gitblit](http://gitblit.com/) which has something called [tickets](http://gitblit.com/tickets_overview.html).

Unknwon:
Thanks your feedback!

hhenkel:
Just recognized that it might be worth mentioning, that it is possible to show all the remote refs of a repository via "git ls-remote". In the case of github (gogs repo) this shows refs to all pulls (issues?) besides the obvious stuff:

```
$ git ls-remote
From https://github.com/gogits/gogs.git
eec06fb3df7a9dc30f758bb1a100b5f1e8e01ab9        HEAD
bd5dc626e82e18d3e619d918e579dc130edcd1fa        refs/heads/develop
10490f1d372ee538231ce3a378e3dd0ec23fef7b        refs/heads/feature/comment
f2ab68e2201e7ddc83fc6363caadb4893bd42fd7        refs/heads/git
eec06fb3df7a9dc30f758bb1a100b5f1e8e01ab9        refs/heads/master
22bb5104c8a0d585ea0e538410af1044bc2c1505        refs/pull/100/head
e63624ca0ecf1052f25e11ff06c6e6b52c0deafb        refs/pull/1001/head
04565ed5deb502c70a964365b1531d681901ae35        refs/pull/1001/merge
698ce99a646f0b9f4224abea5fbdb34278cc305a        refs/pull/1005/head
d98516899e4344a11fa6230d3eacbe8fb5686b30        refs/pull/1007/head
6f094b9829b485fdb88e439263b9bb8bcd2429e5        refs/pull/1007/merge
97b57f085d7231dab99a283600e6851ed14fe30e        refs/pull/1019/head
767e50b2648103b39430219e596ce6a82f0ddfd1        refs/pull/1019/merge
...
...
...
[...]

Issue #1623, opened 2015-09-11
Takes control of Git hooks


Unknwon:
Use Gogs's command to delegate hooks, make many features possible:

1. Reduce HTTP push memory usage by detect old/new commit and ref name in `post-receive` hook https://github.com/gogits/gogs/issues/636
2. Setting Gogs's specific environment variables for custom hooks https://github.com/gogits/gogs/issues/4094
3. Protect branch and prevent force push https://github.com/gogits/gogs/issues/776
3. Users are able to use `update` hook https://github.com/gogits/gogs/issues/1793

busterbot:
test comment

PR #2524, opened 2016-01-30
Enable a way to checkout Pull Requests from remote refs

mhartkorn:
This Pull Request adds a new feature described in #1655. Having this feature would be very nice, especially for Go projects where the imports are often global and need to be manually changed to test Pull Requests. It is work in progress and needs feedback.

To make it work, the patch from https://github.com/gogits/git-module/pull/5 needs to be applied to the imports. This is the reason why the Travis-CI build is currently failing.

##### How it works

When a user pushes commits to a branch gogs will find out if it's part of an open Pull Request. If it is, gogs will create a temporary remote to the base repository and push the changes to `refs/pull/<pr-number>/head`. Everyone with permission can then checkout the code from the Pull Request as described here: [Checking out pull requests locally](https://help.github.com/articles/checking-out-pull-requests-locally/)

##### Current status
- [x] Push commits to the base repository into `refs/pull/<pr-number>/head`
- [x] Block cyclic dependencies (not really a problem because of git objects)
- [x] ~~Forbid manual pushes to `refs/pull/*`~~ (handle it like Gitlab and simply override `refs/pull/<id>/head` if already exists)

Please give feedback on the method I'm using as well as on the coding style.


bkcsoft:
Don't like the last checkbox, since that could still be useful for cherrypicking fixes.

Example:
We have `master`, which has merged pr#2 after creating the tag `v1.0` and no new `v1.1` has been made after the merge. Now if my prod-env needs pr#2 I can still pull (and merge) that one, without having to pull and merge `master`

bkcsoft:
Otherwise :100:

Unknwon:
Quick question: why do we need post-receive instead of update hook?

mhartkorn:
The `update` hook is called before the commits are added to the repository, so if I start a push there, the most recent commits are missing. I searched for the most fitting hook for this feature and `post-receive` was named as the most fitting to do deployments/send the code somewhere.

Maybe the update hook can tweaked for this and I missed something. I would rather have it in the `update` hook but I wasn't able to find out how to do it.

mhartkorn:
@bkcsoft: Good idea, I will remove the last point.

cmd/pull.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2015 The Gogs Authors. All rights reserved.
```
andreynering: Should be 2016?

andreynering:
:+1: on this. It's very useful.

cmd/pull.go
mhartkorn: Whoops, I will change it in the next commit.

Unknwon:
@mhartkorn 

>  The update hook is called before the commits are added to the repository, so if I start a push there, the most recent commits are missing.

Update hook itself receives latest commits already... does not need to check with repository again.

Unknwon:
Not commits, but the start and end commits SHA, see https://github.com/gogits/gogs/blob/master/cmd/update.go#L47-L49

> How it works

When a user pushes commits to a branch gogs will find out if it's part of an open Pull Request. If it is, gogs will create a temporary remote to the base repository and push the changes to refs/pull/<pr-number>/head. Everyone with permission can then checkout the code from the Pull Request as described here: Checking out pull requests locally

Not sure, just an idea: why just base repository simply add a remote points to the branch? Never tried, I'll test out.

mhartkorn:
@Unknwon I'm not 100% sure but as far as I understood the git docs, git is still able to reject the push after `update`, isn't it?

Unknwon:
>  I'm not 100% sure but as far as I understood the git docs, git is still able to reject the push after update, isn't it?

Yes, why does it matter

mhartkorn:
@Unknwon I think Pull Requests should only have commits in the base repo (as `refs/pull/*`) that are also in the head repo.

Unknwon:
> @Unknwon I think Pull Requests should only have commits in the base repo (as refs/pull/*) that are also in the head repo.

Why it needs to be pushed?

Unknwon:
I think we may need to come up a feature design to share our understandings?

mhartkorn:
I can give you a rundown on the thought process I had while I programmed this feature.

##### How it currently works in the code

- Receive commit from client thorugh HTTP(S)/SSH
- Git applies the commit to the repository
- Gogs gets notified through the `post-receive` hook that new commits are in the repository
- Gogs checks if the branch of the commits currently is associated with in an open Pull Request
- If yes:
  - Locate the base repository via the file system (using `pwd` and relative paths)
  - Create a temporary remote in the head repository to the base repository
  - Push the branch that is associated with the Pull Request to the base repository into the ref `refs/pull/<pull-request-number>/head`
  - Remove the temporary remote
- If no:
  - Do nothing

##### Why I used the `post-receive` instead of `update`

First I thought I can use in `update`, which is called for each pushed ref with the parameters `old-rev`, `new-rev`, and `ref-name`. But as far as I know `update` does not guarantee that the pushed refs are actually able to be applied to the repository. So if somebody pushes to a wrong branch by mistake, the push gets eventually rejected by git. So I thought, that `post-receive` would be a better place for this. I also noticed that it was not so easy for me to actually get the contents of the commits in `update` to send them to the base repository.

I would be glad if I am wrong here, and can actually use `update` because it reduces complexity.

Unknwon:
Thanks for the great detail!

> So if somebody pushes to a wrong branch by mistake, the push gets eventually rejected by git.

I don't think `update` hook would be called in this case.

Unknwon:
> Push the branch that is associated with the Pull Request to the base repository into the ref refs/pull/<pull-request-number>/head

Why this step is required?

mhartkorn:
> In my understanding, we could just update the refs/pull/ID/head record. When user tries to pull, Gogs knows which repository and branch this record points to and just response data from there.

That was my first thought, too. But I wasn't able to do it because git was always complaining that it was not able to find the commit IDs. Maybe it would be possible if all git related thing were tunneled through a service that then identifies specific use cases (checkout of pull requests, cyclic dependencies between PRs) and then handles them accordingly. This would increase the complexity a lot, though.

I researched how Gitlab is handling their *merge requests* and as far as I can see, they do the same as my current implementation does. They just add a so called *gitlab-shell* that every git related data is getting through to handle some stuff (like denying a manual push to `refs/merge-requests/<id>/head`).

Regarding the problem with cyclic dependencies I came to the conclusion that they are not really a problem. Git stores everything as objects and if there are cyclic dependencies they are just references between each other that do not significantly increase the size of the repository.

mhartkorn:
I did a bit more research and as far as I can see, my implementation in the current state is the minimal version that does not need a permanently running service that handles git access. If there was such a service between the client and git, there would be more possibilities to restrict access to e.g. the refs for Pull Requests. An example for such a service is the [gitlab-shell](https://gitlab.com/gitlab-org/gitlab-shell) that can transparently handle permissions.

Unknwon:
OK... I think I can agree on push content, but I'm trying to avoid using `post-receive` hook.

What I would do is adding push actions [here](https://github.com/gogits/gogs/blob/master/routers%2Frepo%2Fpull.go#L646).

This trigger won't be called if the user push action was not successful (decline to push, etc).

mhartkorn:
@Unknwon Ok, I will look into it and add a commit to this PR in the next few days.

mhartkorn:
@Unknwon Using `TriggerTask` was a very good idea. The code is much cleaner and more robust now.

Unknwon:
Great, I'll review the push step detials ASAP.

JohnMaguire:
May I ask why you are trying to avoid using `post-receive`? I think it makes more sense here:

`update`:
> This hook is invoked by git-receive-pack on the remote repository, which happens when a git push is done on a local repository. Just before updating the ref on the remote repository, the update hook is invoked. Its exit status determines the success or failure of the ref update.

`post-receive`:
> This hook is invoked by git-receive-pack on the remote repository, which happens when a git push is done on a local repository. It executes on the remote repository once after all the refs have been updated.

There is added safety here by using `post-receive`. `hook` is typically used to determine *whether* something can be pushed, while `post-receive` is for making an action *after* something has been pushed. In this case we only want to update the pull request if the forked repository is updated.

Unknwon:
@JohnMaguire Gogs currently does not delegate (will be https://github.com/gogits/gogs/issues/1623) hooks but directly write content into it.

So if Gogs uses `post-receive`, then users are not able to use this hook themselves anymore, AFAIK many users do use this hook for automations.

mhartkorn:
@Unknwon Did you find time to look into it?

I checked point 3 again and it turns out that Gitlab doesn't protect their `refs/merge-requests/<id>/head` (their equivalent to `refs/pull/<id>/head`) against wrong pushes. So Gogs could do the same in my opinion.

Unknwon:
> I checked point 3 again and it turns out that Gitlab doesn't protect their refs/merge-requests/<id>/head (their equivalent to refs/pull/<id>/head) against wrong pushes. So Gogs could do the same in my opinion.

Sure, it is the easier part in my opinion.

mhartkorn:
Nice, thanks, I'll tick the box above then and leave a comment about it.

mhartkorn:
Sorry if I sound annoying but were you able to review this PR @Unknwon ? I think it's nice to have for Go development because it makes testing PRs easier.

Unknwon:
Will do ASAP. Sorry about that :mask:

Unknwon:
Sorry about the delay, merging now, thanks again!

PR #2523, opened 2016-01-30
Show all orgs on user profile, except the private one's

fnkr:
Previously, the profile page of an user showed all organisations the user owned (regardless if public or not), which makes no sense. After merging this, the profile page will show all organisations the user has joined (not just the one's he owns), except the private (=not public) one's. Just like on GitHub :-)

gogs-private-orgs-on-profile

PR #2527, opened 2016-01-31
Show all orgs on user profile, except the private one's

fnkr:
*Re-PR of #2523 because of wrong branch.*

---

Previously, the profile page of an user showed all organisations the user owned (regardless if public or not), which makes no sense. After merging this, the profile page will show all organisations the user has joined (not just the one's he owns), except the private (=not public) one's. Just like on GitHub :-)

gogs-private-orgs-on-profile

### DIFF
--- a/models/org.go
+++ b/models/org.go
@@ -253,6 +253,26 @@ func IsPublicMembership(orgId, uid int64) bool {
 	return has
 }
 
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks again!

Unknwon:
Updated on https://try.gogs.io/

PR #2528, opened 2016-01-31
Enable syntax highlighting on diff view

### LINKED ISSUE #733: Syntax highlighted diffs
It would be really nice if gogs implemented this new github feature:

https://github.com/blog/1932-syntax-highlighted-diffs

Given there's already syntax highlighting in the code view, I'd imagine this wouldn't be too hard to hack in.

andreynering:
Closes #733

### DIFF
--- a/models/git_diff.go
+++ b/models/git_diff.go
@@ -26,6 +26,7 @@ import (
 	"github.com/gogits/gogs/modules/base"
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/process"
+	"github.com/gogits/gogs/modules/template/highlight"
 )
 
 type DiffLineType uint8
@@ -160,10 +161,18 @@ type DiffFile struct {
 	IsBin              bool
 	IsRenamed          bool
 	Sections           []*DiffSection
+	HighlightClass     string
 }
 
 func (diffFile *DiffFile) GetType() int {
 	return int(diffFile.Type)
+}
+
+func (diffFile *DiffFile) GetHighlightClass() string {
+	if diffFile.HighlightClass == "" {
+		diffFile.HighlightClass = highlight.FileNameToHighlightClass(diffFile.Name)
+	}
+	return diffFile.HighlightClass
 }
 
 type Diff struct {
--- a/models/git_diff_test.go
+++ b/models/git_diff_test.go
@@ -1,70 +1,70 @@
 package models
 
 import (
-  dmp "github.com/sergi/go-diff/diffmatchpatch"
-  "html/template"
-  "testing"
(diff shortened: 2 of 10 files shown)


public/js/gogs.js
```diff
@@ -830,6 +830,10 @@ $(document).ready(function () {
 
     // Highlight JS
     if (typeof hljs != 'undefined') {
+        // creating aliases for languages
```
Unknwon: Why this is needed?

Unknwon:
Any screenshots?

andreynering:
> Any screenshots?

Sure.

Before:
wo1
After:
w1


Before:
wo2
After:
w2

public/js/gogs.js
andreynering: > Why this is needed?

Highlight.js tries to detect the language automatically. But in this particular case, it didn't work as expected. So, I'm explicity setting the language to Highlight.js, as a `language-{{language-extension}}` class. Unfortunally, not all languages are being reconized by its file extension ([see](http://highlightjs.readthedocs.org/en/latest/css-classes-reference.html#language-names-and-aliases)). So I manually created aliases to two cases I found, but for most languages it isn't necessary.

Unknwon: Maybe you can use of this? https://github.com/gogits/gogs/blob/master/modules/template/highlight.go

andreynering: I didn't know that it was already implemented. ¯\_(ツ)_/¯ Updated to use that package, much better.

I also did this patch to Highlight.js, so futures version will have these alises by default: https://github.com/isagalaev/highlight.js/pull/1069/files

There was just a little gotcha: I had to move `highlight.go` to its own package, because we got a cycle import, which isn't allowed in Go, on importing `template` in `models` since `models` also import `template`.

Unknwon:
Takes a bit more time to review :trollface:

public/plugins/highlight-9.1.0/github.css
```diff
@@ -7,9 +7,9 @@ github.com style (c) Vasily Polovnyov <vast@whiteants.net>
 .hljs {
   display: block;
   overflow-x: auto;
-  padding: 0.5em;
+  /*padding: 0.5em;*/
```
Unknwon: I think overwrite this value in `_base.less` is better as we can't care very well when upgrading the plugins.

andreynering: @Unknwon Makes sense. Done!

Unknwon:
Thanks!

Unknwon:
Try demo https://try.gogs.io/gogs/gogs/commit/f15a2f9b25bbec130915fec5b33c7e49dc7cf57d

PR #2530, opened 2016-01-31
Hide other teams & repos from organization page

fnkr:
Users should only be able to see repositories & teams they can access / they have joined. Right now, a user sees all repositories of an organization but when he clicks on it, a 404 page comes up. After merging this, users will only be able to see the teams & repos, they have access to. Just like on GitHub & Co. :-)

screenshot from 2016-01-31 19-38-48

Explanation of screenshot:
```
User left:
- is Owner of org2
- can see all repos
- can see all teams

User right:
- only in Team2
- can see public repos
- can see repos of Team2
- can see team Team2

Anyone:
- can see public repos
```

### DIFF
--- a/models/org.go
+++ b/models/org.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"strconv"
 
 	"github.com/go-xorm/xorm"
 )
@@ -1039,14 +1040,24 @@ func (org *User) getUserRepositories(userID int64) (err error) {
 		return fmt.Errorf("getUserRepositories: get teams: %v", err)
 	}
 
-	var teamIDs []int64
+	var teamIDs []string
 	for _, team := range teams {
-		teamIDs = append(teamIDs, team.ID)
+		teamIDs = append(teamIDs, strconv.FormatInt(team.ID, 10))
+	}
+	if len(teamIDs) == 0 {
+		// user has no team but "IN ()" is invalid SQL
+		teamIDs = append(teamIDs, "0")  // there is no repo with id=0
 	}
 
+	// Due to a bug in xorm using IN() together with OR() is impossible.
+	// As a workaround, we have to build the IN statement on our own, until this is fixed.
+	// https://github.com/go-xorm/xorm/issues/342
+
 	if err := x.Cols("`repository`.*").
-				In("`team_repo`.team_id", teamIDs).
 				Join("INNER", "`team_repo`", "`team_repo`.repo_id=`repository`.id").
+				Where("`repository`.owner_id=?", org.Id).
+				And("`repository`.is_private=?", false).
+				Or("`team_repo`.team_id=(?)", strings.Join(teamIDs, ",")).
 				GroupBy("`repository`.id").
 				Find(&org.Repos); err != nil {
 		return fmt.Errorf("getUserRepositories: get repositories: %v", err)
--- a/models/org.go
+++ b/models/org.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"strconv"
 
 	"github.com/go-xorm/xorm"
 )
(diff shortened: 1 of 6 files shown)


routers/org/teams.go
```diff
@@ -28,10 +28,7 @@ func Teams(ctx *middleware.Context) {
 	ctx.Data["Title"] = org.FullName
 	ctx.Data["PageIsOrgTeams"] = true
 
-	if err := org.GetTeams(); err != nil {
-		ctx.Handle(500, "GetTeams", err)
-		return
-	}
+	// org.Teams is already loaded by middleware
```
bkcsoft: Is this comment really necessary?

models/org.go
```diff
@@ -1028,3 +1029,67 @@ func removeOrgRepo(e Engine, orgID, repoID int64) error {
...
+		return fmt.Errorf("getUserRepositories: get teams: %v", err)
+	}
+
+	var teamIDs []string
+	for _, team := range teams {
+		teamIDs = append(teamIDs, strconv.FormatInt(team.ID, 10))
+	}
+	if len(teamIDs) == 0 {
+		// user has no team but "IN ()" is invalid SQL
+		teamIDs = append(teamIDs, "0")  // there is no repo with id=0
```
bkcsoft: maybe set it to "-1" instead? which is an invalid key... one can set auto_increment to 0 and all of a sudden we have a bug :D

fnkr: You're absolutely right.

fnkr:
The issue and pull request overview works now too (only shows issues/PRs for repos the user access to).

screenshot from 2016-01-31 21-09-25
screenshot from 2016-01-31 21-09-53

Unknwon:
I suggest split this into two PRs.

fnkr:
The last commit is just a minor change and depends on the rest. I think this PR is complete now.

Unknwon:
Oh, OK. I'll review ASAP.

fnkr:
Won't you remove the `feedback needed` label?

Unknwon:
Done :D

fnkr:
Hi, any updates on this?

Unknwon:
It's waiting inline...

models/org.go
```diff
@@ -1028,3 +1029,67 @@ func removeOrgRepo(e Engine, orgID, repoID int64) error {
...
+	if len(teamIDs) == 0 {
+		// user has no team but "IN ()" is invalid SQL
+		teamIDs = append(teamIDs, "-1")  // there is no repo with id=-1
+	}
+
+	// Due to a bug in xorm using IN() together with OR() is impossible.
+	// As a workaround, we have to build the IN statement on our own, until this is fixed.
+	// https://github.com/go-xorm/xorm/issues/342
+
+	if err := x.Cols("`repository`.*").
```
Unknwon: Will the SQL run on MySQL, PostreSQL and SQLite3?

models/org.go
```diff
@@ -1028,3 +1029,67 @@ func removeOrgRepo(e Engine, orgID, repoID int64) error {
...
+
+	org.NumRepos = len(org.Repos)
+
+	return
+}
+
+// GetUserRepositories gets all repositories of an organization,
+// that the user with the given userID has access to.
+func (org *User) GetUserRepositories(userID int64) error {
+	return org.getUserRepositories(userID)
```
Unknwon: What is primary purpose of making a sub-function?

models/org.go
```diff
@@ -1028,3 +1029,67 @@ func removeOrgRepo(e Engine, orgID, repoID int64) error {
...
+
+	org.NumTeams = len(org.Teams)
+
+	return
+}
+
+// GetTeams returns all teams that belong to organization,
+// and that the user has joined.
+func (org *User) GetUserTeams(userID int64) error {
+	return org.getUserTeams(userID)
```
Unknwon: Same as above.

routers/user/home.go
```diff
@@ -314,23 +321,27 @@ func showOrgProfile(ctx *middleware.Context) {
...
+		if err := org.GetUserRepositories(ctx.User.Id); err != nil {
+			ctx.Handle(500, "GetUserRepositories", err)
+			return
+		}
+		ctx.Data["Repos"] = org.Repos
+	} else {
+		if repos, err := models.GetRepositories(org.Id, false); err != nil {
+			ctx.Handle(500, "GetRepositories", err)
+			return
+		} else {
```
Unknwon: No need for `else` if `if` case uses `return`.

Unknwon:
Otherwise, looks good. Thanks!

routers/user/home.go
fnkr: `repos` is a scoped variable and only available within the if/else case. I could do the following, but it would create variables, that do exist outside of the area where they are used.
```go
var repos *models.Repos
var err error
if repo, err = org.GetUserRepositories(ctx.User.Id); err != nil {
    ctx.Handle(500, "GetUserRepositories", err)
    return
}
ctx.Data["Repos"] = repos
```

models/org.go
fnkr: Is not necessary, but a lot of other functions did that too.

models/org.go
fnkr: Haven't tested it with PostgreSQL and SQLite3, will do that ASAP.

routers/user/home.go
Unknwon: You could also do:

```go
repo, err := org.GetUserRepositories(ctx.User.Id);
if err != nil {
    ctx.Handle(500, "GetUserRepositories", err)
    return
}
ctx.Data["Repos"] = repos
```

This block lives inside another else block, so `that do exist outside of the area where they are used.` isn't true.

models/org.go
Unknwon: OK.. I think that's because they involves `session` object, in this case there are only read operations, which does not hurt anything on data if it fails at any given point.

models/org.go
fnkr: Tested with no issues with MySQL, PostgreSQL & SQLite3.

Unknwon: Great, thanks.

routers/user/home.go
fnkr: Oh, sorry, you're absolutely right. Thanks!

Fixed!

PR #2533, opened 2016-01-31
Use icon repo-forked instead of repo-lock for private, forked repos

fnkr:
Makes more sense, because private repositories still do have the yellow background.

**Before**:
screenshot from 2016-01-31 21-39-29

**After**:
screenshot from 2016-01-31 21-40-23

### DIFF
--- a/templates/explore/repo_list.tmpl
+++ b/templates/explore/repo_list.tmpl
@@ -3,10 +3,10 @@
 		<div class="item">
 			<div class="ui header">
 				<a class="name" href="{{AppSubUrl}}/{{if .Owner}}{{.Owner.Name}}{{else if $.Org}}{{$.Org.Name}}{{else}}{{$.Owner.Name}}{{end}}/{{.Name}}">{{if $.PageIsExplore}}{{.Owner.Name}} / {{end}}{{.Name}}</a>
-				{{if .IsFork}}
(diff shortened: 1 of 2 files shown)


fnkr:
Btw. GitHub does the same :D

templates/explore/repo_list.tmpl
```diff
@@ -3,10 +3,10 @@
 		<div class="item">
 			<div class="ui header">
 				<a class="name" href="{{AppSubUrl}}/{{if .Owner}}{{.Owner.Name}}{{else if $.Org}}{{$.Org.Name}}{{else}}{{$.Owner.Name}}{{end}}/{{.Name}}">{{if $.PageIsExplore}}{{.Owner.Name}} / {{end}}{{.Name}}</a>
-				{{if .IsPrivate}}
-					<span class="text gold"><i class="icon octicon octicon-lock"></i></span>
-				{{else if .IsFork}}
+				{{if .IsFork}}
```
Unknwon: I think should still show its private first.

We can add a small line to say where it forks from, but after changed this, we can't tell it is private or not.

fnkr: @Unknwon I reverted the second changeset.

Unknwon:
Thanks!

Unknwon:
Updated: https://try.gogs.io/

PR #2537, opened 2016-02-01
Remember last selected clone protocol

fnkr:
Default is HTTPS.

remember-clone-protocol
remember-clone-protocol2
remember-clone-protocol3

### DIFF
--- a/public/js/gogs.js
+++ b/public/js/gogs.js
@@ -434,12 +434,14 @@ function initRepository() {
         $('#repo-clone-url').val($(this).data('link'));
         $(this).addClass('blue');
         $('#repo-clone-https').removeClass('blue');
+        localStorage.setItem('repo-clone-protocol', 'ssh');
     });
     $('#repo-clone-https').click(function () {
         $('.clone-url').text($(this).data('link'));
         $('#repo-clone-url').val($(this).data('link'));
         $(this).addClass('blue');
         $('#repo-clone-ssh').removeClass('blue');
+        localStorage.setItem('repo-clone-protocol', 'https');
     });
     $('#repo-clone-url').click(function () {
         $(this).select();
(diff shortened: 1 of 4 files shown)


Unknwon:
Thanks. Maybe you need to also care that SSH is disabled after user chosen SSH?

fnkr:
> Maybe you need to also care that SSH is disabled after user chosen SSH?

Good hint, thanks! Fixed!

public/js/gogs.js
```diff
@@ -1035,4 +1037,18 @@ $(window).load(function () {
             }
         }).trigger('hashchange');
     }
+
+    // Repo clone url.
+    if ($('#repo-clone-url').length > 0) {
+        switch (localStorage.getItem('repo-clone-protocol')) {
+            case 'ssh':
+                if ($('#repo-clone-ssh').click().length === 0) {
```
Unknwon: `click()` seems redundant here?

fnkr: `.click()` triggers the onClick event and returns the jQuery object for method chaining. The code I wrote is equivalent to the following, but performs better as we don't have to evaulate the selector twice.

```js
if ($('#repo-clone-ssh').length === 0) {
    $('#repo-clone-https').click();
} else {
    $('#repo-clone-ssh').click();
}
```

Unknwon: Sounds very advanced!

Unknwon:
LGTM.

public/js/gogs.js
fnkr: It's quite simple. jQuery uses callbacks for everything, and returns the original jQuery object. This way you can write chained statements like this:

```js
// Add event listeners .click and .dblclick to #button1
$('#button1').click(function() { console.log('button1 click') })
             .dblclick(function() { console.log('button1 double-click') });
```



`#button1` is the selector, and `$('#button1')` the jQuery object, which is basically a special list with DOM-Elements and jQuery functions.

Calls like `$('#button1')` are quite heavy (performance) because of reasons, and as jQuery returns what we need anyway, I can just do the `.length` check afterwards. Benefit: the selector evaluation is done only once, and we don't have to create a new variable. Yay!

Unknwon: Though not fully understood, but thanks!

PR #2538, opened 2016-02-01
Make highlighted markdown code blocks work on issue pages

fnkr:
**Before**:
screenshot from 2016-02-01 20-55-50

**After**:
screenshot from 2016-02-01 20-55-38

PR #2542, opened 2016-02-01
Fix syntax highlighting for markdown code blocks on issue description/comment save

fnkr:
Looks like I missed something in #2538. Markdown code block syntax highlighting did still not work when updating an issue description or comment. Is fixed now.

**Before**:

**After**:


Unknwon:
Thanks!

PR #2546, opened 2016-02-02
Use pretty 404 pages in repo.HTTPBackend

fnkr:
**Before**:

**After**:

### DIFF
--- a/routers/repo/http.go
+++ b/routers/repo/http.go
@@ -215,7 +215,7 @@ func HTTP(ctx *middleware.Context) {
 		}
 	}
 
-	HTTPBackend(&Config{
+	HTTPBackend(ctx, &Config{
 		RepoRootPath: setting.RepoRootPath,
 		GitBinPath:   "git",
 		UploadPack:   true,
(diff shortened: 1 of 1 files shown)


routers/repo/http.go
```diff
@@ -433,8 +433,6 @@ func sendFile(contentType string, hr handler) {
 	w, r := hr.w, hr.r
 	reqFile := path.Join(hr.Dir, hr.File)
 
-	// fmt.Println("sendFile:", reqFile)
```
fnkr: Removed this comment too.

Unknwon:
Maybe `renderNotFound` could be removed too?

fnkr:
@Unknwon Actually, I tried to archive that but it's still used by `sendFile()` which is used by 6 functions that are used by [this struct](https://github.com/gogits/gogs/blob/v0.8.25/routers/repo/http.go#L251-L263) that [is used to add routes to the URL router](https://github.com/gogits/gogs/blob/v0.8.25/routers/repo/http.go#L291), that are used by the Git client (I believe), and I didn't wanted to return rendered HTML to Git.

Unknwon:
I'm confused by your comments and your changes?

>  that is used to add routes to the URL router, that are used by the Git client (I believe), and I didn't wanted to return rendered HTML to Git as Git probably prints it to the console.

Didn't you just changes those?

fnkr:
[`renderNotFound`](https://github.com/gogits/gogs/blob/v0.8.25/routers/repo/http.go#L527-L530) was used two functions: `HTTPBackend()` and `sendFile()`. This PR just changes `HTTPBackend()` to use `ctx.Handle()` instead of `renderNotFound()`. `sendFile()` still uses `renderNotFound()`.

Unknwon:
OK, cool!

PR #2548, opened 2016-02-02
Escape unescaped periods in route regular expression

fnkr:
Works: `/compare/sha1...sha1`
Works: `/compare/sha1FOOsha1` <-- make this don't work anymore

---

Side note:

Actually, I was trying to make [7-digit sha1 sums](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1) work, but I failed, because you can't make a Hex from an odd length string ([here](https://github.com/gogits/git-module/blob/3c8c495/sha1.go#L86)).

Another thing: I really wonder why gogs does not use [libgit2](https://libgit2.github.com) (like GitHub, BitBucket, etc.). Is that on purpose?

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -516,7 +516,7 @@ func runWeb(ctx *cli.Context) {
 			m.Get("/forks", repo.Forks)
 		}, middleware.RepoRef())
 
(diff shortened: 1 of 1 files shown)


Unknwon:
May consider after 1.0 release.

It uses CGO and very hard (not even know how) to cross-compile...

PR #2557, opened 2016-02-05
Fixed small typo in repository delete

Techwolf12:
Should be 'become' instead of 'became' as it still needs to happen.  
Thanks for building gogs for us! :smile:


Techwolf12:
Accidentally pushed on master and as @AngeloG says, be should be removed. I overlooked that.
New pull request: #2559

PR #2565, opened 2016-02-05
fixes #2563

### LINKED ISSUE #2563: DueDate field on creating/editing milestones modify date by -1 day
Issue found on https://try.gogs.io (v0.8.31.0205).

On the right side of the screen we can see two fields. Shorter one have label "Due Date (optional)" and it allow us to set end date for given milestone. There is a problem with field on saving changes. After save date is modified by 1 day backward. 

This issue can be connected somehow with #2564

tkausl:
fixes #2563 

There were actually _two_ bugs. First, it parsed the user-input as UTC and converted it to local-time when saving to the database, then, when loading it treated it (again) as UTC and converted it to local-time.

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -1179,6 +1179,7 @@ func (m *Milestone) AfterSet(colName string, _ xorm.Cell) {
 		if m.Deadline.Year() == 9999 {
 			return
 		}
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks!

PR #2567, opened 2016-02-06
Only show activities and repositories on the dashboard, that the user has access to

fnkr:

### DIFF
--- a/models/action.go
+++ b/models/action.go
@@ -593,12 +593,29 @@ func MergePullRequestAction(actUser *User, repo *Repository, pull *Issue) error 
 }
 
 // GetFeeds returns action list of given user in given context.
-func GetFeeds(uid, offset int64, isProfile bool) ([]*Action, error) {
+// ctxUserID is the user who's requesting, userID is the user/org that is requested.
+// ctxUserID can be -1, if isProfile is true or in order to skip the permission check.
+func GetFeeds(ctxUserID, userID, offset int64, isProfile bool) ([]*Action, error) {
 	actions := make([]*Action, 0, 20)
-	sess := x.Limit(20, int(offset)).Desc("id").Where("user_id=?", uid)
+	sess := x.Limit(20, int(offset)).Desc("id").Where("user_id=?", userID)
 	if isProfile {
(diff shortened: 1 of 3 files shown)


Eriner:
:+1: for fixing this issue. Thanks!

Unknwon:
Thanks for the PR!

But I think the core problem is why a user would receive feeds from a repository that s/he does not have access to?

fnkr:
What do you mean?

Old behaviour: show repo/activity, if the user has joined the org.
New behaviour: show repo/activity, if the user has access to it (= is in a team together with the repo).

It's exactly the same thing as in #2527 and #2530.

Unknwon:
No, it is not the same thing.

If you don't have access to a repo, how can you receive feeds from that?

fnkr:
In order to access a private organization repository, you need to be in a team that has (at least) read access to that repository. The problem: right now, this doesn't apply to feeds. You get feeds for all repositories in your orgs. It doesn't matter if you have access to that repo or not (read: if you are in a team that has access to that repo or not).

Unknwon:
> You get feeds for all repositories in your orgs.

No, only ones you watched I think.

fnkr:
That's not true. Example: _user2 is **only** in _team1. _team1 has **only** access to _repo1. BUT: right now, _user2 can view activities from _repo2 on the org dashboard. _user2 can't view _repo2. But he can see it's activities.

before

Unknwon:
That's why I said:

> But I think the core problem is why a user would receive feeds from a repository that s/he does not have access to?

fnkr:
What do you mean? The user get's all feeds that match the query?

Unknwon:
We need to stop user receiving feeds from repos that they do not have access to.

fnkr:
Yeah, this is that this PR does, isn't it?

Unknwon:
Sorry, I really mean `receive` when I say `get`.

Why should feeds for these users be ever generated if they don't have access?

fnkr:
The feeds you see when viewing an organization dashboard are stored under the user id of the organization, not your own user id.

Unknwon:
Ah, I see the problem now... Thank you! My bad.

PR #2570, opened 2016-02-06
Workaround delete folder on Windows.

### LINKED ISSUE #1738: Cannot delete read-only files on Windows
1. Create a repository with a name (let's say "myrepository")
2. Push something to "myrepository"
3. Delete "myrepository" on Gogs
4. Try to create another repository called "myrepository"

It returns 500. I think this happens because the repository folder is not deleted, just the database registry. I tried to rename the folder and could create the repository again.

andreynering:
Fix #1738

I tested and it is working properly. The folder is really deleted and I could create another repo with the same name.

### DIFF
--- a/models/admin.go
+++ b/models/admin.go
@@ -7,6 +7,7 @@ package models
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 	"time"
 
@@ -14,6 +15,7 @@ import (
 
 	"github.com/gogits/gogs/modules/base"
 	"github.com/gogits/gogs/modules/log"
+	"github.com/gogits/gogs/modules/setting"
 )
 
 type NoticeType int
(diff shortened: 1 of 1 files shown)


models/admin.go
```diff
@@ -53,13 +55,23 @@ func CreateRepositoryNotice(desc string) error {
...
 		desc := fmt.Sprintf("%s [%s]: %v", title, path, err)
 		log.Warn(desc)
 		if err = CreateRepositoryNotice(desc); err != nil {
 			log.Error(4, "CreateRepositoryNotice: %v", err)
 		}
 	}
+
+	if setting.IsWindows {
+		if err := exec.Command("cmd", "/C", "rmdir", "/S", "/Q", path).Run(); err != nil {
+			createNotice(err)
```
Unknwon: Maybe you can just call this only after `if-else` block?

andreynering: @Unknwon Done!

Unknwon:
Cool!

PR #2576, opened 2016-02-07
Add avatar arrows to both pull request and issue forms

chriswatt:
Examples:
gogscommentarrows

gogpullarrow

### DIFF
--- a/public/less/_repository.less
+++ b/public/less/_repository.less
@@ -268,9 +268,14 @@
 			}
 			.content {
 				margin-left: 4em;
+				#avatararrow;
+				&:after {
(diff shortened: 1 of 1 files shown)


public/less/_repository.less
```diff
@@ -268,6 +268,10 @@
 			}
 			.content {
 				margin-left: 4em;
+				#avatararrow;
```
Unknwon: Maybe rename it to `avatar-arrow`, would that work?

chriswatt: Renamed, added a new commit to fix.

Unknwon:
Thanks!

Issue #2571, opened 2016-02-06
Profile shows organisations if you are an admin


jamierocks:
Profiles only show organisations that your are an owner of.
![img](https://i.lexteam.xyz/zlBmMKOCng.png)
The profile should really show all organisations that your are part of and have set public visibility.

fnkr:
Is already fixed in develop branch.

See #2527.
Related: #2530, #2567.

jamierocks:
Oh - my bad.

fnkr:
@jamierocks Don't you want to close this issue? :-)

Unknwon:
I think @jamierocks 's point is admin could view all organization no matter public or private membership?

jamierocks:
My point was that the profile doesn't adhere to the visibility of that organisation ;)

fnkr:
So #2527 did what you wanted to be done?

jamierocks:
Yes.

fnkr:
So, why don't you close this issue? :D

jamierocks:
My bad - I thought it was closed after your original comment :P

PR #2578, opened 2016-02-07
Admins and user itself sees private org relations on profile

exmex:
This is inspired by @Unknwon 's post in #2571 
The only thing left to do is let org admins see the private repos as well. And maybe get a yellow outline or something on the avatar

### DIFF
--- a/templates/user/profile.tmpl
+++ b/templates/user/profile.tmpl
@@ -60,11 +60,7 @@
 							{{if .Orgs}}
 							<li>
 								{{range .Orgs}}
-									{{if $.Owner.IsPublicMember .Id}}
-										<a href="{{.HomeLink}}"><img class="ui mini image poping up" src="{{.AvatarLink}}" data-content="{{.Name}}" data-position="top center" data-variation="tiny inverted"></a>
-									{{else}}
-										<a href="{{.HomeLink}}"><img class="ui mini image poping up" src="{{.AvatarLink}}" data-content="{{.Name}} (Private)" data-position="top center" data-variation="tiny inverted"></a>
-									{{end}}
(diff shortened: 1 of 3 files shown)


Unknwon:
> The only thing left to do is let org admins see the private repos as well. And maybe get a yellow outline or something on the avatar

Org admin is not as powerful as site admin, so they shouldn't be able to see that I think.

templates/user/profile.tmpl
```diff
@@ -60,7 +60,11 @@
 							{{if .Orgs}}
 							<li>
 								{{range .Orgs}}
-									<a href="{{.HomeLink}}"><img class="ui mini image poping up" src="{{.AvatarLink}}" data-content="{{.Name}}" data-position="top center" data-variation="tiny inverted"></a>
+									{{if $.Owner.IsPublicMember .Id}}
+										<a href="{{.HomeLink}}"><img class="ui mini image poping up" src="{{.AvatarLink}}" data-content="{{.Name}}" data-position="top center" data-variation="tiny inverted"></a>
+									{{else}}
+										<a href="{{.HomeLink}}"><img class="ui mini image poping up" src="{{.AvatarLink}}" data-content="{{.Name}} (Private)" data-position="top center" data-variation="tiny inverted"></a>
```
Unknwon: I think it is not worth adding `(private)`..

exmex:
You're right, org admin is not as powerful. But still, Org owners / admins should see that the user is on the same team.

templates/user/profile.tmpl
exmex: Because own users can see all orgs too, I think it's better to mark it that it's private but is shown for admins & yourself. Or else somebody things that gogs is disclosing informations.

Unknwon: Good point but it does not matter, the email works the same way as I said. You can view yourself but non-login users can't.

Unknwon:
Thanks again!

PR #2579, opened 2016-02-07
Fix #2221 LDAP username attribute must be fetched

### LINKED ISSUE #2221: LDAP authentication does not work as expected
Gogs Version: 0.8.0.1212 with LDAP (via BindDN) configured with Microsoft Active Directory with the following filter: (&(objectClass=Person)(|(sAMAccountName=%[1]s)(mail=%[1]s))) 

If you log in for first time with user: user.name@example.com this will create the user user.name@example.com in gogs. However: 
   1. It is NOT possible to log in with user: user.name 
   2. It is possible to log in only with user: user.name@example.com

If you log in for first time with user: user.name this will create the user user.name in gogs. In this case everything is fine and it is possible to log in with both users: user.name and user.name@example.com

nanoant:
This is fix-up for 573305f. Forgot to fetch AttributeUsername value from the
LDAP server, so the setting was effectively not working as intended.


Unknwon:
Thanks!

PR #2581, opened 2016-02-07
Make commits tab active when on diff page

chriswatt:
Adds an or condition to the **Commits** tab so that the `active` class is enabled when viewing commit/diff pages


chriswatt:
I was debating whether or not to make the **Code** tab the active one as this is what github does.

Unknwon:
Thanks.

PR #2582, opened 2016-02-07
Fix tab index on new issue/comment form

### LINKED ISSUE #2482: tabbing from body of issue textarea and pressing space doesn't submit
1. create a new issue
2. tab from the issue title into the body textarea, type description
3. tab from body and press space (which is a button press when a button is highlighted) and issue is not submitted. instead the page just moves down.

chriswatt:
fixes #2482


Unknwon:
Thanks.

PR #2599, opened 2016-02-10
Fix for server error on release page when a user deleted their account

### LINKED ISSUE #2596: 500 error in releases after deleting an user
2016/02/10 11:27:09 [...ters/repo/release.go:50 Releases()] [E] GetUserById: user does not exist [uid: 2, name: ]

I finally recreated the user and updated the uid in the database.

mhartkorn:
Fixes #2596 using a fake user.


Unknwon:
Thanks!

PR #2612, opened 2016-02-12
update .gopmfile to reflect git-module update

Eriner:
Updated to reflect version increment in commit 47adc0e.


Unknwon:
Thanks!

PR #2617, opened 2016-02-13
Change main content area bg to white, keep area above tabs grey

chriswatt:
Dashboard, repo pages and organization have all have the been modified

See screenshot.

bg1

### DIFF
--- a/public/css/gogs.css
+++ b/public/css/gogs.css
@@ -6,7 +6,7 @@
 }
 body {
   font-family: 'Helvetica Neue', Arial, Helvetica, sans-serif, '微软雅黑';
-  background-color: #FAFAFA;
+  background-color: #fff;
   overflow-y: scroll;
 }
 img {
@@ -1124,6 +1124,11 @@ footer .container .links > *:first-child {
 .repository .metas .ui.list a .text:hover {
   color: #000;
 }
+.repository .header-wrapper {
+  background-color: #FAFAFA;
+  margin-top: -15px;
+  padding-top: 15px;
+}
 .repository .filter.menu .label.color {
   margin-left: 15px;
   padding: 0 8px;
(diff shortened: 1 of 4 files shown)


rajder:
Nice!

Unknwon:
Thanks your PR!

I figure it would be nice to post screenshots before changes as well for comparsion?

chriswatt:
Before and after comparisions

### Repo
gogsba1

### Org
gogsba2

### Dashboard
gogsba3

Unknwon:
Thanks.

What if the page does not have a header?

I think `ui subheader` might not be a good choice right now? Or maybe we could come up something that only change the `padding-top: 15px;` once without add any HTML?

chriswatt:
I can achieve the same effect without removing the `padding-top: 15px` at all

gogssubdiv

However because these 2 elements are on the same level, I still believe a they need to be wrapped in a new element in order to make the background differ

    <div class="ui container">
    <div class="ui tabs container">

Unknwon:
Good point, but these two elements only exists in repo-related views. To make a general solution, I still suggest do not wrapped inside another `div` zone.

Unknwon:
You can argue with me until one of us is convinced.

chriswatt:
Ok I'm back to argue. 

This is a new attempt using the method I mentioned above. I don't touch the padding-top-15px of the page itself, instead modify the new wrapper to do the same thing. 

This change only affects the repo page as it's the only one with tabs thus is the main to benefit from it.

For a more general solution that would affect other pages with a header, I think the each page would need to be restructured in a similar fashion for this to work.

public/css/gogs.css
```diff
@@ -2096,6 +2173,27 @@ footer .container .links > *:first-child {
   width: 15px;
   height: 15px;
 }
+#avatar-arrow:before,
```
Unknwon: What's going on with these lines?

Unknwon:
Can you do `git reset --soft HEAD^7` and create a clean commit again.

Then use `git push --force` to your branch?

Changes loos very strange right now.

public/css/gogs.css
chriswatt: Those are from my previous pr for the arrows on the comment form, gogs.css wasn't generated after you merged it

chriswatt:
Yeah looking at the diff my branch does look a bit weird. I'll try fix

public/css/gogs.css
Unknwon: I see.

Unknwon:
Much better now!

Unknwon:
Can you provide screenshots for an empty repo?

chriswatt:
Sure ( ͡° ͜ʖ ͡°)

gogsemptyrepo

Unknwon:
Thanks!

PR #2624, opened 2016-02-14
Convert mirrors to regular repositories

### LINKED ISSUE #2607: Add ability to alter an existing repository from "mirroring" to "normal" behavior
IMHO it would be nice if one could create a repository mirror and decide later, if that this repository should be standalone from now on. I guess that could be handy to have if one is doing preparations for moving towards gogs. It would be possible to keep the repositories in sync, till everything is in place.

mhartkorn:
This PR intends to fix  #2607.

I refactored `models.MigrateRepository` a bit to extract the steps that are taken if a repository is *not* a mirror into a separate method. With this I was able to create a new option in the repository settings that converts a mirror into a regular repository.

There might be problems with the new translation strings because I was not able to make them work on my local machine. Maybe someone can help me out here?


mhartkorn:
OK, I got the translations to work locally but I don't want to push anything because the generated files looks very odd to me.

models/repo.go
```diff
@@ -654,7 +654,17 @@ func MigrateRepository(u *User, opts MigrateRepoOptions) (*Repository, error) {
...
+	repo, err = FinishMigrateRepository(repo, repoPath)
+	if err != nil {
+		return repo, err
+	}
+
+	return repo, UpdateRepository(repo, false)
+}
+
+// Finish migrating repository with things that don't need to be done for mirrors.
+func FinishMigrateRepository(repo *Repository, repoPath string) (*Repository, error) {
```
Unknwon: Better call it `CleanUpMirrorInfo`?

conf/locale/locale_en-US.ini
```diff
@@ -587,11 +587,16 @@ settings.danger_zone = Danger Zone
...
 settings.transfer_desc = Transfer this repository to another user or to an organization in which you have admin rights.
 settings.new_owner_has_same_repo = The new owner already has a repository with same name. Please choose another name.
+settings.convert = Convert To Regular Repository
+settings.convert_desc = You can convert this mirror to a regular repository. This cannot be reversed.
 settings.delete = Delete This Repository
 settings.delete_desc = Once you delete a repository, there is no going back. Please be certain.
 settings.transfer_notices_1 = - You will lose access if new owner is a individual user.
 settings.transfer_notices_2 = - You will conserve access if new owner is an organization and if you're one of the owners.
 settings.transfer_form_title = Please enter following information to confirm your operation:
+settings.convert_notices_1 = - This operation will convert this repository mirror into a regular repository and cannot be undone.
```
Unknwon: Why not put these lines right after line 591?

Unknwon:
Can you take a screenshot of new settings page?

models/repo.go
mhartkorn: I moved the existing code into a separate method and did not change anything (except for adding an additional return). The code I moved was not executed when creating a mirror but when creating a regular repo.

I thought this would be better than writing code that does essentially the same as the existing code.

mhartkorn:
New settings "Danger Zone":

gogs

models/repo.go
Unknwon: No... I'm talking the function name..

conf/locale/locale_en-US.ini
mhartkorn: I refactored the file a bit and moved all "transfered", "delete", and "convert" strings together. They were all scattered all over the place. Will be pushed with the next commit.

models/repo.go
mhartkorn: Oh sorry. I think `CleanUpRepoInfo` would be better. It is not only used for mirrors.

mhartkorn: Or `CleanUpMigrateInfo`.

Unknwon: OK, `CleanUpMigrateInfo` looks better.

modules/auth/repo_form.go
```diff
@@ -37,6 +37,18 @@ func (f *CreateRepoForm) Validate(ctx *macaron.Context, errs binding.Errors) bin
 	return validate(errs, ctx.Data, f, ctx.Locale)
 }
 
+type ConvertRepoForm struct {
```
Unknwon: Why do you need this?

routers/repo/setting.go
```diff
@@ -141,6 +141,39 @@ func SettingsPost(ctx *middleware.Context, form auth.RepoSettingForm) {
...
+				ctx.Error(404)
+				return
+			}
+		}
+
+		if !repo.IsMirror {
+			ctx.Error(404)
+			return
+		}
+
```
Unknwon: Please remove this line.

routers/repo/setting.go
```diff
@@ -141,6 +141,39 @@ func SettingsPost(ctx *middleware.Context, form auth.RepoSettingForm) {
...
+		repo.IsMirror = false
+
+		if _, err := models.FinishMigrateRepository(repo, models.RepoPath(ctx.Repo.Owner.Name, repo.Name)); err != nil {
+			ctx.RenderWithErr(ctx.Tr("settings.convert.failed"), SETTINGS_OPTIONS, &form)
+			return
+		}
+
+		if 	err := models.UpdateRepository(repo, false); err != nil {
```
Unknwon: Looks like didn't run `gofmt`?

routers/repo/setting.go
```diff
@@ -141,6 +141,39 @@ func SettingsPost(ctx *middleware.Context, form auth.RepoSettingForm) {
...
+			ctx.RenderWithErr(ctx.Tr("settings.convert.failed"), SETTINGS_OPTIONS, &form)
+			return
+		}
+		log.Trace("Repository converted: %s/%s", ctx.Repo.Owner.Name, repo.Name)
```
Unknwon: `Repository converted from mirror to regular`

templates/repo/settings/options.tmpl
```diff
@@ -165,6 +178,36 @@
 	</div>
 </div>
 
+<div class="ui small modal" id="convert-repo-modal">
```
Unknwon: Use this inside the logic block `{{if .Repository.IsMirror}}` as well.

mhartkorn:
Thanks for the good suggestions. I refactored the code.

models/repo.go
```diff
@@ -695,7 +705,7 @@ func MigrateRepository(u *User, opts MigrateRepoOptions) (*Repository, error) {
 		repo.DefaultBranch = headBranch.Name
 	}
 
-	return repo, UpdateRepository(repo, false)
+	return repo, nil
```
Unknwon: Should we call `UpdateRepository` here and looks more consist because normally you wouldn't expect an update operation after such function?

mhartkorn:
The new commit moved `UpdateRepository` into `CleanUpMigrateInfo`. I forgot to remove the repository from the `mirror` table in the database. This is fixed now with the new method `models.DeleteMirrorByRepoID`.

Unknwon:
Thanks!

Issue #2627, opened 2016-02-15
username in HTTPS URL


mrexodia:
When you create a new repo it says:

```
touch README.md
git init
git add README.md
git commit -m "first commit"
git remote add origin https://git.myserver.com:443/user/repo.git
git push -u origin master
```

Tt would be nice if the `user` was in the https clone url like so:

```
touch README.md
git init
git add README.md
git commit -m "first commit"
git remote add origin https://user@git.myserver.com:443/user/repo.git
git push -u origin master
```

This saves time typing the user/password when cloning.

xxxtonixxx:
Yeah, that would be a good feature when you are logged :+1:

mrexodia:
I looked at how to implement it but I never programmed it Go so I'm now
sure I can do it. There should also be checks on if you have access to the
repository and stuff.

Unknwon:
Note for who wants to help:

1. This information will only be shown for who has write access to the empty repository.
2. We can show the user name by using template variable `{{.SignedUserName}}` to whoever signed in and view this page.

sapk:
I can do it

Unknwon:
Thanks @sapk

sapk:
Just to do the right thing, can we apply to all clone (also wiki) url or just for the init of the repo ?

Unknwon:
> just for the init of the repo

I think that's enough.

xxxtonixxx:
why?, I think that it would be well for everything... are there differences between a repo doesn't initialized and a repo already initialized?

Unknwon:
> why?, I think that it would be well for everything...

Init page is more meant to be inited and push, URLs in other pages are not.

xxxtonixxx:
Ok, make sense! :+1:

mrexodia:
I disagree. It should be also in the clone Url (like bitbucket). Including
it there saves typing a username which is never a bad thing in my opinion.

Unknwon:
> it there saves typing a username which is never a bad thing in my opinion.

I do not think this saving is necessary.

mrexodia:
Well in that case add an option to enable/disable adding a username to the
HTTPS url. Bitbucket has been doing this for years now (for clone URLs) and
I never had any trouble with it.

Now that I think of it, I did see some option to customize the HTTPs URL.
Is it possible to include this in there?

Unknwon:
> Now that I think of it, I did see some option to customize the HTTPs URL.
Is it possible to include this in there?

[...]

PR #2631, opened 2016-02-15
Username in clone HTTP(S) URL on init repo page

sapk:
Implement #2627

captureecran2627

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -453,24 +453,24 @@ func (repo *Repository) cloneLink(isWiki bool, signedUserName string) *CloneLink
 		cl.SSH = fmt.Sprintf("%s@%s:%s/%s.git", setting.RunUser, setting.SSHDomain, repo.Owner.Name, repoName)
 	}
 
-	if(len(signedUserName) > 0){
-		cl.HTTPS = fmt.Sprintf("%s@%s:%s/%s/%s/%s.git", signedUserName, setting.Domain, setting.HttpPort, setting.AppSubUrl ,repo.Owner.Name, repoName)
-	}else{
+	if len(signedUserName) > 0 {
+		cl.HTTPS = fmt.Sprintf("%s@%s:%s/%s/%s/%s.git", signedUserName, setting.Domain, setting.HttpPort, setting.AppSubUrl, repo.Owner.Name, repoName)
+	} else {
 		cl.HTTPS = fmt.Sprintf("%s:%s/%s/%s/%s.git", setting.Domain, setting.HttpPort, setting.AppSubUrl, repo.Owner.Name, repoName)
 	}
-	cl.HTTPS = fmt.Sprintf("%s://%s", setting.Protocol, path.Clean(cl.HTTPS));
-	
+	cl.HTTPS = fmt.Sprintf("%s://%s", setting.Protocol, path.Clean(cl.HTTPS))
+
 	return cl
 }
 
 // CloneLink returns clone URLs of repository with the signed username in it.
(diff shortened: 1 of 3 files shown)


sapk:
Maybe a another improvement could be to use net/url to represent HTTPS ? In order to be able to edit the username more easely.

modules/middleware/repo.go
```diff
@@ -186,6 +186,8 @@ func RepoAssignment(args ...bool) macaron.Handler {
 			if displayBare {
 				if !ctx.Repo.IsAdmin() {
 					ctx.Flash.Info(ctx.Tr("repo.repo_is_empty"), true)
+				} else { //Has write access and repo is empty
```
Unknwon: You can put comment in the next line and add a space after `//`.

models/repo.go
```diff
@@ -452,13 +452,25 @@ func (repo *Repository) cloneLink(isWiki bool) *CloneLink {
...
+		cl.HTTPS = fmt.Sprintf("%s@%s:%s/%s/%s/%s.git", signedUserName, setting.Domain, setting.HttpPort, setting.AppSubUrl, repo.Owner.Name, repoName)
+	} else {
+		cl.HTTPS = fmt.Sprintf("%s:%s/%s/%s/%s.git", setting.Domain, setting.HttpPort, setting.AppSubUrl, repo.Owner.Name, repoName)
+	}
+	cl.HTTPS = fmt.Sprintf("%s://%s", setting.Protocol, path.Clean(cl.HTTPS))
+
 	return cl
 }
 
+// CloneLink returns clone URLs of repository with the signed username in it.
```
Unknwon: Wrong start of comment.

models/repo.go
```diff
@@ -439,7 +439,7 @@ type CloneLink struct {
 	Git   string
 }
 
-func (repo *Repository) cloneLink(isWiki bool) *CloneLink {
+func (repo *Repository) cloneLink(isWiki bool, signedUserName string) *CloneLink {
```
Unknwon: Call `userName` is good enough, `signedUserName` is narrowing the actual usage of this method.

models/repo.go
```diff
@@ -452,13 +452,25 @@ func (repo *Repository) cloneLink(isWiki bool) *CloneLink {
 	} else {
 		cl.SSH = fmt.Sprintf("%s@%s:%s/%s.git", setting.RunUser, setting.SSHDomain, repo.Owner.Name, repoName)
 	}
-	cl.HTTPS = fmt.Sprintf("%s%s/%s.git", setting.AppUrl, repo.Owner.Name, repoName)
+
+	if len(signedUserName) > 0 {
```
Unknwon: You can compose URL path without `userName` and then add it in the front.

fnkr:
Why is this necessary? It makes clone links noninterchangeable, and adds functionality to URLs that has nothing to do with the sense of an URL. If you want to store your access credentials use  [git credential store](https://git-scm.com/docs/git-credential-sto re) or SSH with public keys as these are are designed to store your authentication data.

zacheryph:
I would agree with @fnkr.  The username has no purpose being in the URL.  You don't put it there for ssh URLs.  The only point to this is to save a few keystrokes when not using `git-credential-store`.

Unknwon:
@fnkr thanks for the comments. This PR only changes the init page.

fnkr:
@Unknwon This PR changes the repo page, which is where you get the git url
from.

Unknwon:
> This PR changes the repo page

How?

Unknwon:
This is not a good idea in a second thought, closing...

Issue #2628, opened 2016-02-15
LDAP Search failed unexpectedly


kop:
Hi,

I'm trying to configure Gogs to use LDAP service from [JumpCloud](https://jumpcloud.com/) and having the following error during the auth:

```
2016/02/15 18:19:18 LDAP will use BindDN.
2016/02/15 18:19:18 Search for LDAP user: <username>
2016/02/15 18:19:18 Bound as BindDN uid=<bind username>,ou=Users,o=<organization ID>,dc=jumpcloud,dc=com
2016/02/15 18:19:18 Searching using filter (&(objectClass=inetOrgPerson)(uid=<username>))
2016/02/15 18:19:18 Binding with userDN: uid=<username>,ou=Users,o=55fcb6467ff1ab131d018642,dc=jumpcloud,dc=com
2016/02/15 18:19:18 Bound successfully with userDN: uid=<username>,ou=Users,o=<organization ID>,dc=jumpcloud,dc=com
2016/02/15 18:19:18 ...gogs/models/login.go:246 LoginUserLDAPSource()] [E] LDAP Search failed unexpectedly! (0 entries)
```

Can't really understand what does this mean.

And here is the config I'm using:

![gogs-auth](https://cloud.githubusercontent.com/assets/645148/13059605/058522ea-d434-11e5-96df-3e4710efb219.png)

Unknwon:
Hi, I think it means nothing return/found by the information you provide.

Unknwon:
cc @SergioBenitez @nanoant

kop:
@Unknwon, I thought about this too, however it looks strange because it says that it's binded to userDN. No user data found (like first name, last name, etc)? But it should be there and I'm using all field names provided by JumpCloud support.

Unknwon:
I think you would have to take a screenshot of your config first..

kop:
Sure, config is added.

Unknwon:
Thanks! I don't know LDAP very well, we'll wait who knows better come to help. :sweat:

nanoant:
@Unknwon this is something else, because user lookup using BindDN succeeds, but then we Gogs tries to bind the user and using his/her password then it queries again for the user attributes in `modules/auth/ldap/ldap.go:154`. For some reason this query does not return anything.

@kop It looks like your server does not return anything on user asking for his/her own attributes, which is weird. Thoughts!?

Unknwon:
@nanoant ah, good to learn!

[...]

PR #2634, opened 2016-02-16
LDAP: Fetch attributes in Bind DN context

nanoant:
This is follow up for #2628 introducing option to fetch attributes in Bind DN context and making Gogs compatible with JumpCloud security pattern.

I have also added extra improvement to use single connection for BindDN mode, instead 2 connection. Since LDAP proto allows sending 2 bind requests within same connection.


modules/auth/ldap/ldap.go
```diff
@@ -124,33 +125,28 @@ func (ls *Source) SearchEntry(name, passwd string, directBind bool) (string, str
...
-		return "", "", "", "", false, false
-	}
-	defer l.Close()
-
-	log.Trace("Binding with userDN: %s", userDN)
-	err = l.Bind(userDN, passwd)
-	if err != nil {
-		log.Debug("LDAP auth. failed for %s, reason: %v", userDN, err)
-		return "", "", "", "", false, false
+	if directBind || !ls.AttributesInBind {
```
Unknwon: Looks like exactly same `if` block occurs twice?

nanoant: @Unknwon No its not. These 2 block have opposite condition `if directBind || !ls.AttributesInBind` vs `if !directBind && ls.AttributesInBind`. I could move out their body to some `bindUser` helper proc, but this wouldn't reduce code much.

Unknwon: I suggest you use a helper function `bindUser`, and I do not quite understand this part why they have opposite conditions?

nanoant: Because `bind` changes the authentication context in LDAP connection, if we bind user before looking up & fetching attributes then this lookup is done in user context, if we bind user after, then the fetch is done in BindDN context. Which is anyway the subject of the commit.

Anyway I'll try to improve readability of this code part and I'll update the PR.

nanoant:
@Unknwon I have factored out `bindUser` helper func and added some 2 extra comments. Lemme know if now it looks more clear.

Unknwon:
Thanks, just to confirm, these changes do not break compatibilities of LDAP at least on your workflow?

nanoant:
@Unknwon Yes, it does not break my workflow. Moreover the default `AttributesInBind` is false, which is the current Gogs behavior. So it won't affect current users.
If you want to test it yourself here's some simple LDAP server written in Go for testing purposes: https://github.com/nanoant/ldaptestserver

Issue #2179, opened 2015-12-13
Cannot add SSH key when running via systemd


david-at:
OS: CentOS 7. I have a systemd script setup to start gogs, when I am running via that systemd script I cannot add ssh keys and get the following error: 

Sorry, we're not able to verify your SSH key: ssh-keygen -lf: ssh-keygen: /tmp/keytest422813223: Permission denied

If I run the service by hand it works fine: sudo ./gogs web

Here is my systemd service: 

```
[Unit]
Description=Gogs (Go Git Service)
After=syslog.target
After=network.target
#After=mysqld.service
#After=postgresql.service
#After=memcached.service
#After=redis.service

[Service]
Type=simple
User=root
Group=root
WorkingDirectory=/home/gogs/gogs
ExecStart=/home/gogs/gogs/gogs web
Restart=always
Environment=USER=root HOME=/home/gogs

[Install]
WantedBy=multi-user.target
```

Originally I had it setup to use the gogs user,  but for testing I switched it to root. Still no luck.

Unknwon:
This is the sysmted config for demo site:

```
[Unit]
Description=Gogs - try.gogs.io
After=syslog.target
After=network.target
After=mysqld.service

[Service]
User=git
Group=git
WorkingDirectory=/home/git/apps/gogs
ExecStart=/home/git/apps/gogs/gogs web
Restart=always
Environment="USER=git","HOME=/home/git"

[Install]
WantedBy=multi-user.target
```

Compare two versions, I think the problem is `Type=simple` even though I don't know what it does.

xor-gate:
`Type=simple` means the process stays in the foreground and wont fork. The default when omitted is `Type=simple` which is good in both of you.

http://www.freedesktop.org/software/systemd/man/systemd.service.html#Options

I think the problem is here:

```
Environment=USER=root HOME=/home/gogs
vs
Environment="USER=git","HOME=/home/git"
````

@david-at see if this works.

**Comment by Unknwon** [comment]
@xor-gate thanks your help!

**Comment by xor-gate** [comment]
Have a deep look how `Environment=` is expanded: http://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines

**Comment by Unknwon** [comment]
Close for lacking of feedback.

**Comment by david-at** [comment]
Sorry for just now getting to this.

I updated the conf with the env settings, and still no luck: 

```
[Unit]
Description=Gogs (Go Git Service)
After=syslog.target
After=network.target
[...]

PR #2637, opened 2016-02-16
allow native and ssh-keygen public key check

Gibheer:
Hi,

this is the first part of fixing #2179, namely the SELinux and ssh-keygen problem. The first commit is pretty large, but adding public keys now works again using ssh-keygen.

Please do not merge yet, as there is still some stuff missing. I introduced new settings to make it possible to adjust the working directory for ssh-keygen and even adjust the ssh-keygen path. I have to update all occurrences of ssh-keygen for that still.
But if possible, I welcome everyone to test the patch :)


Unknwon:
PS: build is failing...

Gibheer:
Now this error I have no idea how to fix that. I tried with openssh 5.9 and that returns "is not a public key file" for ed25519 public keys. But what is travis using in this case, that it returns "file too long"?

I would like to keep the tests, but if it breaks the build, it is kind of stupid. Should I delete them or should I just remove the ed25519 key from the test?

Unknwon:
I guess we need special instructions to tell Travis (but maybe not possible?) support ed25519 key.

Is this PR finished now?

Gibheer:
Not yet. I want to adjust all occurences of ssh-keygen to use the new variables. But if you want, you could pull this request in and the other stuff will be done extra.

models/ssh_key.go
```diff
@@ -33,7 +35,10 @@ const (
 	_TPL_PUBLICK_KEY = `command="%s serv key-%d --config='%s'",no-port-forwarding,no-X11-forwarding,no-agent-forwarding,no-pty %s` + "\n"
 )
 
-var sshOpLocker = sync.Mutex{}
+var (
+	sshOpLocker          = sync.Mutex{}
+	SSH_UNKNOWN_KEY_TYPE = fmt.Errorf("unknown key type")
```
Unknwon: Variable should not use ALL_CAPS.

Unknwon:
What is the different between this PR and revert old approach back?

Gibheer:
This is a combination of the old behavior and the new one. At the end it should be possible to use the go native ssh library only without any dependency on ssh-keygen at all. Currently, there are still some places in need of ssh-keygen.
That would also mean, that depending on what server you choose to use (sshd, gogs), you would get the full support from the specific system.

modules/setting/setting.go
```diff
@@ -30,9 +30,11 @@ import (
 type Scheme string
 
 const (
-	HTTP  Scheme = "http"
-	HTTPS Scheme = "https"
-	FCGI  Scheme = "fcgi"
+	HTTP                       Scheme = "http"
+	HTTPS                      Scheme = "https"
+	FCGI                       Scheme = "fcgi"
+	SSH_PUBLICKEY_CHECK_NATIVE        = "native"
```
Unknwon: This `const` section is for `Scheme`, do not mixed up with them.

modules/setting/setting.go
```diff
@@ -328,6 +333,29 @@ func NewContext() {
 	if err := os.MkdirAll(SSHRootPath, 0700); err != nil {
 		log.Fatal(4, "Fail to create '%s': %v", SSHRootPath, err)
 	}
+	checkDefault := SSH_PUBLICKEY_CHECK_KEYGEN
+	if DisableSSH {
+		checkDefault = SSH_PUBLICKEY_CHECK_NATIVE
+	}
```
Unknwon: When disable SSH, why should we care the method at all?

Unknwon:
ping @Gibheer

Gibheer:
@Unknwon sorry, took a while longer. I fixed the stuff you found and replaced the ssh-keygen call with the new variable from setting. You can merge this for now, as I do not know when I have the time to do the rest (making ssh-keygen only necessary when the system sshd is used).
Thank you for your help :)

Unknwon:
Thanks!

PR #2647, opened 2016-02-17
Implement issue and pull request templates

andreynering:
GitHub just announced [issue and pull request templates](https://github.com/blog/2111-issue-and-pull-request-templates).

This PR implements it in Gogs and also adds templates to be used on GitHub.

### DIFF
--- a/routers/repo/issue.go
+++ b/routers/repo/issue.go
@@ -7,6 +7,8 @@ package repo
 import (
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
@@ -34,11 +36,19 @@ const (
 	MILESTONE      base.TplName = "repo/issue/milestones"
 	MILESTONE_NEW  base.TplName = "repo/issue/milestone_new"
 	MILESTONE_EDIT base.TplName = "repo/issue/milestone_edit"
+
+	ISSUE_TEMPLATE_KEY = "IssueTemplate"
 )
 
 var (
 	ErrFileTypeForbidden = errors.New("File type is not allowed")
 	ErrTooManyFiles      = errors.New("Maximum number of files to upload exceeded")
+
+	IssueTemplateCandidates = []string{
+		"ISSUE_TEMPLATE.md",
+		".gogs/ISSUE_TEMPLATE.md",
+		".github/ISSUE_TEMPLATE.md",
+	}
 )
 
 func MustEnableIssues(ctx *middleware.Context) {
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks, but I suggest split this into two PRs.

andreynering:
@Unknwon  Ok, done.

andreynering:
@Unknown We can open the possibility of permitting templates. inside `.gogs` folder (and maybe also `.github` folder, for compatibility). What do you think?

Unknwon:
> possibility of permitting templates. inside .gogs folder (and maybe also .github folder, for compatibility). What do you think?

Yes, that's what I'm thinking, too

Unknwon:
The priority would be `root -> .gogs -> .github`

Eriner:
> The priority would be `root -> .gogs -> .github`

I really like this; will help keep folders clean. Nice to include `.github` as well for compatibility reasons. :+1:

andreynering:
@Unknwon  Updated. It is working.

There's just a little detail, but I don't thing it is an issue: when opening a PR, the template will be taken from the user repo, not the base one. So it won't show if the user didn't update his repo. Since this is often made before starting working on something, I don't think it will be a problem.

routers/repo/pull.go
```diff
@@ -540,6 +540,9 @@ func CompareAndPullRequest(ctx *middleware.Context) {
 	ctx.Data["PageIsComparePull"] = true
 	ctx.Data["IsDiffCompare"] = true
 	ctx.Data["RequireHighlightJS"] = true
+	setTemplateIfExists(ctx, ".github/PULL_REQUEST_TEMPLATE.md", "PullRequestTemplate")
```
Unknwon: Why would you call 3 times in the handler? Shouldn't we make a function that make 3 attempts and stop once found?

andreynering: You are right. I did bit of refactoring. Updated.

routers/repo/issue.go
```diff
@@ -281,9 +283,61 @@ func RetrieveRepoMetas(ctx *middleware.Context, repo *models.Repository) []*mode
...
+	for _, filename := range possibleFiles {
+		content, found := getFileContentFromDefaultBranch(ctx, filename)
+		if found {
+			ctx.Data["IssueTemplate"] = content
+			return
+		}
+	}
+}
+
+func setPullRequestTemplateIfExists(ctx *middleware.Context) {
```
Unknwon: `setIssueTemplateIfExists` and `setPullRequestTemplateIfExists` can be abstracted again.

andreynering: Done.

routers/repo/issue.go
```diff
@@ -34,11 +36,19 @@ const (
...
 	MILESTONE_EDIT base.TplName = "repo/issue/milestone_edit"
+
+	ISSUE_TEMPLATE_KEY = "IssueTemplate"
 )
 
 var (
 	ErrFileTypeForbidden = errors.New("File type is not allowed")
 	ErrTooManyFiles      = errors.New("Maximum number of files to upload exceeded")
+
+	ISSUE_POSSIBLE_TEMPLATE_FILES = []string{
```
Unknwon: ALL_CAPS is not the coding style for variables...

routers/repo/pull.go
```diff
@@ -26,6 +26,16 @@ const (
 	COMPARE_PULL base.TplName = "repo/pulls/compare"
 	PULL_COMMITS base.TplName = "repo/pulls/commits"
 	PULL_FILES   base.TplName = "repo/pulls/files"
+
+	PULL_REQUEST_TEMPLATE_KEY = "PullRequestTemplate"
+)
+
+var (
+	PULL_REQUEST_POSSIBLE_TEMPLATE_FILES = []string{
```
Unknwon: Save as above. `PullReuqestTemplateCandidates` might be better?

routers/repo/issue.go
andreynering: Ops. Fixed.

Unknwon:
Thanks!

PR #2658, opened 2016-02-18
Fix chmod for several files in conf/locale/ and public/

fnkr:
Non-executable should not be executable.


Unknwon:
..... holy (bi)

Issue #2505, opened 2016-01-28
Fork a mirror feature


maxchill:
Since the mirrors are readonly, it would be handy to still have the ability to fork, even if you can't push it back to origin.

Unknwon:
Thanks your feedback!

But it may have very low priority...

Unknwon:
This feature is claimed to be implemented by merging https://github.com/gogits/gogs/pull/2663

Unknwon:
Close as implemented.

PR #2663, opened 2016-02-19
#2505 Allow to fork and disallow to create PRs for mirrors.

Download-Fritz:
Allow to fork and disallow to create PRs for mirrors, which addresses issue #2505.

Changes have been made to handle cases of prior created mirror repositories to provide backwards-compatibility.

Pull Requests were disabled, because...
  1) Mirrors are supposed to be read-only (merging a pull request would write to the it).
  2) prior to this change, no forks could have been made from within the GUI, making Pull Request support redundant.

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -332,7 +332,17 @@ func (repo *Repository) IsOwnedBy(userID int64) bool {
 
 // CanBeForked returns true if repository meets the requirements of being forked.
 func (repo *Repository) CanBeForked() bool {
-	return !repo.IsBare && !repo.IsMirror
+	return !repo.IsBare
+}
+
+// CanEnablePulls returns true if repository meets the requirements of accepting pulls.
+func (repo *Repository) CanEnablePulls() bool {
+	return !repo.IsMirror
+}
+
+// AllowPulls returns true if repository meets the requirements of accepting pulls and has them enabled.
+func (repo *Repository) AllowsPulls() bool {
+	return repo.CanEnablePulls() && repo.EnablePulls;
 }
 
 func (repo *Repository) NextIssueIndex() int64 {
--- a/routers/repo/issue.go
+++ b/routers/repo/issue.go
@@ -48,7 +48,9 @@ func MustEnableIssues(ctx *middleware.Context) {
 }
 
 func MustEnablePulls(ctx *middleware.Context) {
-	if !ctx.Repo.Repository.EnablePulls {
+	if !ctx.Repo.Repository.CanEnablePulls() {
+		ctx.Handle(404, "Unsupported", nil)
+	} else if !ctx.Repo.Repository.EnablePulls {
 		ctx.Handle(404, "MustEnablePulls", nil)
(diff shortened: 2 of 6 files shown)


models/repo.go
```diff
@@ -931,6 +941,10 @@ func CreateRepository(u *User, opts CreateRepoOptions) (_ *Repository, err error
 		EnablePulls:  true,
 	}
 
+	if !repo.CanEnablePulls() {
+		repo.EnablePulls = false
```
Unknwon: Isn't this redundant?

models/repo.go
```diff
@@ -332,7 +332,12 @@ func (repo *Repository) IsOwnedBy(userID int64) bool {
 
 // CanBeForked returns true if repository meets the requirements of being forked.
 func (repo *Repository) CanBeForked() bool {
-	return !repo.IsBare && !repo.IsMirror
+	return !repo.IsBare
+}
+
+// CanEnablePulls returns true if repository meets the requirements of enabling pulls.
+func (repo *Repository) CanEnablePulls() bool {
```
Unknwon: Maybe rename to `AllowPulls`?

Download-Fritz:
@Unknwon Thank you for your feedback!

I left the CanEnablePulls() function in for the case the conditions need to be adapted in the future.
Though I implemented AllowsPulls() and got rid of the 'backwards-compatibility' code.

It indeed looks cleaner now. Thanks for yout feedback. ;)

routers/repo/issue.go
```diff
@@ -48,7 +48,9 @@ func MustEnableIssues(ctx *middleware.Context) {
 }
 
 func MustEnablePulls(ctx *middleware.Context) {
-	if !ctx.Repo.Repository.EnablePulls {
+	if !ctx.Repo.Repository.CanEnablePulls() {
```
Unknwon: Based on assumption that when `EnablePulls` is `true`, `CanEnablePulls` must be `true` as well, then this `if` check is redundant.

Download-Fritz: @Unknwon That is not the case, though. Present mirror repositories will have EnablePulls set to true (as it is the default value), while mirrors should not support Pull Requests as of this commit.

routers/repo/issue.go
```diff
@@ -48,7 +48,9 @@ func MustEnableIssues(ctx *middleware.Context) {
 }
 
 func MustEnablePulls(ctx *middleware.Context) {
```
Unknwon: Then, this function should be renamed to `MustAllowPulls`.

routers/repo/issue.go
Unknwon: See my edited comments.

routers/repo/issue.go
Download-Fritz: @Unknwon You take clean code very seriously. :)
Fixed in a second commit. Thanks for the feedback.

Unknwon:
Thanks!

PR #2664, opened 2016-02-19
Update contributing guidlines link

jwdeitch:
contributing link in README.md previously displays 404 error


Unknwon:
Thanks, but the correct link should be https://github.com/gogits/gogs/blob/master/.github/CONTRIBUTING.md

jwdeitch:
updated :)

Unknwon:
Thanks!

PR #2665, opened 2016-02-20
Dockerfile cleanup for Alpine 3.3

muhmuhten:
Eliminates some package-manager wrangling which has become unnecessary since the Dockerfile was written (~half a year ago, targetting alpine 3.2). In particular, by eliminating the edge pin, we're a bit more insulated from possible breaking changes in edge (package upgrades in stable are not supposed to be breaking).

- s6 is in main in 3.3, so we no longer need to mangle the repos file
- official image is periodically updated, so it's not preferred to do
  upgrades downstream (usually harmless, but inelegant)
- apk-tools in 3.3 supports --no-cache to avoid leaving the APKINDEX
  files in the image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,8 @@ MAINTAINER jp@roemer.im
 
 # Install system utils & Gogs runtime dependencies
 ADD https://github.com/tianon/gosu/releases/download/1.6/gosu-amd64 /usr/sbin/gosu
-RUN echo "@edge http://dl-4.alpinelinux.org/alpine/edge/main" | tee -a /etc/apk/repositories \
- && apk -U --no-progress upgrade \
- && apk -U --no-progress add ca-certificates bash git linux-pam s6@edge curl openssh socat \
- && chmod +x /usr/sbin/gosu
+RUN chmod +x /usr/sbin/gosu \
+ && apk --no-cache --no-progress add ca-certificates bash git linux-pam s6 curl openssh socat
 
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR!

cc @0rax

0rax:
Hi, I did not check if s6 was push to main in 3.3 and was not aware of this new feature of apk-tools, thanks for your PR and those different infos !

This can be merged with no problem at all (just tested the change). I will just have to add this `--no-cache` argument in `docker/build.sh`, as `APKINDEX` will still be created while building the software.

Unknwon:
Thanks!

@muhmuhten would you mind update [this file](https://github.com/gogits/gogs/blob/master/Dockerfile.rpi) as well?

muhmuhten:
@Unknwon It doesn't look like I can, sorry. The image it's based off hasn't had an update for 3.3, and while there are other ARM alpine bases (e.g. container4armhf/armhf-alpine, armbuild/alpine), I couldn't vouch for them and I don't have the hardware to test them.

Unknwon:
Sure, that's OK!

0rax:
@muhmuhten To be sure that the same version of alpine is used for both container, I upgrade to 3.3 before doing anythin so I will try to make the change. (And also see if there is an alpine 3.3 image worth, but yeah currently I am using this one as I trust the maintainer and just update to 3.3 to be sure that both container runs on the same version of the OS).
I will do the necessary testing and update what's needed today.

PR #2674, opened 2016-02-21
Little refactoring of diff highlight.

andreynering:
Moving cache variable to template instead of in the struct.


Unknwon:
Thanks!

PR #2685, opened 2016-02-23
Test SMTP settings during installation

Jonny007-MKD:
Hello,

I wanted to know whether my SMTP settings are correct without creating users/commits and looking for the notification mails. So I implemented a mailer.Test method that is called in InstallPost() for checking the values. The test works quite well, but I think I messed things up ;)

 - go-bindata isn't working at my place, so I could not work further on locale
 - I couldn't find a reason why you are using your own Send method, so I removed it and completely use gomail.v2 instead.
 - templates/home.tmpl seems to be deleted :-1:

This would fix #2432.

Some notes:
 - I didn't test TLS client auth
 - I didn't test sending (multiple) mails using gomail (in `processMailQueue`)
 - Are there more (SMTP) errors that should be implemented?
 - We should validate the field `form.SMTPFrom`
 - In most cases the `Sender Email` is an email address, but sometimes it's simply a plain username. So checking for a valid mail address seems to be wrong.
 - The term `Sender Email` isn't quite clear. Other mail clients use `Username`


Unknwon:
Thanks your PR!

But I can't accept until you fix the `templates/home.tmpl` problem first.

Unknwon:
> I couldn't find a reason why you are using your own Send method, so I removed it and completely use gomail.v2 instead.

From what I remember, gomail.v2 missing some important support that Gogs allows more detailed configuration about SMTP.

Since the purpose of this PR is support test SMTP, please remove irrelevant changes to the purpose and possibly make another PR.

Unknwon:
Any updates on thsi?

Jonny007-MKD:
If gomail.v2 is missing some important feature I have to implement the test using the existing Send() function (to support this feature while sending, too). To avoid duplicate code I would have to split the Send() function in three parts (Dial, Send, Close), where the test can leave out the Send part.

But I'm not happy with this. In the Gogs code I see almost nothing that gomail cannot do. Here my analysis of [mailer.go Send()](https://github.com/gogits/gogs/blob/master/modules/mailer/mailer.go#L81) compared to [gomail smtp.go and auth.go](https://github.com/go-gomail/gomail/blob/v2/):

 - L81: `SplitHostPort` we have to do this anyways
 - L86: `tlsConfig` is the same structure that gomail uses
 - L92: I cannot see any code about using the X.509 certs, so I suppose that the TLS stack is using them in gogs as well as in gomail.
 - L99: We wouldn't have to dial on our own
 - L107: SMTPS (port 465) support also exists
 - L112: SMTP dial could also be removed
 - L117: `DisableHelo` can achieved by setting the gomail `HeloHostname` to `""` 
 - L132: STARTTLS is supported
 - L139: gomail does auth, but **not CRAM-MD5**
 - L159: MAIL FROM is supported by gomail, too :)
 - L163: Multiple RCPT TO, too
 - L169: `Data()` and `msg.WriteTo(w)` are used by gomail, too

So the only difference I see is the CRAM-MD5 auth, which could and should be added to gomail. I opened an [issue](https://github.com/go-gomail/gomail/issues/52).

The advantages of completely using gomail are less code (to maintain) and the ability to send multiple mails over one SMTP connection.

Unknwon:
Thanks for the details!

But please keep this info for next PR or discuss on https://discuss.gogs.io.

For this PR, please cleanup the commits first.

Jonny007-MKD:
Let's close this PR, as it includes many weird commits (I didn't know what I'm doing).
I am also not going to implement some test with the current Send() routine as long as gomail.v2 may implement CRAM-MD5. As soon as that's clear, I will create a new PR.

alexcesaro:
> From what I remember, gomail.v2 missing some important support that Gogs allows more detailed configuration about SMTP.

Hello @Unknwon, I'm the author of Gomail. If you think any feature is missing in Gomail do not hesitate to open an issue.

Congrats for this great project!

Unknwon:
Thanks @alexcesaro !

Gomail is great and does solve many problems for me, the `Sender` part is the only thing I remember not directly use Gomail. Like `CRAM-MD5` that I did not realize how to use it.

I'll wait for PR from @Jonny007-MKD and do feature request as need.

PR #2694, opened 2016-02-24
Improved Pull Request refs

mhartkorn:
The previous PR #2524 was not creating `refs/pull/<pr-id>/head` when a PR was freshly created but only after a explicit push to the corresponding head branch.

This PR adds this functionality (issue #1655).


Unknwon:
Thanks!

PR #2706, opened 2016-02-26
Image attachments keep aspect ratio

ChubbyNinja:
Currently attachments in issues are stretched into their 150x150px box - most of the time this looks fine but for long, non-square or icons ize images this can soon look untidy and pixelated

I have made CSS changes to change them to look like this:
example-attachments_r1_c1


Unknwon:
Thanks!

Unknwon:
Please help test on https://try.gogs.io

PR #2707, opened 2016-02-26
Docker Container update

0rax:
- Following #2665 I've updated the `docker/build.sh` script and the `Dockerfile.rpi` to use the `--no-cache` option of `apk-tools`
- Updated `gosu` to latest version (v1.7)
- `apk upgrade` call is still present in `Dockerfile.rpi` to make sure that the container is running on `alpine 3.3`. This is necessary to ensure the RPi container is running on the same environment as the standard container.
- Updated the `.dockerignore` file to add new undesired file in the Docker context
- Update `docker/README.md` to remove a resolved known issue about automated Docker Hub build & added a note about the fact that we currently do not support container building on the Raspberry Pi 1

This PR has been tested on OS X (Boot2Docker w/ Docker 1.9 & 1.10.2) and on Raspberry Pi 2 (Arch Linux w/ Docker 1.10.2)


Unknwon:
Thanks!

PR #2708, opened 2016-02-26
template ssh url with other username (works only with  build-in/internal ssh server)

genofire:
I hope it will work.
I trying to test it.


Download-Fritz:
This cannot work... all you did was adding an option to display a different name without any changes to the actual code.

genofire:
Thats the idee, the internal gogs-ssh server accept every username.

Download-Fritz:
Sorry, yeah, with the internal SSH server, this makes more sense. Though probably is a good idea to check if the user defined even exists, and if not, fall back to RunUser, no?

Unknwon:
> Thats the idee, the internal gogs-ssh server accept every username.

Aha, never noticed that.

But I would consider it is a security flaw? Not sure though, because built-in SSH server only allows git commands.

genofire:
It looks so (at the first view) that the built-in only allows git commands. Nothing matter.
For gogs/go it is not possible to break out of the `RUN_USER` which is equal of the current user.
(Why is the settings `RUN_USER` nessesary?)

So this PR is just a beautiful think, to rewrite the username in URLs.

Unknwon:
@genofire have you tested these changes?

genofire:
Jap, it works.
(yesterday night i was to stuppit to relace the import urls -> from github.com/gogits/gogs to github.com/genofire/gogs :( )

MayBe we shoud recall the name of the variable from `SSH_USER` to `SSH_TEMPLATE_USER` or so, to make it more clear.

Unknwon:
> MayBe we shoud recall the name of the variable from SSH_USER to SSH_TEMPLATE_USER or so, to make it more clear.

Yes I prefer `SSH_TEMPLATE_USER`.

genofire:
Done!
I add also a check it it runs with the internal/buildin ssh server.

Unknwon:
Thanks, but I there the PR has some problem here. Please do a rebase against `develop` branch.

genofire:
xD dadurch kamm doch erst die Probleme :( -> https://github.com/gogits/gogs/compare/develop...genofire:develop?expand=1

Unknwon:
In a second thought it is not consistent to other parts of Gogs, even if the username is not used, allow user to config arbitrary values cause more confusions.

PR #2717, opened 2016-02-28
Change colour of numbers on pull request tabs if greater than zero

chriswatt:
Adding the same condition to the pull request tabs that the main tabs have.

Before
gogsprbefore

After
gogsprafter


Unknwon:
Thanks!

Issue #2718, opened 2016-02-28
Download as archive button on wiki pages downloads main repo archive, not wiki archive


chriswatt:
![gogsdownloadarchive](https://cloud.githubusercontent.com/assets/488912/13382854/5134f17e-de76-11e5-9a87-6195446486b2.PNG)

Unknwon:
I think it is actually a bug to show this button on wiki page.

ultralisk-:
If no one else has claimed it I'll remove the hide the button.

chriswatt:
If no one else has claimed it I'll remove the hide the button.

Unknwon:
Fixed by #2722.

PR #2722, opened 2016-02-29
Hide the download archive button when viewing wiki pages

chriswatt:
After 
gogsdlwikibutton


chriswatt:
Fixes #2718

Unknwon:
I think you can remove that part completely.

chriswatt:
New commit removing the element entirely.

Unknwon:
Thanks!

PR #2733, opened 2016-03-02
ISSUE_TEMPLATE: suggestion to test at try.gogs.io

neolit123:
I've noticed that a lot of issues cannot be reproduced on http://try.gogs.io,
which either hints about specific database type problems or
hints about bugs which are already solved in the newer version
(as http://try.gogs.io is usually a newer build).

This patch adds the suggestion to test the issue at http://try.gogs.io in
the Github "issue template". The user can answer: "Yes", "No", "Haven't tried",
"Not relevant".

The "Haven't tried" answer is there to make this test *optional* to the user,
as forcing them to try the website might be a bit too much. Hopefully *some* users
will test the bug there and that will reduce the work that the maintainers have
to do in the issue tracker, asking the same question all the time.

"Not relevant" is an option where testing on http://try.gogs.io makes no sense as
the bug is unrelated to the Web UI or is very specific in nature.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -14,7 +14,6 @@ For bug reports, please give the relevant info:
 - Can you reproduce the bug at http://try.gogs.io:
   - [ ] Yes
   - [ ] No
-  - [ ] Haven't tried
   - [ ] Not relevant
 - Log gist:
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -11,6 +11,11 @@ For bug reports, please give the relevant info:
   - [ ] PostgreSQL
   - [ ] MySQL
   - [ ] SQLite
+- Can you reproduce the bug at http://try.gogs.io:
+  - [ ] Yes
+  - [ ] No
+  - [ ] Haven't tried
+  - [ ] Not relevant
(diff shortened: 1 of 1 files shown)


Unknwon:
I think you should remove option `- [ ] Haven't tried`, because I will ask to test eventually.

neolit123:
> I think you should remove option - [ ] Haven't tried, because I will ask to test eventually.

OK, i will remove it and resend the PR.

Unknwon:
You can update the branch and PR will be updated.

neolit123:
> You can update the branch and PR will be updated automatically.

Done, let me know if you want a rebase of the branch, to squash the 2 commits.

Unknwon:
> let me know if you want a rebase of the branch, to squash the 2 commits.

Yes, a rebase would be awesome!

neolit123:
closing this PL, the rebase doesn't work because it seems as if github has locked the branch.

Unknwon:
You actually need to do soft reset.

neolit123:
i squashed the two commits into one and tried to do a "git push --force <myrepo> <the-PL-branch>".
but it says "<the-PL-branch> does not match any. failed to push some refs to...".

i've done a rebase with a "git push --force" before, but not for a branch which is currently under a pull request, so i though it might be locked of sorts. no idea...

Issue #2696, opened 2016-02-24
Show if a commit is a merge in the commit view


SarenCurrie:
It would be good if merge commits were differentiated in the commit view by being breed out similar to how bitbucket does it:
<img width="1484" alt="screen shot 2016-02-25 at 11 27 24 am" src="https://cloud.githubusercontent.com/assets/8222891/13303254/15ef94c2-dbb3-11e5-91db-493cf02990d3.png">

Unknwon:
Good suggestion!

But is there an official way from Git to determine if a commit is a merge?

SarenCurrie:
I assume if a commit has more than one parent it's a merge?

Unknwon:
Yeah, that is a reasonable thought.

Any git experts could help on this?

p-v:
I noticed in some of my projects that Bitbucket doesn't always grey out the merge commit. Looks like it only greys out when there are no merge conflicts.

Unknwon:
@p-v thanks for the info!

> Looks like it only greys out when there are no merge conflicts.

Not dig into it, but feels like complicted things somehow.

Rukenshia:
> I assume if a commit has more than one parent it's a merge?

that would grey out every commit that is merged, not just the "Merge" commit (as shown in your screenshot). The merge commit itself only has a single parent afaik.

Unknwon:
Hmm.. so how does BitBucket determine that..

Rukenshia:
According to #2761 it does seem to work, though, that gets me curious now. When I tested my own solution for that (which basically is the same as #2761), the single commit I merged was greyed out, too.

Unknwon:
@Rukenshia What is your Go version?

Rukenshia:
1.6, but what does it have to do with it?

Unknwon:
@Rukenshia our of ideas..

p-v:
> that would grey out every commit that is merged, not the "Merge" commit (as shown in your screenshot). The merge commit itself only has a single parent afaik.

@Rukenshia A merge commit will always has more than one parent. Even in the screenshot you can clearly see in the graph on the left that the merge commit has two parents.

Rukenshia:
@p-v what I tried locally was the following:

1. Create new Repository
2. Create "README" on master
3. Create new branch
4. Checkout new branch
5. Update "README" file
6. Merge new branch onto master

[...]

PR #2761, opened 2016-03-05
Grey out merge commits

soudy:
Implement #2696
Greyed out commit

### DIFF
--- a/templates/repo/commits_table.tmpl
+++ b/templates/repo/commits_table.tmpl
@@ -35,7 +35,12 @@
 								<img class="ui avatar image" src="{{AvatarLink .Author.Email}}" alt=""/>&nbsp;&nbsp;{{.Author.Name}}
 							{{end}}
 						</td>
+
+						{{if gt .ParentCount 1}}
+						<td class="message collapsing has-emoji grey text">
+						{{else}}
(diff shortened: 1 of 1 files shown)


templates/repo/commits_table.tmpl
```diff
@@ -35,7 +35,12 @@
 								<img class="ui avatar image" src="{{AvatarLink .Author.Email}}" alt=""/>&nbsp;&nbsp;{{.Author.Name}}
 							{{end}}
 						</td>
+
+						{{if gt .ParentCount 1}}
+						<td class="message collapsing has-emoji grey text">
```
Unknwon: Why don't you just put `grey text` inside the `if` block, instead of the whole line?

soudy: @Unknwon because stretching out HTML tags with if statements makes it way less readable imo

Unknwon: Disagree. Duplicated majority of content is worse, unless case branches are a lot. In this case we only have two branches.

soudy:
@Unknwon I've updated it to your likings

Unknwon:
Thanks!

Issue #2763, opened 2016-03-06
Allow to delete current avatar of organizations


xxxtonixxx:
An image
![image](https://cloud.githubusercontent.com/assets/7110786/13552165/d6322736-e356-11e5-98bf-6b34d3747ea8.png)

Unknwon:
Same thing has been done for user settings, if someone want to take this, claim and go ahead.

Unknwon:
This is claimed to be implemented by merging https://github.com/gogits/gogs/pull/2766

Unknwon:
Close as implemented.

PR #2766, opened 2016-03-06
Added: Ability to delete org avatar.

moltam:
Implements #2763.

I created a new DeleteAvatar() method in the User model. Both settings router (user and org) uses this method.

Because the 'Enable Custom Avatar' option is not present in the org's settings form, after deleting the org's avatar no new custom avatar is generated, the default Gravatar picture is displayed (https://secure.gravatar.com/avatar/?s=100). Does this needs a fix? May I include the option in the org's form too, or generate a random avatar after deleting the previous?

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -368,6 +368,7 @@ func runWeb(ctx *cli.Context) {
 				m.Combo("").Get(org.Settings).
 					Post(bindIgnErr(auth.UpdateOrgSettingForm{}), org.SettingsPost)
 				m.Post("/avatar", binding.MultipartForm(auth.UploadAvatarForm{}), org.SettingsAvatar)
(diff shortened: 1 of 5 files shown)


Unknwon:
> Because the 'Enable Custom Avatar' option is not present in the org's settings form, after deleting the org's avatar no new custom avatar is generated, the default Gravatar picture is displayed (https://secure.gravatar.com/avatar/?s=100). Does this needs a fix? May I include the option in the org's form too, or generate a random avatar after deleting the previous?

I think you didn't config to disable Gravatar, if you do, I think there will be a random one generated.

moltam:
You are right. If I disable Gravatar, a random avatar is generated on orgs too, after deleting the custom one.

PR #2772, opened 2016-03-06
Add ru-RU to home template

OhDaeto:
Added russian traslation for main page
https://r208.tk/ for check this commit and click on "Русский"


Unknwon:
Thanks!

PR #2777, opened 2016-03-07
Completely use gomail.v2 and test SMTP settings during install

Jonny007-MKD:
Here comes my new PR :)

@alexcesaro added support for automatic CRAM-MD5 auth in gomail.v2 so we are ready to completely rely on gomail.v2.

Changes:

 - removed the old `Send()` function
 - implemented the Dialer for gomail.v2
 - added synchronous `mailer.Send()` function (for `mail.SendTestMail()`, not tested)
 - added SMTP test functionality in mailer and install (tested)

Many tests I haven't done:

 - What if the mailserver doesn't need/support auth (like localhost servers)
    - Shall we accept empty credentials? Is this something we want to support?
    - Do we accept empty credentials?
    - Does gomail accept empty credentials?
    - Does gomail support SMTP servers without auth as currently used? (currently my local Exim4 returns: "503 AUTH command used when not advertised")
 - Sending a test mail
 - Sending mails over the queue (perhaps we should make the test mail async, too, and remove the sync function?)
 - Are all error messages clear to the user

More stuff:

 - As @alexcesaro stated in [gomail issue 52](https://github.com/go-gomail/gomail/issues/52), smtp packages **always** sends a HELO (with `"localhost"` by default), so the setting `DisableHelo` is useless.
 - The test fails if more specific settings (e.g. TLS client certs) have to be used which prevents the user from saving the settings. Is it clear for the user to use the INI file for this stuff?
 - It's the same with the field `Sender Email`: Normally it is called `Username` because it isn't an email address on all servers.


modules/mailer/mailer.go
```diff
@@ -72,120 +71,121 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
...
-	sender := &Sender{}
+	opts := setting.MailService
+
+	dialer, err := newDialer(opts)
+	if err != nil {
+		return
+	}
+
+	log.Debug("Mailer: Dialing %s", opts.Host)
+	conn, err := dialer.Dial()
```
Unknwon: I think there is no guarantee the connection will be valid as long as application lifetime.

modules/mailer/mailer.go
```diff
@@ -72,120 +71,121 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
...
-		ServerName:         host,
-	}
-
-	if opts.UseCertificate {
-		cert, err := tls.LoadX509KeyPair(opts.CertFile, opts.KeyFile)
-		if err != nil {
-			return err
-		}
-		tlsconfig.Certificates = []tls.Certificate{cert}
+		log.Error(3, "Mailer: Failed convert hostname %s: %v", opts.Host, err)
```
Unknwon: Only need to return formatted error, not print them. Same as other places.

modules/mailer/mailer.go
Jonny007-MKD: That's true, I didn't think the method runs for the whole lifetime. I've got two suggestions:

 -  @alexcesaro what happens when the server closes the connection. Can we detected and recover from that easily?
 -  close the connection when the queue is empty (for some seconds) and reopen it in the next iteration

Am 13. März 2016 22:07:42 MEZ, schrieb "无闻" <notifications@github.com>:

alexcesaro: You could do something like this example: https://godoc.org/gopkg.in/gomail.v2#example-package--Daemon

alexcesaro: I just updated Gomail to automatically redial in case of a timeout: https://github.com/go-gomail/gomail/commit/afff51fd8cbc246672451a1b0b67e8f4a507bad8

Jonny007-MKD: @alexcesaro: Sweet, thank you!
[RFC 2821](https://www.ietf.org/rfc/rfc2821.txt) recommends several timeouts in the order of several minutes per command. So I think I will implement the timeout as shown in the [example](https://godoc.org/gopkg.in/gomail.v2#example-package--Daemon) to be friendly to the SMTP server :)

Jonny007-MKD:
I tried to send test emails over my mail server **without auth**. The error messages it gives me are quite cryptical and I don't think, every admin will understand the underlying problem:

  > 554 5.7.1 <doesntexist at xx.com>: Relay access denied
  > 450 4.2.0 <something at myserver.com>: Recipient address rejected: Greylisted
  > 550 5.7.1 rejected by DMARC policy for nordlichter-brv.de

Should we append a notice to the error message if the test mail fails and no credentials were given?

modules/mailer/mailer.go
```diff
@@ -45,151 +44,141 @@ func NewMessage(to []string, subject, body string) *Message {
...
+				return nil, err
 			}
 		}
-
-		if err = client.Hello(hostname); err != nil {
-			return fmt.Errorf("Hello: %v", err)
-		}
+		dialer.LocalName = hostname
+	} else {
+		dialer.LocalName = ""
```
Unknwon: I think this is redundant? Default value is empty string already.

modules/mailer/mailer.go
```diff
@@ -45,151 +44,141 @@ func NewMessage(to []string, subject, body string) *Message {
...
+	}
+	open := false
+	var conn gomail.SendCloser
 
 	for {
 		select {
+		// A new message is pending
 		case msg := <-mailQueue:
+			// Dial SMTP if neccessary
+			if !open {
```
Unknwon: I thought gomail.v2 has auto-redial already? https://github.com/go-gomail/gomail/commit/afff51fd8cbc246672451a1b0b67e8f4a507bad8

Jonny007-MKD: Yes, that's true.
But in my opinion it is an unfriendly behavior of a client to leave the connection open for minutes and thus consuming resources on server and client side without a real need.

Unknwon: OK, but your code is not setting `open` back to `false` anywhere.

modules/mailer/mailer.go
```diff
@@ -45,151 +44,141 @@ func NewMessage(to []string, subject, body string) *Message {
...
+				var err error
+				if conn, err = dialer.Dial(); err != nil {
+					panic(err)
```
Unknwon: `panic` is not acceptable, print log.

modules/mailer/mailer.go
```diff
@@ -155,7 +155,7 @@ func processMailQueue() {
 			if !open {
 				var err error
 				if conn, err = dialer.Dial(); err != nil {
-					panic(err)
+					log.Error(4, "Fail to connect to SMTP server: %v", err)
```
Jonny007-MKD: Perhaps we should check whether this is a temporal or a permanent error, reenqueue the message and try again later?

modules/mailer/mailer.go
Jonny007-MKD: I missed that line :)

Makefile
```diff
@@ -11,7 +11,7 @@ RELEASE_ROOT = "release"
 RELEASE_GOGS = "release/gogs"
 NOW = $(shell date -u '+%Y%m%d%I%M%S')
 
-.PHONY: build pack release bindata clean
```
Unknwon: Why do you have changes to `Makefile`?

Jonny007-MKD: Oh sorry, I removed bindata to test the build and accidentally committed it... I'll revert that.

templates/admin/config.tmpl
```diff
@@ -152,7 +152,6 @@
 							<dt>{{.i18n.Tr "admin.config.mailer_name"}}</dt>
 							<dd>{{.Mailer.Name}}</dd>
 							<dt>{{.i18n.Tr "admin.config.mailer_disable_helo"}}</dt>
-							<dd><i class="fa fa{{if .Mailer.DisableHelo}}-check{{end}}-square-o"></i></dd>
```
Unknwon: Why remove this line?

Jonny007-MKD: As @alexcesaro pointed out in go-gomail/gomail#52, the smtp package always sends a HELO (with "localhost" by default), so the setting DisableHelo is useless.

modules/mailer/mailer.go
```diff
@@ -45,151 +44,138 @@ func NewMessage(to []string, subject, body string) *Message {
...
-	hasStartTLS, _ := client.Extension("STARTTLS")
-	if !isSecureConn && hasStartTLS {
-		if err = client.StartTLS(tlsconfig); err != nil {
-			return fmt.Errorf("StartTLS: %v", err)
+		if terr, ok := err.(*textproto.Error); ok {
+			terr.Msg = "Failed to connect: " + terr.Msg
 		}
+		return err
+	} else {
+		conn.Close()
```
Unknwon: Why not use `defer conn.Close()`?

modules/mailer/mailer.go
```diff
@@ -45,151 +44,138 @@ func NewMessage(to []string, subject, body string) *Message {
...
-	if err != nil {
-		return fmt.Errorf("Data: %v", err)
-	} else if _, err = msg.WriteTo(w); err != nil {
-		return fmt.Errorf("WriteTo: %v", err)
-	} else if err = w.Close(); err != nil {
-		return fmt.Errorf("Close: %v", err)
+	if err := conn.Send(opts.From, msg.GetHeader("To"), msg.Message); err != nil {
+		return err
+	} else {
+		log.Trace("E-mail sent %s: %s", msg.GetHeader("To"), msg.Info)
```
Unknwon: You don't need `else` block since you have `return` in `if`.

Unknwon:
Hi, can make separate PRs, this single PR contains too many changes.

Issue #552, opened 2014-10-11
HTTP Security Headers


sts:
Enable various security extensions to prevent common attacks such as - list taken from [OWASP.org](https://www.owasp.org/index.php/List_of_useful_HTTP_headers)
- Strict-Transport-Security:max-age=31536000; includeSubdomains; preload - enforces secure connections to the server.
- X-Frame-Options:deny - provides clickjacking protection.
- X-XSS-Protection:1; mode=block - enables the Cross-site scripting (XSS) filter built into most recent web browsers.
- Content-Security-Policy

There might be configuration options needed. Also the session cookie should be set to [HttpOnly](https://www.owasp.org/index.php/HttpOnly).

Also including maxcdn resources into the codebase, would reduce the possibility that external parties are able to track users from using the application.

Unknwon:
Anyone knows better can review this?

Jonny007-MKD:
Where would you set these headers? In `modules/httplib/httplib.go`?
I would set those headers:

 -  `Strict-Transport-Security: max-age=31536000`

    active: only when CERT_FILE and KEY_FILE are set (so HTTPS is used), otherwise it's an option for the proxy
    Maybe we should an option for the max-age value; I (personally) never change it.

    https://scotthelme.co.uk/hsts-the-missing-link-in-tls/

 - `Content-Security-Policy: "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' *.gravatar.com; style-src 'self' 'unsafe-inline'; font-src 'self'; frame-src 'none'; object-src 'none'"

    Here I disabled `<frames>` and `<objects>`. Are there more external sources (like Fonts, CDNs) we have to add? Should the user be able to set own sources? Is JS `eval()` used?

    https://scotthelme.co.uk/content-security-policy-an-introduction/

 - `Public-Key-Pins 'pin-sha256="xxx"'`

    Could be used when HTTPS is used, but it may render a domain useless and should not be activated by default. We could let the user decide to enable this after a throrough warning

    https://scotthelme.co.uk/hpkp-http-public-key-pinning/

 - `X-Frame-Options "SAMEORIGIN" always;`

    The user should be able to disable it in case he wants to use gogs in a `<frame>`.

    https://scotthelme.co.uk/hardening-your-http-response-headers/#x-frame-options

 - `X-Content-Type-Options "nosniff" always;`

    As long as gogs is sending correct mime-types this should be no problem and always active.

[...]

Issue #2211, opened 2015-12-16
Config option to auto-redirect HTTP to HTTPS


yurivict:
This is the normal practice and about every site does this.

Jonny007-MKD:
Could be integrated into #2782:

    if http && redirect enabled {
        send 302 code with https URL
    }

Jonny007-MKD:
I stumpled upon the *Upgrade-Insecure-Requests* HTTP header. We can support this as well.
https://www.w3.org/TR/upgrade-insecure-requests/#preference

Jonny007-MKD:
This redirect feature implicates listening on two ports: 80 and 443. Otherwise no redirect is possible.

Unknwon:
> This redirect feature implicates listening on two ports: 80 and 443. Otherwise no redirect is possible/usable

It does not have to, if behind a reverse proxy, Gogs only listen on HTTP, the web server (such as NGINX) handles HTTP.

Jonny007-MKD:
That's true :)

But after a quick search it seems as nginx by default is not telling gogs whether it is serving a http or a https request. If there is more to do in nginx setup than setting `proxy_pass` then the forwarding can be done in nginx as well.

Unknwon:
@Jonny007-MKD that's the point, this is very very rare case.

Jonny007-MKD:
So I'd say as long as gogs doesn't implement a second listener just for this feature it is no case at all.

Unknwon:
Yes, might close this issue.

Unknwon:
I think it is more likely a job for reverse proxy.

PR #2782, opened 2016-03-07
Security headers

Jonny007-MKD:
See #552 

Comments where code needed. @Unknwon: you are allowed to push to my branch.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -171,6 +171,7 @@ func newMacaron() *macaron.Macaron {
 		},
 	}))
 	m.Use(middleware.Contexter())
+	m.Use(middleware.SecureHeaders)
 	return m
 }
 
--- a/modules/middleware/secureheaders.go
+++ b/modules/middleware/secureheaders.go
@@ -5,6 +5,9 @@
 package middleware
 
 import (
+	"strconv"
+
+	"github.com/gogits/gogs/modules/log"
 	// "github.com/gogits/gogs/modules/setting"
 )
 
@@ -12,12 +15,25 @@ func SecureHeaders(ctx *Context) {
 	// enable XSS protection in Internet Explorer, Chrome & Safari
 	ctx.Resp.Header().Set("X-Xss-Protection", "1; mode=block")
 
-	// Allow framing only for own domain
-	ctx.Resp.Header().Set("X-Frame-Options", "SAMEORIGIN")
-
 	// Disallow mime sniffing in Internet Explorer & Chrome
 	ctx.Resp.Header().Set("X-Content-Type-Options", "nosniff")
 
 	// Only allow loading of resources of these domains
 	ctx.Resp.Header().Set("Content-Security-Policy", "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' '*.gravatar.com'; style-src 'self' 'unsafe-inline'; font-src 'self'; frame-src 'none'; object-src 'none'")
+
+	// Allow framing only for own domain
+	ctx.Resp.Header().Set("X-Frame-Options", "SAMEORIGIN")
+
+	// if (we are delivering https) {
+		// maxAge := get value from setting, unit: seconds, default: "31536000" (1 year)
+		maxAge := "31536000"
+		maxSeconds, err := strconv.Atoi(maxAge)
+		if err != null {
+			// check skip value
+			// insert name of setting here
+			log.Error(4, "Invalid value for HSTS setting: %v", err)
+			maxSeconds := 31536000
+		}
+		ctx.Resp.Header().Set("Strict-Transport-Security", "max-age=" + strconv.Itoa(maxSeconds))
+	// }
 }
--- a/modules/middleware/secureheaders.go
+++ b/modules/middleware/secureheaders.go
@@ -1,0 +1,23 @@
+// Copyright 2014 The Gogs Authors. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+package middleware
+
+import (
+	// "github.com/gogits/gogs/modules/setting"
+)
+
+func SecureHeaders(ctx *Context) {
+	// enable XSS protection in Internet Explorer, Chrome & Safari
+	ctx.Resp.Header().Set("X-Xss-Protection", "1; mode=block")
+
+	// Allow framing only for own domain
+	ctx.Resp.Header().Set("X-Frame-Options", "SAMEORIGIN")
+
+	// Disallow mime sniffing in Internet Explorer & Chrome
+	ctx.Resp.Header().Set("X-Content-Type-Options", "nosniff")
+
+	// Only allow loading of resources of these domains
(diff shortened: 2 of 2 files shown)


roblabla:
Aren't there potentially other sources for img ? Like the gravatar provider duoshuo, and maybe inline images (I don't know if those actually exist). Or images in the README pointing from outside sources. I think the Content-Security-Policy implemented here is way too strict and probably breaks a ton of things.

Jonny007-MKD:
Hum, that's true. We should allow all image sources. After a quick research I couldn't find an 'all' value!?

How about other resources? Here is a list of resource types: https://www.owasp.org/index.php/Content_Security_Policy

Am 8. März 2016 02:41:22 MEZ, schrieb Robin Lambertz <notifications@github.com>:

emansom:
This header should do ths job nicely: `Content-Security-Policy: upgrade-insecure-requests`

[Documentation on upgrade-insecure-requests](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#upgrade-insecure-requests)

[See vader.jpg being upgraded to HTTPS automagicly](https://git.alex-dev.org/alex/Icarus/issues/1)

Jonny007-MKD:
My new suggestion:

 - `default-src 'self'` (sets the values for the following:)
    - `child-src 'self'` (no framing of external domains)
    - `connect-src 'self'` (no XMLHttpReq, Websocket, ... to external domains)
    - `font-src 'self'` (`@font-face` only from this domain)
    - `manifest-src 'self'` (?)
    - `media-src 'self'` (`<audio>` and `<video>`. Is markdown generating these with external URIs? Then `*` would be applicable)
    - `script-src 'self'` (only allow `<script>` in HTML header)
 - `img-src *` (all image sources)
 - `style-src 'self' 'unsafe-inline'` (`<style>` in HTML header and body. Do we need body?)
 - `frame-src 'none'` (no `<frame>`)
 - `object-src 'none'"` (no `<object>`, `<embed>`, `<applet>`)
 - `upgrade-insecure-requests` (as suggested by @emansom)
 - `referrer no-referrer-when-downgrade` (some referrer privacy)
 - `reflected-xss block` (equals to the X-XSS-Protection header)

roblabla:
upgrade-insecure-requests should be configurable (not everyone has gogs behind https, some people might want to link images that aren't https, etc...). The rest looks great.

emansom:
There is no point serving these headers over plain HTTP, there will be no benefit from a security standpoint since traffic can still be MITMed. I suggest these headers only apply when traffic is served from HTTPS and/or gogs is running behind a reverse-proxy that serves HTTPS.

@Jonny007-MKD Seems like good defaults to me. Good job :)

Jonny007-MKD:
@roblabla: You're right. As we can assume that most servers support https, we cannot say this for each server, some may support http only. The spec sounds as if the request fails in this case, and without an option to disable the upgrading there would be no way around that. WIthout upgrading the user gets a warning about insecure content, which isn't nice, too. So IMO it's up to the admin to decide what to do. I'd go with upgrade-insecure-requests enabled by default if serving HTTPS requests.

@emansom: I don't agree. Only (good) SSL can prevent MITM; this is not the purpose of these headers:
 - HSTS forces the browser to always use HTTPS. Over HTTP a MITM may remove this header, but if the user connects to the server without a MITM only once, he is out.
 - CSP prohibits loading resources from unwanted sources (like the attackers server). The attacker may find a way to insert a `<script>` tag in a comment, to `eval()` some code or to include some flash object. With CSP this isn't possible anymore.

roblabla:
@emansom You're right that the headers themselves are mostly important on HTTPS-enabled gogs. However, adding upgrade-insecure-requests to a gogs with lots of mixed-contents (http img memes for instance) would break stuff. Probably nothing *major*, just some images would become broken. But still worth considering :)

A configurable, secure-by-default upgrade-insecure-requests sounds reasonable to me :D

Unknwon:
Just a quick comment: I would like to have a config to disable/enable security header at all, and defaults to disabled.

roblabla:
Why should the security headers be disabled by default ? I think it would make sense for them to be enabled by default, as they make sure some potential security failings aren't exploitable.

Unknwon:
@roblabla there are many possible failing cases by adding these header which could break the entire application. 

Should not be enabled until long-term tested.

Jonny007-MKD:
This is true for CSP and HSTS, if the user doesn't know what he is doing (and we don't know what is used or there). Those two are aimed to be configurable, a disable is not in there, yet.
The other two are very unlikely to break anything :)

Unknwon:
It's been quite a while and not many people comments on that thread; I would consider this to be completely optional.

mhartkorn:
I wouldn't implement HSTS. You can make your whole installation unaccessible if you mess up the SSL certs and have HSTS enabled.

This might sound like "the admin was just dumb and deserves this" but in my eyes Gogs is often used at smaller teams where one developer has the "burden" to maintain the infrastructure. So you cannot assume that all Gogs admins are very savvy when it comes to SSL and its quirks. Mix an outdated SSL cert with a weekend and a big part of the development comes to a stop because nobody can access Gogs.

Jonny007-MKD:
@mhartkorn HSTS doesn't enlarge the trouble when SSL certs are messed up. It only prevents you from using HTTP. I suppose you mix HSTS with Public Key Pinning.

HSTS forces the browser to use HTTPS. If you request a website with an outdated SSL cert the browser gives you a warning which you can ignore. Afterwards, you can use the site just as usual. I don't know any browser that doesn't allow ignoring outdated certs.

`git pull` afaik ignores HSTS anyways and uses HTTPS or HTTP, just as configured.

mhartkorn:
@Jonny007-MKD I know but imagine an admin thinks "oh what's this, let's test that" and then has no certificate ready. You cannot make a browser forget the HSTS header with cache cleaning, sadly. I think the other headers are fine but HSTS and HPKP are in my opinion nothing that Gogs should support at the moment. Maybe in a few years when SSL is really the default for browsing the web and certificates are easier to obtain.

Jonny007-MKD:
@mhartkorn After some googling it's relatively easy to forget HSTS settings ([here](http://classically.me/blogs/how-clear-hsts-settings-major-browsers)). It indeed is nothing for "normal" users, but for admins that just changed some setting and know what to google for. Also, this can be warned about & linked in the documentation.

Perhaps we should only send HSTS headers when the connection is HTTPS, i.e. when the ROOT_URL starts with `https://`. In this case no HSTS setting can be saved without a valid certificate. As discussed in #2211 a HTTP->HTTPS redirection should not be done by Gogs but by the reverse proxy, so HSTS over HTTP is no good solution there, too.

Unknwon:
Seems having conflict.

bkcsoft:
In my opinion, just allow for setting an array of custom headers, since it will break for some :)

Jonny007-MKD:
If a normal user without HTTP background shall administer the gogs installation, it is quite unlikely that he will change an option in the INI to add sensible HTTP headers.

The aim of this PR is to improve the security of HTTPS connections for most users which wouldn't be achieved with custom headers.

mhartkorn:
Maybe add it into the config file as an example that is commented out with an appropriate description.

Unknwon:
Please make default to not enable security headers, and fix the conflict.

Issue #2732, opened 2016-03-02
Admin users cannot see organisations private repositories


xor-gate:
- Gogs version: 0.8.43.0223
- Operating system: Debian Linux 8 amd64
- Database:
  - [ ] PostgreSQL
  - [X] MySQL
  - [ ] SQLite

## Description

As a Admin user I cannot see the private repositories on the organisation page. When managing teams with repository access the autocomplete is probably also broken due to non-visible private repositories.

To reproduce probably:
* Create organisation
* Create private repos in organisation
* Dont add Admin user to any team
* As admin user browse to the organisation landing page
* See the organisation page is empty
* When adding the Admin user to the Owners it is still only able to view the Owner Team added repositories
* When adding "all" repositories of the organisation into the Owner Team it is visible.

Together with this, the Owner team should see all repositories in the Organisation or am I wrong?
Probably I should create a seperate issue for that.

When something is not clear or reproducable. Please ask.

neolit123:
hello,

```
Create organisation
Create private repos in organisation
Dont add Admin user to any team
As admin user browse to the organisation landing page
See the organisation page is empty
```

i can confirm the above; haven't tried the rest.
perhaps the admin user should be able to view all repositories regardless of the "private" flag.

xor-gate:
Yeah you are right about:

> perhaps the admin user should be able to view all repositories regardless of the "private" flag.

Unknwon:
Thanks your feedback!

Can you make a example organization on https://try.gogs.io?

xor-gate:
@Unknwon I have created:

* Organisation `1337org`
* Private repo `1337privaterepo`
* Public repo `1337publicrepo`

See https://try.gogs.io/1337org/

Unknwon:
@xor-gate thanks! As admin, I can navigate to the private repository through admin panel, but yeah, it should be a display problem. Thanks!

xor-gate:
You are right about them displaying/showing up under the admin panel. But having them not displayed in the view of the organisation is a bit odd when I first noticed.

mhartkorn:
Seems to be working with `master` now.

Unknwon:
@mhartkorn really?... I don't fix it at all...

mhartkorn:
Funny. I checked out the latest Docker container and compiled from source to make sure. I mean, it's working for me at the moment. The corresponding SQL query for getting the repos from organizations also changed which I explicitly checked before updating. Can someone else test it with the latest `master`, please?

[...]

PR #2785, opened 2016-03-08
Allowing site admins to view private repositories in org/home

Rukenshia:
Now with the proposed solution by @Unknwon.


Unknwon:
Which issue this addresses to?

Unknwon:
Found.

xor-gate:
Looking good, I will close #2732. When I see some Issues I will reopen. Thanks for your effort @Rukenshia !

PR #2793, opened 2016-03-10
Improving issue and pull request templates.

andreynering:
This makes the behavior a bit more compliant with GitHub's:

- Case insensitive:

`ISSUE_TEMPLATE`, `issue_template`, etc. all work.

- Any extension:

The file can have any extension (.txt, .md, etc) or none.

### DIFF
--- a/routers/repo/issue.go
+++ b/routers/repo/issue.go
@@ -17,6 +17,8 @@ import (
 	"github.com/Unknwon/com"
 	"github.com/Unknwon/paginater"
 
+	"github.com/gogits/git-module"
+
 	"github.com/gogits/gogs/models"
 	"github.com/gogits/gogs/modules/auth"
 	"github.com/gogits/gogs/modules/base"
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks your PR!

It feels like hurts performance quite a bit...

andreynering:
I didn't notice any performance penalty testing locally.

Anyway, I am open for suggestions.

Unknwon:
Used to loop 3 files, not 3 directories...

andreynering:
It's a good thing that we watch for performance, but I don't think it hurts in this case.

This code will only run on new issue/PR page. And it's seems instantaneous to me, I actually can't tell the difference. There's slower things in Gogs, as seeing a commit, as example.

Unknwon:
Since Gogs only supports render Markdown, what is worth to support all extensions?

andreynering:
The question is: should Gogs mimic GitHub in this case? GitHub supports it. If we don't have the intention to make Gogs compatible we can give up on this.

Unknwon:
Any format that is not Markdown, is not compatible already.

andreynering:
The idea is that the extension doesn't matter, but the content.

But now that you said, I realized that I didn't test if it crashes with a binary file there. It should be text, no matter the extension.

If you decide this PR is worth, I have to check this. But not today.

Unknwon:
I suggest keep this PR open and review maybe later.

andreynering:
I made a little change and now it only accepts no extension or `.md`.

We still have to loop over 3 folders, though, to make a case insensitive search. Will let you choose whether merge it or close.

Issue #2794, opened 2016-03-10
Cannot customize local wiki path


saboya:
Currently, the local wiki path is hardcoded to use AppDataPath, and, from what I could test, does not take GOGS_CUSTOM into consideration. So, it will always use the running path for the process + "data/". This leads to permission errors if you are running gogs from a path where the user does not have write permissions.

Unknwon:
Close as merged.

PR #2795, opened 2016-03-10
Adding a config entry to customize local wiki path.

saboya:
This is a proposed solution for #2794


Unknwon:
Thanks for your PR!

But it does not seem a good fix. Instead, you should add a config option for all `AppDataPath`.

saboya:
Closing, work done on #2796.

PR #2799, opened 2016-03-10
Fix logging adapter changes

### LINKED ISSUE #2798: Console logging ignores app.ini settings
- Gogs version (or commit ref): 0.9.0.0306
- Git version: 1.9.1
- Operating system: Ubuntu 14.04 LTS
- Database:
  - [ ] PostgreSQL
  - [ ] MySQL
  - [X] SQLite
  - [ ] Yes
  - [ ] No
  - [X] Not relevant
- Log gist:
```
2016/03/09 21:59:26 [T] Custom path: /home/git/gogs/custom
2016/03/09 21:59:26 [T] Log path: /home/git/gogs/log
2016/03/09 21:59:26 [I] [REDACTED] Gogs 0.9.0.0306
2016/03/09 21:59:26 [I] Log Mode: File(Info)
2016/03/09 21:59:26 [I] Cache Service Enabled
2016/03/09 21:59:26 [I] Session Service Enabled
2016/03/09 21:59:26 [I] Git Version: 1.9.1
2016/03/09 21:59:26 [I] SQLite3 Supported
2016/03/09 21:59:26 [I] Run Mode: Production
2016/03/09 21:59:26 [T] Doing: CheckRepoStats
2016/03/09 21:59:26 [I] Listen: http://127.0.0.1:3000
```
## Description

My config is:
```
[log]
MODE      = file
LEVEL     = Info
ROOT_PATH = <snip>/log
```

However, when launching Gogs, it does an early initialization of logging (`setting.go` line 234):

```go
log.NewLogger(0, "console", `{"level": 0}`)
```

This registers a logger that logs _all_ messages to the console, regardless of if they are debug messages or errors. Furthermore, it does not seem that the settings from `app.ini` are obeyed. For instance, this config fails to stop the logging:

```
[log]
MODE      = console
LEVEL     = Info

[log.console]
LEVEL     = Info
```

minecrafter:
Fixes #2798

This allows `config.ini` to override logging levels for the console.

### DIFF
--- a/modules/log/log.go
+++ b/modules/log/log.go
@@ -26,7 +26,7 @@ func NewLogger(bufLen int64, mode, config string) {
 	for _, l := range loggers {
 		if l.adapter == mode {
 			isExist = true
-			l = logger
+			logger = l
 		}
 	}
 	if !isExist {


Unknwon:
Thanks... but I don't understand why this fixes the problem?

minecrafter:
`l` is the element from the iterated scale. `logger` is the `Logger` that is being configured.

Unknwon:
Yes, but shouldn't new one overwrite the current one?

minecrafter:
`NewLogger` is configuring the existing logger with new setting, which seemed to be the intent here.

Unknwon:
Yes, that why assign `logger` to current one `l`...

Unknwon:
Any updates?

minecrafter:
This looked like a simple oversight which could be easily corrected. 
Your last response wasn't entirely clear to me - I thought that you were 
agreeing with me and was putting off the PR for other reasons.

Unknwon:
I don't understand why old one overwrite new one?

minecrafter:
To put this in simpler terms, `l` (which is from the `loggers` slice 
that is being iterated over) was effectively being ignored because the 
assignment was swapped.

Unknwon:
What do you mean by ignored?

The purpose of the loop is to find logger that has same mode, and use newly created logger to overwrite current one.

minecrafter:
If the loop finds the logger, it will try to use it. However, as the 
assignment is swapped, it continues as if the logger did not exist at all.

Unknwon:
I don't understand, `isExist = true` marks exists or not.

minecrafter:
True, but the desired logger will not get found because of the swapped 
assignment.

See 
http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#range_val_update 
for what I mean by this.

Unknwon:
I don't get it, why can't be found?

Only one instance of same mode could exist.

minecrafter:
I have already detailed the reason for you. It's why 
https://play.golang.org/p/0sPrIN9Khw produces the zero value, and 
https://play.golang.org/p/Q_s-lhL8mv doesn't.

Unknwon:
1. `logger` is a pointer to `Logger` strcut: https://github.com/minecrafter/gogs/blob/fix-logging-adapter-changes/modules/log/log.go#L156 Use string as example is invalid.
2. What did in the so-called "working" example, you did nothing but recreate the old object.

minecrafter:
1. We are still looking for an object in a slice to modify. The example
    is valid.
 2. Look at line 10 on both examples. Then compare with this patch.

Unknwon:
> 1. We are still looking for an object in a slice to modify. The example
    is valid.

No, string is copied by value which creates another string. Pointer however, copies the its value which is the address of the struct, and still points to the same object.

minecrafter:
The fact that strings are copied by value does not affect anything in 
the example. Everything in that example is meant to allude to the point.

Unknwon:
I think you do not understand what this loop is trying to do.

1. Create a new `logger`.
2. Match existing loggers with same mode, and use new `logger` to overwrite existing one.
3. Do whatever need to do to the new `logger`.

minecrafter:
The problem is that due to a bug (which this fixes), step 2 does not work correctly. This fixes step 2.

Unknwon:
What bug? What step2 did is ignore new buflen from new logger.

Unknwon:
Teach how to use Go is out of scope of this thread. Closing and we can continue on https://discuss.gogs.io/

PR #2804, opened 2016-03-11
fix create time in releases

allonsy:
For releases created with the web portal, the time is not saved correctly (it doesn't seem to have been saved at all). The attached screenshot shows what was happening before

Let me know what you think!

This PR fixes some formatting issues in release.go and saves the time created for tags
releasessave


Unknwon:
Thanks your PR!

`CreatedUnix` is handled automatically here: https://github.com/allonsy/gogs/blob/develop/models/release.go#L40

So you only need to fix it by doing something here: https://github.com/allonsy/gogs/blob/develop/models/issue.go#L1292

allonsy:
I'm not sure exactly what you mean. Do you mean that we don't need to set createdUnix time in `routers/repo/release.go`?
If that is the case, then I agree, it seems like created Unix is indeed populated in the beforeInsert functions so we only need to set the `created` field in `routers/repo/release.go` file (my newest commit reflects that)

Unknwon:
I've pushed a change to close this PR.

allonsy:
Ah, I see
I've tested your changes and everything seems to be working

Unknwon:
:smile:

PR #2810, opened 2016-03-11
Update uses of updated and created

maxlazio:
Using mysql/5.7.11 and a new installation from deploy branch 5267dce21023802818b0c0c16ba2ca47c146e1b5 , various pages show error page 500.

Navigating to the dashboard will give:

```
.routers/user/home.go:123 Dashboard()] [E] GetRepositories: Error 1054: Unknown column 'updated' in 'order clause'
```

To reproduce:

1. Pull the latest develop branch (5267dce21023802818b0c0c16ba2ca47c146e1b5)
1. Create a new database
1. Install, configure gogs, login as user
1.  Try to navigate to the dashboard

I've investigated a bit and found a change was introduced in ad513a20e939691828ba415c9a565e8ff3daa95f that migrated all `updated` and `created` fields to `updated_unix` and `created_unix`.

Checking out a commit before ad513a, gogs will work without issues.

On clean databases the original fields (updated, created) do not exist.
Renaming the fields makes the page load.

I've read the contributing guide but please let me know if I missed something.

### DIFF
--- a/models/access.go
+++ b/models/access.go
@@ -110,7 +110,7 @@ func (u *User) GetAccessibleRepositories() ([]*Repository, error) {
 		repoIDs = append(repoIDs, access.RepoID)
 	}
 	repos := make([]*Repository, 0, len(repoIDs))
-	return repos, x.Where("owner_id != ?", u.Id).In("id", repoIDs).Desc("updated").Find(&repos)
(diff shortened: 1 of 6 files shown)


Unknwon:
Thanks a lot!

PR #2816, opened 2016-03-11
Add Japanese Fonts

MATTENN:
By adding a Japanese font in the less file, easier to see the display in Japanese.


Unknwon:
Thanks!

Issue #1149, opened 2015-03-25
Directory/Folder/Path jumping


flinkgutt:
To quickly navigate a repository it would be nice to be able to skip directories with only a single subdirectory. 
Think Java package/directory structure. i.e. src/main/java/com/example would be a single link to click rather than main/java/com/example being individual clicks.

Github has this feature as explained (and visualized) better here. 
 [https://github.com/blog/1877-folder-jumping](https://github.com/blog/1877-folder-jumping)

Unknwon:
Hi, thanks your feedback!

It's a great feature that I enjoy using it on GitHub, too.

But we cannot promise it to be available in the near future.

Armored-Dragon:
Any updates on this? It'd be great to have this, as I've recently been working on a Java project hosted with Gogs.

PR #2818, opened 2016-03-13
Folder jumping

sapk:
Simple implemation with the adding in git-module (gogits/git-module#7) for implementing  #1149

Certainly not the best implementation in  term of efficiency.

capture d ecran_2016-03-13_03-09-40

### DIFF
--- a/templates/repo/view_list.tmpl
+++ b/templates/repo/view_list.tmpl
@@ -36,15 +36,10 @@
 							{{$entry.Name}} @ {{ShortSha $commit.RefId}}
 						{{end}}
 					</td>
-				{{else if $entry.IsJumpableDir}}
-					<td class="name">
-						<span class="icon octicon octicon-file-{{if or $entry.IsDir}}directory{{else}}text{{end}}"></span>
-						<a href="{{EscapePound $.BranchLink}}/{{EscapePound $.TreePath}}{{EscapePound $entry.JumpablePath}}{{EscapePound $entry.JumpableName}}"><span style="color:#ccc">{{$entry.JumpablePath}}</span>{{$entry.JumpableName}}</a>
-					</td>
 				{{else}}
 					<td class="name">
-						<span class="icon octicon octicon-file-{{if or $entry.IsDir}}directory{{else}}text{{end}}"></span>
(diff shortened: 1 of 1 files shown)


sapk:
Maybe a better choice could be to edit the TreeEntry struct and a JumpedFolder attribute relink the treeentry to the pointing file. In this case the treeEntry would reflect the real pointed file and not the folder. It could also limite the calculation to one time.

saboya:
I tested this in my local install, latest master + this merge. It worked fine, but the performance impact was huge, loading times went up A LOT.

sapk:
I rebase code since it was base on old version and i did some measures. (done with sqlite for db and gogs as repo with some folders to jump)

Before Rebase :
Gogs Version: 0.9.1.0306 Page: 740ms Template: 10ms
Gogs Version (with PR): 0.9.1.0306 Page: 1119ms Template: 14ms

After rebase :
Gogs Version: 0.9.12.0316 Page: 725ms Template: 4ms
Gogs Version (with PR): 0.9.12.0316 Page: 1011ms Template: 4ms

xor-gate:
Very nice :+1:

mhartkorn:
I find the performance impact of the current solution very high. On a small Java project (3 MB code) the response times went from 270ms to 4200ms. Java projects often have a lot of empty folders because of the package names and this sums up to a lot of loops in the code.

xor-gate:
@mhartkorn you are right on this, gogs would benefit from caching at a few places. On our KVM hosted gogs even loading the mainpage takes ~1 second for rendering.

mhartkorn:
@xor-gate In theory gogs supports redis and memcached but they don't seem to be used at all. But even if gogs was able to use redis for folder structures, a 4 second page load for every single folder in a repository for people that don't use redis is too long to be considered useful in my opinion. Maybe this feature can be an option.

xor-gate:
People could benefit if in-memory cache works for this long operations. IMHO 4 seconds is crazy for a repo of 3 megabyte. Seems a little performance regression here probably. Even when the disk cache is not warmed up it should not take 4 seconds.

Unknwon:
Just a note Gogs has unified cache interface, it does not matter you use in-memory, redis or memcache.

PR #2823, opened 2016-03-14
Prevent `Forgot Password` for non local users

### LINKED ISSUE #2811: LDAP User can change his/her password.
- Gogs version (or commit ref): 0.9.0
- Git version: 1.8
- Operating system: GNU/Linux
- Database:
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
  - [ ] Yes
  - [ ] No
  - [x] Not relevant
- Log gist:
No log entry describes the problem

## Description
LDAP Users who click on "Forgot Password" are able to reset his/her password. However, this is not the desired or consistent behavior. "Forgot Password" should identify the login source, and display an error to the user if the password cannot be changed.

zacheryph:
This resolves #2811.

One question I have is should I update bindata.go or does this get updated 'automatically' when needed?

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -162,6 +162,7 @@ reset_password = Reset Your Password
 invalid_code = Sorry, your confirmation code has expired or not valid.
 reset_password_helper = Click here to reset your password
 password_too_short = Password length cannot be less then 6.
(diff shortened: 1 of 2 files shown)


Unknwon:
> does this get updated 'automatically' when needed?

I'll do it.

Unknwon:
Thanks!

Issue #2670, opened 2016-02-21
Set HTML meta values for repository


enplotz:
Hi there :)
While sharing a gogs link, I noticed that the preview had the generic "Gogs (Go Git Service) is a ..." in it. While I am sure this is nice advertisement :) I would like to propose a feature:

Could the description of the repository and author be included as the value in the HTML meta fields?

So e.g. instead of 
```html
<meta name="author" content="Gogs - Go Git Service" />
<meta name="description" content="Gogs (Go Git Service) is a painless self-hosted Git service written in Go" />
```

```html
<meta name="author" content="enplotz" />
<meta name="description" content="Cool program for foo." />
```

I realize it's just a small feature, but I find it nice nonetheless, so users can see what the shared link is really about  :)

Unknwon:
Yep, sounds good!

Unknwon:
This is claimed to be fixed by merging https://github.com/gogits/gogs/pull/2824, you can test on https://try.gogs.io/

moltam:
If a repo description is empty, you get something like this:
`<meta name="description" content="yii2 - " />` (reponame - )

I don't know how important to fix this.

Apart from this looks good to me!

Unknwon:
@moltam should be fixed by https://github.com/gogits/gogs/pull/2829

Unknwon:
Close as implemented.

PR #2824, opened 2016-03-14
Set HTML meta values corresponding to repository

odinuge:
Use the authors name, and the description of the repo in HTML meta.
Pages without repository context will display as before. This implements #2670.


Unknwon:
Thanks!

PR #2829, opened 2016-03-15
Set description meta tag correctly

odinuge:
Set the description meta tag correctly when there is no repo
description. Also use the  ability to trim trailing whitespaces,
to make the template cleaner.

This cleans up the current code, and fixes [moltam's](https://github.com/moltam) comment in #2670.

### DIFF
--- a/templates/base/head.tmpl
+++ b/templates/base/head.tmpl
@@ -3,19 +3,8 @@
 <head data-suburl="{{AppSubUrl}}">
 	<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 	<meta http-equiv="X-UA-Compatible" content="IE=edge"/>
-	<meta name="author" content="
-	{{- if .Repository -}}
-		{{.Owner.Name}}
-	{{- else -}}
-		Gogs - Go Git Service
-	{{- end}}" />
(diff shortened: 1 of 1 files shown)


templates/base/head.tmpl
```diff
@@ -3,8 +3,19 @@
 <head data-suburl="{{AppSubUrl}}">
 	<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 	<meta http-equiv="X-UA-Compatible" content="IE=edge"/>
-	<meta name="author" content="{{if .Repository }}{{.Owner.Name}}{{else}}Gogs - Go Git Service{{end}}" />
-	<meta name="description" content="{{if .Repository }}{{.Repository.Name}} - {{.Repository.Description}}{{else}}Gogs (Go Git Service) is a painless self-hosted Git service written in Go{{end}}" />
+	<meta name="author" content="
+	{{- if .Repository -}}
```
Unknwon: Please do not use template syntax only available in go1.6, Gogs has backward compatibility to go1.4.

odinuge:
Should be golang 1.4 compatible now.

Unknwon:
Thanks!

Issue #2822, opened 2016-03-14
Problems with '#' in branch name


dafes:
- Gogs version (or commit ref):  0.9.0.0306
- Operating system: debian
- GIT: 2.14
- Database:
  - MySQL
- Can you reproduce the bug at http://try.gogs.io:
  - [X ] Yes

Hi,
with an '#' in the Tag name a diff and Pull-Requests creation are not possible.
You only receive a Page Not Found 404
Example:
https://try.gogs.io/dafes/test123/compare/test#11...master

Unknwon:
This is claimed to be fixed by merging https://github.com/gogits/gogs/pull/2830, you can test on the demo site.

Unknwon:
Close as fixed.

PR #2830, opened 2016-03-15
Fix problems with '#' in branchname

odinuge:
Add proper escaping of '#' in branchname in compare when doing pull
requests. This addresses issue #2822.


Unknwon:
Thanks!

Issue #2814, opened 2016-03-11
Allow searching for users by full name


richmahn:
With the search on the explore/users and admin users page, it would be nice to also be able to search the full_name of the user. Thus the search would search for the username OR the full_name.

stonerl:
#2792

richmahn:
@stonerl: This is a little different. This is regarding the new search field at the top of expore/users and the users page in the admin tools, not when adding a user to a team.

stonerl:
@richmahn I know, but I think that this feature should not only be available in certain parts of gogs but in general. Therefore I referenced the issue I opened, because both have the request for full name search in common.

Unknwon:
This is claimed to be implemented by merging https://github.com/gogits/gogs/pull/2831, you can test on demo site.

richmahn:
This is not working for me...I just updated from develop, did a rebuild...restarted server...still nothing: http://test.door43.org:3000/explore/users?q=Admini (yet if you remove the last 'i' it finds admin)

richmahn:
hmm, but it does work on try.gogs.io. Maybe doesn't work with Postgres? Will troubleshoot.

moltam:
I think it is PostgreSQL specific, the search is case sensitive: `SELECT count(*) FROM user WHERE ((lower_name like ?) OR (full_name like ?)) AND (type = ?)`.

Because the keyword is [transformed into lower case](https://github.com/odinuge/gogs/blob/6ccb2d36cfbe030cb1a72263d6889e79d738df3f/models/user.go#L1131) the search only finds lower case names with PostreSQL.

odinuge:
That could be the case... I dont have PostgreSQL accessible right now, so i can't test.

Here is a patch you can try: [search-sql-case](https://github.com/odinuge/gogs/tree/search-sql-case)

Unknwon:
Looks like a workaround: http://stackoverflow.com/questions/7005302/postgresql-how-to-make-not-case-sensitive-queries

moltam:
It works with `LOWER(full_name) like ?`.

Unknwon:
Not sure... have not tested...

moltam:
@Unknwon I tested it, and it does :)

richmahn:
Is it also calling LOWER() on the search term?

moltam:
Search term is already in lower case for lower_name search. No need for calling LOWER on it.

richmahn:
[...]

Issue #2792, opened 2016-03-09
Fullname search


stonerl:
- Gogs version (or commit ref): 0.9.0.0306
- Git version:
- Operating system:
- Database:
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at http://try.gogs.io:
  - [x] Yes
  - [ ] No
  - [ ] Not relevant
- Log gist:

## Description

I would not say that this is a bug but an annoyance. When one tries to invite another user to his team, Gogs only searches for the username but not the fullname. This is quite cumbersome because we use the smtp-auth-backend  and the usernames look like this: xbnd76 or klaw01. So if one does not know the username but the fullname there is nearly no way to add a user.

Unknwon:
This is claimed to be implemented by merging https://github.com/gogits/gogs/pull/2831, you can test on demo site.

stonerl:
Looks good.

stonerl:
Closing this one.

PR #2831, opened 2016-03-15
Make user search look in username, name and email

odinuge:
Make user search function look in username (lower_name), full name
(full_name) and primary email (email). This will benefit searching after
user in "explore", admin panel and when adding new collaborators.

As i see it, there are no security issues regarding searching for the primary email and full name. Both are already a part of the users public profile.

This implements #2814, but somehow also #2792.

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -1142,7 +1142,6 @@ func SearchUserByName(opts *SearchUserOptions) (users []*User, _ int64, _ error)
 	// Append conditions
 	sess := x.Where("lower_name like ?", searchQuery).
 		Or("full_name like ?", searchQuery).
-		Or("email like ?", searchQuery).
 		And("type = ?", opts.Type)
 
 	var countSess xorm.Session
--- a/models/user.go
+++ b/models/user.go
@@ -1137,9 +1137,13 @@ func SearchUserByName(opts *SearchUserOptions) (users []*User, _ int64, _ error)
 		opts.Page = 1
 	}
 
+	searchQuery := "%" + opts.Keyword + "%"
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR!

But I don't think search email is a good idea, please remove that.

odinuge:
Want a new pull request, or just a new commit in this one?

Unknwon:
Push new commit the PR should be updated automatically.

odinuge:
Removed the possibility to search email, so now it will search only in username and full name.

Unknwon:
Can you confirm this is actually working? Because I see `Or` and `And` in the same level, the result SQL could be wrong.

You can put `Or` inside `Where`, such as `Where("xxx LIKE ? OR xxx LIKE ?", xxx).And(...`.

odinuge:
It works on my system, and should be ok from what i have read in the [godoc](https://godoc.org/github.com/go-xorm/xorm). But i'm no sql-expert, and have never worked with orm before. I'll look more into it tomorrow.

Unknwon:
Thanks!

odinuge:
After digging further in and reading the documentation, i ended up fetching the real SQL code. With the current state of develop (after this merge), the SQL code looks like this:
`SELECT count(*) FROM user WHERE ((lower_name like ?) OR (full_name like ?)) AND (type = ?)`

I think the simplicity of xorm removes some of the elegance of normal sql, but it somehow always ends up right.

richmahn:
It would be really nice to be able to search by email address in the Admin panel, but yes, not in explore. Or maybe if the user logged in is_admin for both?

PR #2832, opened 2016-03-15
Make go's build flags a variable

dankm:
This allows the user to specify build flags from the make command
line. For example to force a complete rebuild one could use:

make BUILD_FLAGS='-v -a' TAGS="sqlite"


Unknwon:
Thanks!

PR #2836, opened 2016-03-15
Fix relative links in markdown files

### LINKED ISSUE #2545: Relative Links in markdown
Relative links are a great thing. I'd like to have them in gogs too.

https://github.com/blog/1395-relative-links-in-markup-files

I've tried it like so:
`[Sample File](REPONAME/SampleFile.obj)`

it didn't show a link to it.

odinuge:
Replace spaces with "%20" in "urlPrefix", before markdon processing.
The spaces were causing blackfriday (markdown processor) to behave
strange. This fixes #2545 (hope it works this time...)

### DIFF
--- a/modules/markdown/markdown.go
+++ b/modules/markdown/markdown.go
@@ -159,6 +159,8 @@ var (
 	svgSuffixWithMark = []byte(".svg?")
 	spaceBytes        = []byte(" ")
 	spaceEncodedBytes = []byte("%20")
+	space             = " "
+	spaceEncoded      = "%20"
 )
 
 // Image defines how images should be processed to produce corresponding HTML elements.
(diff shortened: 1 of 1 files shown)


modules/markdown/markdown.go
```diff
@@ -357,6 +357,7 @@ OUTER_LOOP:
 
 // Render renders Markdown to HTML with special links.
 func Render(rawBytes []byte, urlPrefix string, metas map[string]string) []byte {
+	urlPrefix = strings.Replace(urlPrefix, string(spaceBytes), string(spaceEncodedBytes), -1)
```
Unknwon: For performance, I suggest creating `space` and `spaceEncoded` in type string.

odinuge: So you want two new variables, if i understand you right?
```golang
spaceString := " "
spaceEncodedString := "%20"
```

Unknwon: `space` and `spaceEncoded`

odinuge:
The last commit should do it.

When working on it, i saw that the implementation of links is kinda wrong. There should be _"two"_ urlPrefixes. Now there is no difference between links with a leading slash, and without. Links with a leading slash should link to the root of the repo, while links without  one should link within their directory.

Will look into it tomorrow.

Unknwon:
Thanks!

Issue #2641, opened 2016-02-17
Ability to download diff as patch


biji:
Please add feature to download as patch while viewing commit detail:
[Split View]  [Show Diff Stats] [Download as patch]

Thanks

eriknelson:
+1

Is it possible to generate a .diff or .patch via a commit today? Very useful feature available on Github and Gitlab.

jag3773:
Looks like Github accomplishes this by simply allowing you to append '.patch' to the URL, might be a good way to go if you don't want to add a UI for this.

odinuge:
Need any help on this?

Have been looking into implementing the ability to append '.patch' and '.diff' to the url. Would be great for quickly getting a  .patch file, or getting the _whole_ diff.

Unknwon:
@odinuge yes, please go ahead :smiley:

biji:
when implemented, maybe adding explicit link to .patch and .diff would be great. so this is not a hidden feature :smile:

renard:
Hi,

Will this feature be added in the next release? That would be great. Thanks.

Unknwon:
Patch has pushed to fix this issue, please test on `master` branch or https://try.gogs.io.

PR #2846, opened 2016-03-16
Add support for .diff and .patch

odinuge:
Add the ability to get text-diff and format-patch by adding .diff or
.patch in the end of a commit url. Issue #2641

With the current behavior with the wildcard in the route, this is the only solution i see. There are probably more elegant ways to solve the problem, but that will require some rethinking. What is your opinion @Unknwon?

And "maxlines", that should not have any impact on this..?

### DIFF
--- a/models/git_diff.go
+++ b/models/git_diff.go
@@ -425,3 +425,41 @@ func GetDiffRange(repoPath, beforeCommitID string, afterCommitID string, maxLine
 func GetDiffCommit(repoPath, commitId string, maxLines, maxLineCharacteres, maxFiles int) (*Diff, error) {
 	return GetDiffRange(repoPath, "", commitId, maxLines, maxLineCharacteres, maxFiles)
 }
+
+func GetRawDiff(repoPath, commitId string, diffType string) (string, error) {
+	repo, err := git.OpenRepository(repoPath)
+	if err != nil {
+		return "", err
+	}
+
+	commit, err := repo.GetCommit(commitId)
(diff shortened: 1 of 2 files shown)


Unknwon:
@odinuge 

I've added a route for you: https://github.com/gogits/gogs/commit/60ae8ac3d2995d46156ead8ad93004801daad4ce#diff-da541ef36e39344f2e12bddd91d363e2R526

odinuge:
Hi!

I did think about that, but that breaks the ability to use refs instead of sha1sum. [Here](https://try.gogs.io/odinuge/issue-2545-test/commit/master) is an example. Using regex instead of the wildcard is a good idea! I'll keep working on it, but I don't have more time today or tomorrow.

Unknwon:
That use case was never part of the design.. maybe somehow supported, but not intended to.

barracks510:
I've tested this after rebasing it to the latest develop branch and it works very nicely. Thanks!

Unknwon:
PR has conflict...

barracks510:
I've rebased the patch to the latest develop tree. You should be able to merge the PR without conflict.

odinuge:
If you think it is ready, i can rebase it to the latest develop.

Unknwon:
There are some changes on https://github.com/gogits/gogs/pull/3300 as well, you would have to update those at the same time.

odinuge:
Rebased the PR.
New PR, see #3300..

Issue #2217, opened 2015-12-16
Ability to serve HTTP requests on unix socket


kernel164:

nivan22:
For example, a mysql sever could be staging or production, but it could also be master or slave. Allowing multi-tag-- DNS perspective production.master.mysql.service.consul and master.production.mysql.service.consul are supposed to be separate hosts, but for service discovery it would be the same. A small price. That's if we really care that production.primary.mysql.service.consul and primary.production.mysql.service.consul are the same. Personally I think that's a non-issue that only affects DNS caching and frankly not !!!!

Unknwon:
@nivan22 thanks your comment.... but I don't quite get what you were trying to say..?

Unknwon:
@kernel164 Thanks your feedback!

Please clarify your title more specific, unix socket support of what

JokerQyou:
I suppose database connection or something?

Unknwon:
Database socket has supported I think.

kernel164:
Oops I missed the description. I have a requirement to make gogs listen in [unix socket](https://en.wikipedia.org/wiki/Unix_domain_socket). We have nginx to handle ssl and gogs would be served @ /gogs/ subpath.

We were able to make gogs 0.8.0 work with the following config in nginx.
```
    location /gogs/ {
        rewrite ^/gogs/(.*)$ /$1 break;
        proxy_set_header  X-Real-IP  $remote_addr;
        proxy_pass http://localhost:3000;
        proxy_redirect http://localhost:3000/ $scheme://$host/gogs/;
        access_log /var/log/nginx/gogs.access.log;
        error_log /var/log/nginx/gogs.error.log;
    }
```

We need ```proxy_redirect http://unix:/var/run/gogs.sock; ``` support for better security.

Unknwon:
Thanks for following up!

> We need proxy_pass http://unix:/var/run/gogs.sock; support for better security.

Why this have better security?

kernel164:
If we listen on localhost any one can access gogs from localhost. If we listen on unix socket we could give permission like root:nginx which would make only nginx to access it.

Unknwon:
OK... make sense.

But I currently don't know how to let Go listens in unix socket and serve HTTP requests...

kernel164:
Here is the code that gin uses for unix socket support - https://github.com/gin-gonic/gin/blob/52fcc5dbf6e94df33ad313858fb94b713e9d1b4a/gin.go#L254

[...]

PR #2852, opened 2016-03-17
Add initial support for unix sockets

odinuge:
Initial support for unix sockets. I have tested it on Arch Linux, but would be nice with some more testing. Right now you have to use `HTTP_ADDR` to set the location of the socket. It would definitely be better to rename it to something like `LISTEN_ADDR`, or creating a `UNIX_SOCKET_ADDR`. 

Issue #2217

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -8,6 +8,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/http/fcgi"
 	"os"
@@ -576,6 +577,9 @@ func runWeb(ctx *cli.Context) error {
 
 	var err error
 	listenAddr := fmt.Sprintf("%s:%s", setting.HttpAddr, setting.HttpPort)
+	if setting.Protocol == setting.UNIX_SOCKET {
+		listenAddr = fmt.Sprintf("%s", setting.HttpAddr)
+	}
 	log.Info("Listen: %v://%s%s", setting.Protocol, listenAddr, setting.AppSubUrl)
 	switch setting.Protocol {
 	case setting.HTTP:
@@ -585,6 +589,20 @@ func runWeb(ctx *cli.Context) error {
 		err = server.ListenAndServeTLS(setting.CertFile, setting.KeyFile)
 	case setting.FCGI:
 		err = fcgi.Serve(nil, m)
+	case setting.UNIX_SOCKET:
+		os.Remove(listenAddr)
+		listener, err := net.ListenUnix("unix", &net.UnixAddr{listenAddr, "unix"})
+		if err != nil {
+			break
+		}
+		// FIXME add proper implementation of signal capture on all protocols
+		// execute this on SIGTERM or SIGINT: listener.Close()
+		err = os.Chmod(listenAddr, os.FileMode(setting.UnixSocketPermission))
+		if err != nil {
+			log.Fatal(4, "Failed to set permission of unix socket: %v", err)
+		}
+		err = http.Serve(listener, m)
+
 	default:
 		log.Fatal(4, "Invalid protocol: %s", setting.Protocol)
 	}
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -57,6 +57,8 @@ DOMAIN = localhost
 ROOT_URL = %(PROTOCOL)s://%(DOMAIN)s:%(HTTP_PORT)s/
 HTTP_ADDR = 0.0.0.0
 HTTP_PORT = 3000
+; Permission for unix socket
+UNIX_SOCKET_PERMISSION = 666
 ; Local (DMZ) URL for Gogs workers (such as SSH update) accessing web service.
 ; In most cases you do not need to change the default value.
 ; Alter it only if your SSH server node is not the same as HTTP node.
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -12,6 +12,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -30,9 +31,10 @@ import (
 type Scheme string
 
 const (
-	HTTP  Scheme = "http"
-	HTTPS Scheme = "https"
-	FCGI  Scheme = "fcgi"
+	HTTP        Scheme = "http"
+	HTTPS       Scheme = "https"
+	FCGI        Scheme = "fcgi"
+	UNIX_SOCKET Scheme = "unix"
 )
 
 type LandingPage string
@@ -57,16 +59,17 @@ var (
 	AppDataPath    string
 
 	// Server settings
-	Protocol           Scheme
-	Domain             string
-	HttpAddr, HttpPort string
-	LocalURL           string
-	OfflineMode        bool
-	DisableRouterLog   bool
-	CertFile, KeyFile  string
-	StaticRootPath     string
-	EnableGzip         bool
-	LandingPageUrl     LandingPage
+	Protocol             Scheme
+	Domain               string
+	HttpAddr, HttpPort   string
+	LocalURL             string
+	OfflineMode          bool
+	DisableRouterLog     bool
+	CertFile, KeyFile    string
+	StaticRootPath       string
+	EnableGzip           bool
+	LandingPageUrl       LandingPage
+	UnixSocketPermission uint32
 
 	SSH struct {
 		Disabled            bool           `ini:"DISABLE_SSH"`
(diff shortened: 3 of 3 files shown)


cmd/web.go
```diff
@@ -577,6 +581,18 @@ func runWeb(ctx *cli.Context) {
...
+		os.Remove(listenAddr)
+		listener, err := net.Listen("unix", listenAddr)
+		if err != nil {
+			break
+		}
+		err = os.Chmod(listenAddr, os.FileMode(setting.UnixSocketPermission))
+		if err != nil {
+			break
+		}
+		defer listener.Close()
```
mhartkorn: Shouldn't this be called directly after line 586?

cmd/web.go
```diff
@@ -577,6 +581,18 @@ func runWeb(ctx *cli.Context) {
 		err = server.ListenAndServeTLS(setting.CertFile, setting.KeyFile)
 	case setting.FCGI:
 		err = fcgi.Serve(nil, m)
+	case setting.UNIX_SOCKET:
+		os.Remove(listenAddr)
+		listener, err := net.Listen("unix", listenAddr)
+		defer listener.Close()
```
Jonny007-MKD: Can the listener be closed when an error happened?

mhartkorn: You are right, this needs to be moved after the error check. The Go source for `net.UnixListen` is has this line in it:

```
if err != nil {
	return nil, &OpError{Op: "listen", Net: net, Source: nil, Addr: laddr.opAddr(), Err: err}
}
```

Jonny007-MKD: I've seen a panic recently because of a similar `defer Close` ;)

Jonny007-MKD:
What can I do with the permissions? My webserver is running as `www-data:www-data` whereas gogs runs as `git:git`. I assume, that thus the socket will be owned by `git:git`. Do I have to add `www-data` to the `git` group of is there a better solution?

I think, I prefer a world readable socket by default.

odinuge:
I did actually think about that, but the result was to set `660` as default. Another good approach would be to add the possibility to `chown` the file to another group, so that the socket don't have to be `666`. Adding www-data to the git group would give it access to a bit more than needed.

To change the permission of the socket, set this in your config:
`UNIX_SOCKET_PERMISSION = 0666`

But yes, you are right. As of now, no other user will be able to use the socket by default. Setting `666` as default would probably be best as of now.

Jonny007-MKD:
Is it correct that `chown` would only work if gogs was running as `root` :-1: or if `git` belongs to the target group (e.g. `www-data`)?

At my installation I will use the `0660` permission, but I think that not every user is capable of performing the neccessary changes. Is it advisable to create a new group which only `www-data` and `git` belong to and which is only used for this socket?

odinuge:
As long as the user running the program belongs to the target group, then there should be no problem. I just have to look into how to implement it in golang.

Yes, in my opinion, a group with only the gogs/git user and the webserver-user (www-data/http) will be the best solution.

And: No, listener.Close() does not delete the file. Adding a new defer that removes the file would be a great addon! Thanks!

conf/app.ini
```diff
@@ -55,6 +55,8 @@ DOMAIN = localhost
 ROOT_URL = %(PROTOCOL)s://%(DOMAIN)s:%(HTTP_PORT)s/
 HTTP_ADDR =
 HTTP_PORT = 3000
+; Permission for unix socket
+UNIX_SOCKET_PERMISSION = 0660
```
Unknwon: Can you confirm this is really working?

cmd/web.go
```diff
@@ -577,6 +581,18 @@ func runWeb(ctx *cli.Context) {
 		err = server.ListenAndServeTLS(setting.CertFile, setting.KeyFile)
 	case setting.FCGI:
 		err = fcgi.Serve(nil, m)
+	case setting.UNIX_SOCKET:
+		os.Remove(listenAddr)
+		listener, err := net.Listen("unix", listenAddr)
+		if err != nil {
+			break
+		}
+		defer listener.Close()
```
Unknwon: My bad, forgot this is in the `web.go`, so `defer` would actually never be called, you can directly call `listener.Close()` here.

cmd/web.go
```diff
@@ -577,6 +581,18 @@ func runWeb(ctx *cli.Context) {
...
+		err = os.Chmod(listenAddr, os.FileMode(setting.UnixSocketPermission))
+		if err != nil {
+			break
```
Unknwon: `break`? Die in slience?

cmd/web.go
Jonny007-MKD: if he calls `listener.Close()` here, the listener will never listen!? Do you mean he can omit the closing?

Unknwon: Oh, yes, you're right.

cmd/web.go
mhartkorn: Doesn't the `break` make the `switch` fall-through to line 600?

conf/app.ini
odinuge: Can confirm its not working... go-ini removes the leading zero, making it decimal 660, instead of octal 660

odinuge:
After some investigation, it appears that if gogs is killed, `listener.close()` isn't executed. That means the socket is not deleted after exit. That is not a major problem, because the socket is deleted at startup.

Anyway, it is nice to do cleanup. The easiest way to do it is to listen for SIGTERM, and delete the file before shutdown. 

@Unknwon, do you prefer a SIGTERM listener, or to leave the socket file in the filesystem?

conf/app.ini
Unknwon: Try:

```
`0660`
```

Unknwon:
@odinuge I see, maybe it's good to add a `FIXME: xxx` comments say that we should capture SIGTERM for better handling.

conf/app.ini
odinuge: I'll start using 666, without the octal zero. And convert it to int from string.

mhartkorn: @odinuge People might get confused by this. Permissions are normally notated as octal.

odinuge:
Turns out that group lookup in os/user (so that gogs can set owner group of socket) was implemented in January, thus not available before golang 1.7. That means gogs have to execute the shell command `chown user:group filename`to set owner. What do you think @Unknwon? It is just as easy to manually adding a folder for the socket, with the correct owner.

Unknwon:
Somehow I think you're going too far from the thing we actually need.

What is the problem right now?

cmd/web.go
```diff
@@ -577,6 +583,26 @@ func runWeb(ctx *cli.Context) {
...
 		err = fcgi.Serve(nil, m)
+	case setting.UNIX_SOCKET:
+		os.Remove(listenAddr)
+		listener, err := net.ListenUnix("unix", &net.UnixAddr{listenAddr, "unix"})
+		if err != nil {
+			break
+		}
+
+		// FIXME add proper implementation of signal capture on all protocols
+		sigc := make(chan os.Signal, 1)
```
Unknwon: Sorry I did not say it clear, we do not catch SIGTERM right now, but leave a comment (you already have) to implement in the future.

odinuge:
Here is an easy command for testing the unix socket. it binds http port 3001 to the unix socket.
`socat TCP-LISTEN:3001,reuseaddr,fork UNIX-CLIENT:gogs-socket.sock`

Do you want a complete rebase, to get one clean implementation commit?

Unknwon:
> Do you want a complete rebase, to get one clean implementation commit?

Yes, please do that.

cmd/web.go
```diff
@@ -577,6 +581,20 @@ func runWeb(ctx *cli.Context) {
...
+		if err != nil {
+			break
+		}
+		// FIXME add proper implementation of signal capture on all protocols
+		// execute this on SIGTERM or SIGINT: listener.Close()
+		err = os.Chmod(listenAddr, os.FileMode(setting.UnixSocketPermission))
+		if err != nil {
+			log.Fatal(4, "Failed to set permission of unix socket: %v", err)
+		}
+		http.Serve(listener, m)
```
Unknwon: Forgot to catch error?

odinuge: What? Where did that one go...?

Unknwon: https://gowalker.org/net/http#Serve this returns an error.

odinuge: Yes, i know. I just don't understand why i have removed it, because it was in the initial commit..

odinuge:
@Unknwon, any news on this?

Unknwon:
Please sign the CLA.

Unknwon:
Thanks!

PR #2883, opened 2016-03-24
Fixes #2842 Parses URL through EscapePound template function

### LINKED ISSUE #2842: raw link 'broken' when filename contains spaces
- Gogs version (or commit ref): 0.9.0.0306
- Git version: any
- Operating system: Server -> Archlinux, Debian // Client -> Archlinux, Windows
- Database:
  - [ ] PostgreSQL
  - [ ] MySQL
  - [X] SQLite
  - [X] Yes  -> https://try.gogs.io/scnet/linkTest
  - [ ] No
  - [ ] Not relevant
- Log gist:

## Description

When adding a binary file with a name containing spaces, the RAW link generates a wrong filename, e.g. EMPTY.zip for "EMPTY with spaces.pptx"

Also the mimetype appears to be zip for office files.

https://try.gogs.io/scnet/linkTest

nikkomiu:

### DIFF
--- a/modules/template/template.go
+++ b/modules/template/template.go
@@ -97,7 +97,7 @@ func NewFuncMap() []template.FuncMap {
 		"ActionContent2Commits": ActionContent2Commits,
 		"ToUtf8":                ToUtf8,
 		"EscapePound": func(str string) string {
-			return strings.Replace(strings.Replace(str, "%", "%25", -1), "#", "%23", -1)
+			return strings.NewReplacer("%", "%25", "#", "%23", " ", "%20").Replace(str)
 		},
 		"RenderCommitMessage": RenderCommitMessage,
 		"ThemeColorMetaTag": func() string {


modules/template/template.go
```diff
@@ -97,7 +98,12 @@ func NewFuncMap() []template.FuncMap {
 		"ActionContent2Commits": ActionContent2Commits,
 		"ToUtf8":                ToUtf8,
 		"EscapePound": func(str string) string {
-			return strings.Replace(strings.Replace(str, "%", "%25", -1), "#", "%23", -1)
+			str = strings.Replace(strings.Replace(str, "%", "%25", -1), "#", "%23", -1)
+			u, err := url.Parse(str)
```
Unknwon: What does this do?

nikkomiu:
It runs the URL parser so that spaces and other special characters are properly escaped in the URL path. (Fixing Issue #2842)

nikkomiu:
If this is approved I was planning on renaming then the function because it is partially changing its definition/purpose. However, I wanted to make sure this would go through beforehand since I haven't contributed before so not sure how open to change the project is.

Unknwon:
> so that spaces and other special characters are properly escaped in the URL path 

If this is the case, I think `strings.Replace`s are not needed anymore.

nikkomiu:
If the string replaces are not there then the resulting URL would look like:
http://localhost:3000/nikko.miu/%2Fnikko.miu%2Ftestrepo%2Fsrc%2Fmaster/Test+Ing%23something%2522.txt

and not like:

http://localhost:3000/nikko.miu/testrepo/src/master/Test%20Ing%23something%2522%20Jackedup.txt

which is the actual URL that is needed. This appears to be caused by the number of different URL parts that get passed into this function.

Unknwon:
OK... then the `url.Parse` thing feels like not working, maybe what we need is [`strings.Replacer`](https://gowalker.org/strings#NewReplacer)

nikkomiu:
You are correct, I just tried it and that does look cleaner and seem to work better. Should I open a new Pull Request?

Unknwon:
You can rebase your branch and only submit new **1** commit.

nikkomiu:
Should be working better now.

Unknwon:
Thanks!

Issue #2889, opened 2016-03-27
Reverse Proxy Authentication Clone Password


joe-french:
- Gogs version (or commit ref): 0.9.0.0306
- Git version: 1.8.3.1
- Operating system: CentOS 7
- Database:
  - [ ] PostgreSQL
  - [ ] MySQL
  - SQLite
- Can you reproduce the bug at http://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  -  Not relevant
- Log gist: 
```
2016/03/27 00:45:47 [D] Session ID: <Session ID Redacted>
2016/03/27 00:45:47 [D] CSRF Token: <CSRF Token Redacted>
2016/03/27 00:45:48 [...routers/repo/http.go:122 HTTP()] [E] invalid token
```

## Description
I believe there's an issue with authentication when doing a git clone / git pull / etc. on a private repository using HTTP, and a reverse proxy. 

Using apache basic auth, and reverse proxy I can login and browse gogs, but I can't clone a repository. Unless I set my gogs password to match the password in the htaccess file. 

From looking at the code in routers/repo/http.go there's a comment that looks awfully suspect around line 97 https://github.com/gogits/gogs/blob/master/routers/repo/http.go#L97-L100 .

Basically I believe git clones / pulls aren't honoring the `REVERSE_PROXY_AUTHENTICATION` parameters.

BlackVoid:
You can solve it by setting an environment variable in the site config `SetEnv proxy-chain-auth On` which passes the information to Gogs.

Issue #2170, opened 2015-12-13
User authentication via reverse proxy request header for git HTTP


bonitoDeMadrid:
As of #165 it's possible to bypass the gogs authentication when working with a reverse proxy in front of gogs using a special header and setting up gogs with:

	[security]
	REVERSE_PROXY_AUTHENTICATION_USER = x-webauth-user

	[service]
	ENABLE_REVERSE_PROXY_AUTHENTICATION = true

So a typically setup would be:

	BROWSER => https://gogs.mycompany.com [reverse proxy] (doing the authentication magically) => gogs (some intranet machine)

This works pretty well, when the user works with the frontend of gogs using a browser, cauz he doesn't bother with  authentication at all. 

Nevertheless, working with git over http(s) in this setup:

	GIT client => https://gogs.mycompany.com [reverse proxy] => gogs

works only _sort of_ cauz the user is asked for (the gogs account) credentials.
But in this setup, the user doesn't know the credentials at all, cauz he never registered an gogs account; the reverse proxy doing the whole authentication magic. 

	Expected:
	git init 
	git remote add origin https://gogs.mycomany.com 
	git fetch
	' It work's withouth asking for user credentials, cauz the reverse proxy provided them

Why?
Mainly, cauz the reserve proxy in such setups can be the gold standard in security and are moslty fully implemented in the company intranet already. So if the reverse proxy says: "This is Bonzo; he is accessing from his office ip at midnight (yeah, let's log that incident); but anyway, he is allowed to access gogs.intranet-ip under this conditions...", then yeah, he is authenticated

So I would love to disable ssh port at all for gogs, so users can't bypass the existing security model at all. Nevertheless, letting users deal with usernames and passwords in bigger companies or raise another authentication model besides the existing one leads to unlimit service calles like "i forget my password! where can i register for gogs? what is an public key? where do i put it? it doesn't work, it doesn't work - help help help - mommy - I messed my diaper..."

cirocosta:
Hey, so this was solved at #2891  ?

PR #2890, opened 2016-03-27
Develop

joe-french:
This pull request is targeting #2889 and possibly #2170 

The code in the pull request:

Added reverse proxy auth to HTTP push/pull
    
Added check of reverse proxy authentication options to allow for pushing / pulling of repositories using reverse proxy authentication.
    
This implementation likely isn't the best, as it seems like there needs to be some single sign on method that aligns with the code in modules/auth/auth.go.


joe-french:
Made PR by mistake. Not to used to this github thing.

Issue #2692, opened 2016-02-23
Clone fails (401 Unauthorized)


Zamicol:
## Relevant info
Gogs version: v0.8.25
Git version: 2.7.0.windows.1
Operating system: Windows Server 2008 R2 Standard
Database: SQLite
Last line in gogs.log: Listen: http://0.0.0.0:3000/gogs

Reversed proxied through IIS 7.5

### This is what I see on the command line output server side:

[Macaron] Started GET [repo].git/info/refs?service=git-upload-pack for [IP address]:50673
2016/02/23 14:36:26 [D] Session ID: 32ceb61ccd7132c0
2016/02/23 14:36:26 [D] CSRF Token: DUpRkQ_r7kYBSRwMijEY7P3ekmk6MTQ1NjI2MzM4NjU1
MTA5MjIwMA==
[Macaron] Completed [repo].git/info/refs?service=git-upload-pack 401
 Unauthorized in 5.0003ms

### This is what I see client side: 

git clone [repoUrl].git --verbose
Cloning into '[repo]'...
fatal: unable to access '[repoUrl].git/': The requested URL returned error: 500

I'm trying to see if there is any other way to debug this.  Nothing is really showing in the logs.  

So far everything non-git works fine.  Nothing appears in the IIS logs.

Unknwon:
Hi, I think you did not config sub-path right.

Git receives 502 for bad gateway as I can think of.

Zamicol:
That might be the case.  It's a 500 error, not 502.   Server side it is saying "401 Unauthorized".

It works fine when not using the proxy.

Unknwon:
> That might be the case. It's a 500 error, not 502. Server side it is saying "401 Unauthorized".

It works fine when not using the proxy.

I think your proxy is not sending basic auth header to Gogs?

Zamicol:
I think you are probably right.  

The logs appear sparse on information to pinpoint the problem.  Any suggestions?

Unknwon:
Maybe this helps? http://stackoverflow.com/questions/6264878/no-basic-authentication-choice-available-in-iis7

Unknwon:
ping @Zamicol

Zamicol:
Thanks Unknwon, 

Basic Authentication is enabled.  I've looked through every IIS option I know of and nothing appears in the IIS logs.  Not sure how to debug from here.

Unknwon:
[...]

PR #2891, opened 2016-03-27
Added reverse proxy auth to HTTP push/pull

joe-french:
Added check of reverse proxy authentication options to allow for
pushing / pulling of repositories using reverse proxy authentication.

This implementation likely isn't the best, as it seems like there needs to
be some single sign on method that aligns with the code in
modules/auth/auth.go.

### DIFF
--- a/routers/repo/http.go
+++ b/routers/repo/http.go
@@ -86,53 +86,67 @@ func HTTP(ctx *context.Context) {
 	)
 
 	// check access
+	// currently check basic auth
+	// TODO: support digit auth
+	// FIXME: Maybe should move all auth into one unified auth call.
+	// maybe could use that one.
 	if askAuth {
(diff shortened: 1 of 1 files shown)


joe-french:
I made this PR regarding the #2889 that I created, but it may resolve #2170, and the issue I was running into sounds a lot like #2692 .

Zamicol:
Thanks @joe-french for your work.  

I compiled joe-french:develop and it didn't fix the issues from #2692.

Unknwon:
Please solve the conflict...

dl5rcw:
After getting gogs + reverse proxy (apache 2.4) to work, I enabled the auto create user function. Works fine with basic auth now. Starting to merge my svn, I ran into the clone/pull/push conflict. Seems like public repos can be clones, but not pull/push. 
This seems to be a real issue as all those related issues are mentioned here.

Issue #2892, opened 2016-03-27
Insecure database queries


joepie91:
The Gogs codebase seems to be very inconsistent in how it generates queries. There are entire [chunks of code](https://github.com/gogits/gogs/blob/master/models/issue.go#L801-L842) where parameterized queries are used alongside string-concatenated queries, and this [has led to SQL injection vulnerabilities in the past](https://www.cvedetails.com/vulnerability-list/vendor_id-14955/product_id-30369/opsqli-1/Gogits-Gogs.html).

Is there any particular reason why Gogs isn't using parameterized queries *everywhere*? Given that all major databases now support them, it's kind of shocking to me that this isn't already the case, and makes me rather concerned about deploying Gogs on my system, security-wise.

andreynering:
You are right.

I'd say: PRs are welcome if you are willing to help.

joepie91:
Unfortunately, I don't speak Go, otherwise I definitely would've sent in PRs.

It does seem like something that really needs to be prioritized, though - it's relatively little effort to rule out a very large attack surface, as it'd be just a change of syntax for the most part.

tboerger:
Do you have any insecure example?

elithrar:
The link in the OP to [`issue.go#L805`](https://github.com/gogits/gogs/blob/master/models/issue.go#L805-L817) provides a good example:

```go
	queryStr := "SELECT COUNT(*) FROM `issue` "
	if opts.LabelID > 0 {
		queryStr += "INNER JOIN `issue_label` ON `issue`.id=`issue_label`.issue_id AND `issue_label`.label_id=" + com.ToStr(opts.LabelID)
	}

	baseCond := " WHERE issue.repo_id=" + com.ToStr(opts.RepoID) + " AND issue.is_closed=?"
	if opts.MilestoneID > 0 {
		baseCond += " AND issue.milestone_id=" + com.ToStr(opts.MilestoneID)
	}
	if opts.AssigneeID > 0 {
		baseCond += " AND assignee_id=" + com.ToStr(opts.AssigneeID)
	}
	baseCond += " AND issue.is_pull=?"
```

`com.ToStr` naively converts any type to a string, and the code here is just concatenating string - e.g.

```
baseCond := " WHERE issue.repo_id=" + com.ToStr(opts.RepoID) + " AND issue.is_closed=?"
```

This (and any query like it) needs to be parameterized.

tboerger:
[...]

PR #2893, opened 2016-03-27
Try to make the SQL queries cleaner and more secure

tboerger:
It's a start as a followup to https://github.com/gogits/gogs/issues/2892

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -5,7 +5,6 @@
 package models
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -513,7 +512,7 @@ func Issues(opts *IssuesOptions) ([]*Issue, error) {
 		if len(opts.RepoIDs) == 0 {
 			return make([]*Issue, 0), nil
 		}
-		sess.Where("issue.repo_id IN ("+strings.Join(base.Int64sToStrings(opts.RepoIDs), ",")+")").And("issue.is_closed=?", opts.IsClosed)
+		sess.In("issue.repo_id", base.Int64sToStrings(opts.RepoIDs)).And("issue.is_closed=?", opts.IsClosed)
 	} else {
 		sess.Where("issue.is_closed=?", opts.IsClosed)
 	}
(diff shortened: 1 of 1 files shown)


models/issue.go
```diff
@@ -684,18 +683,8 @@ func GetIssueUserPairsByRepoIds(rids []int64, isClosed bool, page int) ([]*Issue
...
-		buf.WriteString(com.ToStr(rid))
-		buf.WriteString(" OR ")
-	}
-	cond := strings.TrimSuffix(buf.String(), " OR ")
 	ius := make([]*IssueUser, 0, 10)
-	sess := x.Limit(20, (page-1)*20).Where("is_closed=?", isClosed)
-	if len(cond) > 0 {
-		sess.And(cond)
-	}
+	sess := x.Limit(20, (page-1)*20).Where("is_closed=?", isClosed).In("repo_id", rids)
```
andreynering: More personal preference than anything else, but I like to have each sentence in each line:

```go
sess := x.
    Limit(20, (page-1)*20).
    Where("is_closed = ?", isClosed).
    In("repo_id", rids)
```

Apart from that, :+1:

tboerger: As @Unknwon have not done that so far I won't change it like that :)

tboerger:
@Unknwon if you are fine with that you can already merge that. If I can improve also other parts I will create more followup pull requests.

models/issue.go
Unknwon: Just to verify, what happens if `rids` is empty?

tboerger: > Just to verify, what happens if rids is empty?

https://github.com/go-xorm/xorm/blob/master/statement.go#L743-L745

models/issue.go
```diff
@@ -548,27 +547,16 @@ func Issues(opts *IssuesOptions) ([]*Issue, error) {
...
+		sess.Join("INNER", "issue_label", "issue.id = issue_label.issue_id").In("issue_label.label_id", labelIDs)
 	}
 
 	if opts.IsMention {
-		queryStr := "issue.id=issue_user.issue_id AND issue_user.is_mentioned=1"
+		sess.Join("INNER", "issue_user", "issue.id = issue_user.issue_id AND issue_user.is_mentioned = 1")
+
 		if opts.UserID > 0 {
-			queryStr += " AND issue_user.uid=" + com.ToStr(opts.UserID)
+			sess.Where("issue_user.uid = ?", opts.UserID)
```
andreynering: Shouldn't this be?:

```go
sess = sess.Where("issue_user.uid = ?", opts.UserID)
```

At least is how it would work on Gorm.

models/issue.go
tboerger: > Just to verify, what happens if rids is empty?

Look above within the function, you have done already a check, e.g. `if len(rids) == 0 {`

models/issue.go
tboerger: As you can see on all the query updates above this seems to be not required.

Unknwon:
Thanks!

models/issue.go
```diff
@@ -548,27 +547,16 @@ func Issues(opts *IssuesOptions) ([]*Issue, error) {
...
-		if validJoin {
-			sess.Join("INNER", "issue_label", queryStr)
-		}
+	if len(labelIDs) > 1 {
+		sess.Join("INNER", "issue_label", "issue.id = issue_label.issue_id").In("issue_label.label_id", labelIDs)
 	}
 
 	if opts.IsMention {
-		queryStr := "issue.id=issue_user.issue_id AND issue_user.is_mentioned=1"
+		sess.Join("INNER", "issue_user", "issue.id = issue_user.issue_id AND issue_user.is_mentioned = 1")
```
Unknwon: Why join here as it did not?

models/issue.go
Unknwon: You may need to change to `sess.Add`.

PR #2898, opened 2016-03-28
Set utf-8 charset header for text files when serving raw content

### LINKED ISSUE #2888: Raw Code View UTF8 Problem
## Description
My code is filled with turkish comments and when i tried to view the code via clicking the 'raw' button, characters are garbled. No problem on default viewing though. Thank you...
adsiz

moltam:
Fixes #2888

### DIFF
--- a/routers/repo/download.go
+++ b/routers/repo/download.go
@@ -29,7 +29,7 @@ func ServeData(ctx *context.Context, name string, reader io.Reader) error {
 			ctx.Resp.Header().Set("Content-Transfer-Encoding", "binary")
 		}
 	} else {
-		ctx.Resp.Header().Set("Content-Type", "text/plain")
+		ctx.Resp.Header().Set("Content-Type", "text/plain; charset=utf-8")
 	}
 	ctx.Resp.Write(buf)
 	_, err := io.Copy(ctx.Resp, reader)


Unknwon:
What happens if the encoding is not `utf-8`? I think it's not an ideal solution...?

moltam:
For ASCII it causes no problem. For others I'm not sure..
I checked with ISO-8859-2, and it displays incorrectly with utf-8 header.

I think most files are encoded with utf8, so this may be not a big problem.

But github doesn't seem to detect the proper encoding either, for example:
https://raw.githubusercontent.com/moltam/test/master/iso-8859-2.txt

I guess they go with fixed utf-8 header for text files raw content.

We could detect encoding, but the `buf` contains only the first 1000 bytes.


Others:
I may have found a bug. Gogs panics with non-uf8 files, for example: https://try.gogs.io/moltam/utf8-test/src/master/iso-8859-2.txt
Log from my instance: http://pastebin.com/G53tqzvw
Should I open an Issue?

wqweto:
Commit 8be0bb177b3f28b5cb0afb87bb18db208a11f67d reverted this
```
	if isTextFile {
		charset, _ := base.DetectEncoding(buf)
		if charset != "utf-8" {
			ctx.Resp.Header().Set("Content-Type", "text/plain; charset="+charset)
		}
```
which if the way to go IMO.

jekriske:
This utf-8 issue got me today...  can't wait to try out the fix.

Unknwon:
My suggest would be detect encoding first, if it's UTF-8 then we set the encoding header, otherwise set nothing.

jekriske:
In the Unix world most things are UTF-8, git will typically default to UTF-8 as well... I'd say err on the side of UTF-8 instead of defaulting to nothing. Even Github, for example, defaults to Content-Type:text/plain; charset=utf-8

Unknwon:
@jekriske OK, make sense.

moltam:
Thanks!

PR #2300, opened 2015-12-28
typo fix

zhuharev:

PR #2908, opened 2016-03-29
Update README.md

dharamgollapudi:
Fixes a typo


andreynering:
An PR for this was already opened in the past (https://github.com/gogits/gogs/pull/2300), but it is correct as it is. It's in Portuguese.

PR #2911, opened 2016-03-29
add docker-compose v2

ArchiFleKs:
Hi, I just did a docker-compose template which is up to date with the latest docker-compose best practices. I thought it might be useful.

### DIFF
--- a/docker/contrib/docker-compose/README.md
+++ b/docker/contrib/docker-compose/README.md
@@ -1,0 +1,43 @@
+# Docker Compose
+
+[Gogs](https://github.com/gogits/gogs) : Go Git Service
+
+Docker Compose v2 (Docker 1.10) with mysql as database
+
+Place docker-compose.yaml in  `~/docker-compose/gogs/docker-compose.yaml` for exemple:
+
+In the `gogs` folder, bring up the stack :
+
+```
+# docker-compose up -d
+Creating network "gogs_default" with the default driver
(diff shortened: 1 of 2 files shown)


Unknwon:
cc @0rax

0rax:
Hi,

The idea of having a Docker-compose file is not bad by itself, but as @Unknwon mentionned there is so much different docker-compose configuration possible for gogs.

Right now providing only a simple deployement scheme using docker-compose does not seem that interesting and will just add another things that we need to maintain for it. I totally understand that some wants to use docker-compose to deploy gogs but as of right now, as gogs need to be configured using the installer or by providing a configuration file, it does not seem sensible to add one.

This could be interesting in the future when we will be able to configure gogs using environment variable, this way we can juste document those environment variable and provide a sensible default for those in a docker-compose file creating a basic deployement scheme that can be followed either to test it locally or to deploy it. Until we leverage the need for the installer / a specific configuration file for the compose run to be reproductible (each docker-compose up -d will provide a completly configured gogs instance), it feels like it will just add a layer of confusion to the deployement on docker.

Thanks for your work though !

rothgar:
Commenting to follow this and cross reference my original compose file (older syntax) #2274

Unknwon:
I have updated wiki to explicitly state currently we do not accept Docker compose file: https://github.com/gogits/gogs/wiki/Contributing-Code

PR #2914, opened 2016-03-30
Reduce line number creation to one DOM manipulation

psychomario:
I noticed that there was a large delay when loading a file in the web interface before the line numbers displayed. This PR does not remove the delay, but it does improve it slightly.

This PR reduces the on load event handler time from ~700-900ms to ~70-90ms.

To verify my timings:
Use the Chrome developer tools "Timeline", reload the page. Mouseover the 'load' Event to see the timing.

Note that I am seeing a lot of fluctuation in the timings between refresh, but my changes are consistently <100ms, and the original is consistently >700ms

### DIFF
--- a/public/js/gogs.js
+++ b/public/js/gogs.js
@@ -1043,10 +1043,14 @@ $(window).load(function () {
         var $num_list = $('.code-view .lines-num');
 
         // Building blocks.
+        var $toappendblock = [];
+        var $toappendnum_list = [];
         for (var i = 0; i < lines.length; i++) {
-            $block.append('<li class="L' + (i + 1) + '" rel="L' + (i + 1) + '">' + lines[i] + '</li>');
-            $num_list.append('<span id="L' + (i + 1) + '">' + (i + 1) + '</span>');
+            $toappendblock.push('<li class="L' + (i + 1) + '" rel="L' + (i + 1) + '">' + lines[i] + '</li>');
+            $toappendnum_list.push('<span id="L' + (i + 1) + '">' + (i + 1) + '</span>');
         }
+        $block.append($toappendblock.join(''));
+        $num_list.append($toappendnum_list.join(''));
 
         $(document).on('click', '.lines-num span', function (e) {
             var $select = $(this);


Eeems:
It would probably be a better idea to use document fragments for this.
https://davidwalsh.name/documentfragment

psychomario:
I gave `documentFragment` a go, with the following code:
```
var $toappendblock = document.createDocumentFragment();
var $toappendnum_list = document.createDocumentFragment();
for (var i = 0; i < lines.length; i++) {
    var $li = document.createElement('li');
    var $line = "L" + (i + 1);
    $li.setAttribute('class', $line);
    $li.setAttribute('rel', $line);
    $li.innerHTML = lines[i];
    $toappendblock.appendChild($li);

    var $span = document.createElement('span');
    $span.setAttribute('id', $line);
    $span.innerText = i + 1;
    $toappendnum_list.appendChild($span);

}
$block.append($toappendblock);
$num_list.append($toappendnum_list);
```

I'm getting ~3x slower (~200ms). This seems to be due to the `createElement` parsing the HTML. I'm not sure how to do this without `createElement`, as `appendChild` requires a `Node` apparently.

Eeems:
Hmm, still better then inserting it directly into the DOM. I wonder if it would be faster to create a div, append to it and then insert it's innerhtml at the end.
That said, is it faster to just append to a string instead of using arrays with a join?

psychomario:
~80ms vs ~110ms in favor of array/join. I read somewhere that it's faster because of reallocations IIRC, hence the use.

psychomario:
On a related note, is there a particular reason this needs to be done on the client side? The template is already creating the table (numbers column, code column), and Go is dumping the code on the RHS. Would it be worth doing this generation server side, and bypassing this issue, along with the reflow? I guess it wouldn't save a huge amount vs the syntax highlighting, but that would be a lot more work to implement server side.

Eeems:
Makes sense. This is assuming all JavaScript engines handle it similarly.

Unknwon:
@psychomario thanks!

For your question:

Because syntax highlighting is done on client side, server side does not go through the code content at all. But given the truth all response are going through from Git stdout to HTTP body, I think there is indeed a possible improvement to compose these `div` blocks on the fly.

Unknwon:
This is a lot faster right now: https://try.gogs.io/gogs/gogs/src/develop/models/issue.go

psychomario:
I had a quick (& naive) look at doing the numbering server side, [here: view.go](https://github.com/gogits/gogs/blob/master/routers/repo/view.go#L108-L114). I was getting a nice speedup (this was a few days ago, forgotten the numbers), but the html encoding (`< -> &lt;` etc), I think [here](https://github.com/gogits/gogs/blob/master/routers/repo/view.go#L219), renders the generated html useless (also meaning we're cutting out a little but of the html rendering client side), but I think the timing would still be an improvement.

If I get time, I'll try again at some point, but my Go skill is non-existent so it might be worthless.

Thanks for merging this PR!

PR #2920, opened 2016-04-03
Update ISC license

l2dy:
New one from Wikipedia and https://opensource.org/licenses/ISC.
I think the original one is for code owned by the Internet Systems Consortium.


Unknwon:
Thanks!

Issue #2918, opened 2016-04-02
Bug:Centos7 can't auto start


niluanxy:
On centos7,i try many ways to set auto start gogs,like write service,add gogs to rc.d,but all can't auto start gogs. But when i ssh login my server,and run gogs start,it can run.How should i do?

- Gogs version: 0.8.43
- Git version:
- Operating system: centos7 x64
- Database: MySQL
- Run user: git

jaydio:
Same issue here. As the following log entry suggests this might be because the database socket doesn't exist prior to starting gogs:

<pre>
2016/04/04 20:46:15 [...s/routers/install.go:71 GlobalInit()] [E] Fail to initialize ORM engine: migrate: sync: dial tcp 127.0.0.1:3306: getsockopt: connection refused
</pre>

By the time trying to start gogs manually the database socket already exists.

jaydio:
Some packager.io magic might be the cause. I can see systemd and sysvinit files shipped with the rpm. Need to take a closer look but here's a quick workaround. Add the following to the ```[Unit]``` section within ```/etc/systemd/system/gogs.service```:

<pre>
After=mariadb.service
</pre>

Afterwards reload your systemd configuration and reboot ...

<pre>
[root@git01 ~]# systemctl daemon-reload
[root@git01 ~]# reboot
</pre>

You'll see that gogs will come up just fine.

When installing from source one may follow the FAQ: https://gogs.io/docs/intro/faqs#how-do-i-run-gogs-at-startup-with-systemd%3F

Jonny007-MKD:
I recommend adding (additionally) `Requires=mariadb.service` as this tells systemd to not even try to start gogs when the database isn't running.

jaydio:
@Jonny007-MKD Well now, that makes perfect sense, thanks.

jaydio:
New pull request is #2935

niluanxy:
@jaydio Think you,i hava auto start.I editd gogs.service,revemo USER=git line,and add nginx after,code is:
```c
[Unit]
Description=Gogs (Go Git Service)

After=network.target
After=nginx.service
After=mysqld.service

[Service]
Type=simple
WorkingDirectory=/opt/gogs
ExecStart=/opt/gogs/gogs web
Environment=USER=git HOME=/home/git

[Install]
WantedBy=multi-user.target
```

PR #2934, opened 2016-04-06
fixed sample gogs.service to require rather than just start gogs afte… issue #2918

jaydio:
See issue #2918 for further details.


jaydio:
will submit a new pull request as I've didn't base my changes on the development branch. my bad.

PR #2939, opened 2016-04-06
Update glide.lock with latest go-gogs-client

nawa:
Need to use latest version of gogits/go-gogs-client with Team support


bkcsoft:
@Unknwon This should be merged right? Since it's messing with building (cleanly) at the moment :unamused: (Why wasn't this in the PR that broke it?? :stuck_out_tongue: )

Unknwon:
Thanks!

PR #2942, opened 2016-04-06
Render the Code view on the server (minus syntax highlighting)

psychomario:
As discussed in #2914.

The Go is probably horrible, I've never actually written with it before, so I'm _very_ open to criticism.

WRT numbers, the page doesn't seem to finish any quicker for me (still takes ~3 seconds to highlight), but everything visible bar syntax highlighting is finished much quicker, so the code doesn't jump to the right when the line numbers get done. So even though it's not faster, it feels it, and the highlighting just smoothly kicks in when it's done.

Next step would probably be to move the syntax highlighting to the server side. This would be very nice for caching etc. I found [sourcegraph/syntaxhighlight](https://github.com/sourcegraph/syntaxhighlight), but I don't want to just add a random library to this. And I wouldn't know where to start if I was going to.

Please try this out, see what it feels like, and mercilessly criticize my Go.


mhartkorn:
I don't know if it really matters but you put the content of the file into the RAM at least 4 times. Larger files could maybe result in massive slowdowns on smaller servers. Does anybody have experience with this?

psychomario:
That could be cut down (to 3?) if more HTML was generated serverside rather than in the template. i.e inject a `<tr>` once, rather than a `<td>` for the number, and a `<td>` for the line.

mhartkorn:
You could scan the input for line breaks manually and so avert to use `strings.Split`. But I don't know if it's worth it.

@Unknwon Do you have an opinion on this?

Unknwon:
I think syntax highlighting is before split into lines with line number?

psychomario:
That's correct, Syntax highlighting is at [L870-2](https://github.com/gogits/gogs/blob/master/public/js/gogs.js#L870-L872) and the split is at [L1070-5](https://github.com/gogits/gogs/blob/master/public/js/gogs.js#L1070-L1075). But the line numbers don't need syntax highlighting so I don't see why they need to be done in that order.

Unknwon:
I think because the content of line number will be treated as code to be highlighted as well.

psychomario:
Surely it would be better to not do that, only highlight the actual code and put a static style on the numbers. It might fractionally speed up page rendering.

Unknwon:
I don't think I get what you're trying to say...?

psychomario:
Why does it matter where we do the syntax highlighting? We only need to be applying it to code anyway.

Unknwon:
... that is the exact reason we do highlighting before create line numbers.

psychomario:
Unless I'm missing something though, Highlight.js only applies highlighting to content inside `<pre><code></code></pre>`, which does not include the line numbers at all, see `templates/repo/view_file.tmpl` in the diff. So doing the split and line numbering server side doesn't affect the content inside the `<pre><code></code></pre>`, so there is no change in behaviour.

Unknwon:
OK, please attach some screenshots of rendered code.

psychomario:
Apologies, it took me a while to set up my environment.

The first image is the normal code view with syntax highlighting.

Highlighted code

And the second image is the page source view, with the pre-created html.

Source View

The more interesting part, which I can't really show you in a static image, is that when you load the page the numbers are already there, so the code doesn't jump to the right when the javascript has created the numbering column.

Unknwon:
Thanks, how about click on line number and highlight the line(s) selected?

psychomario:
Turns out there was a bug because `i` was not defined. Fixed and pushed to this PR. The behaviour is the same as it should be.

Multiline select

Unknwon:
Thanks, the way you concatenate strings are very low performance, please refer to http://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go for a better approach.

psychomario:
That's not surprising as I've never used Go before. I should have chance next week to implement those changes.

Unknwon:
Please solve the conflict so I can merge and fix the performance issue.

Unknwon:
Just found an issue that `.go` file cannot be rendered correctly anymore.

Unknwon:
Could be a bug in highlightjs, waiting for 9.6 see if it would be fixed.

psychomario:
The `.go` bug was due to my code. Fixing it with the changes to string concatenation.

psychomario:
More detail on the above: The new code had no newlines in it, so highlight.js wasn't highlighting correctly as it was seeing one huge line. This was very obvious in Javascript code with `//` comments, which greyed out the rest of the code. I've now implemented proper newlines in the outputted HTML, so this issue has been resolved, and also `.go` highlighting is functional again.

I've also fixed the merge conflicts, however I'm getting a build failure:
```
# github.com/gogits/gogs/routers/api/v1/admin
routers/api/v1/admin/user.go:42: form.FullName undefined (type gogs.CreateUserOption has no field or method FullName)
```

I've not touched the mentioned file so I'm unsure what the issue is.

psychomario:
The above is fixed, turns out I'm bad at go.

New issue now with the line number highlightings interacting with the css changes I made.

psychomario:
I think this is ready now, selection highlighting has been fixed.

public/css/gogs.css
```diff
@@ -1311,6 +1311,8 @@ footer .ui.language .menu {
 .repository.file.list #file-content .code-view .lines-num .hljs li,
 .repository.file.list #file-content .code-view .lines-code .hljs li {
   padding-left: 5px;
+  display: inline-block;
```
Unknwon: What is this?

>All .less and .tmpl files must use 1 tab as indention.

psychomario: Because the code is inside `pre` and `code`, when I add the newlines so highlightjs doesn't treat it as one massive block two are rendered (one for the `\n` and one for the `li`). This essentially removes the newline from the `li` so only one is rendered and it looks right. Furthermore the `width: 100%` fixes up the `li` so it is the with of the surrounding code box rather than just the actual code text.

I'll fix up the indent now.

psychomario: It appears that the whole file is two space indented, rather than tabs. See [here](https://raw.githubusercontent.com/gogits/gogs/master/public/css/gogs.css). I don't mind either way but currently it seems to match the rest of the file.

Unknwon: https://github.com/gogits/gogs/wiki/Contributing-Code#coding-guidelines

psychomario: Apologies, I misunderstood your comment. I've moved the changes to the `.less` file.

Unknwon:
Please sign the CLA.

psychomario:
Done, although I'm not sure how to change the check below.

Unknwon:
Thanks for fast response, would you mind re-sign the CLA again, maybe because the content was adjust by me few seconds ago..

psychomario:
Re-Signed, should be all good now.

Unknwon:
@psychomario I think the problem is the email you used to commit is not added to your GitHub account, please do that to pass the check..

psychomario:
Ah, apologies, I forgot to update my git config email address. Seems to be sorted now.

Unknwon:
Tested, really awesome! 

Will improve the code after merge.

psychomario:
Thanks for merging!

I would appreciate it if you could just reply here when you've improved the code so I can have a look.

Unknwon:
@psychomario I see you already using `bytes.Buffer`, that's already faster then string concatenations. We can also only operate file content in `[]byte` level (only iterate content once without `strings.Split`, which allocates memory again for whole file content) but not sure if it's worth doing, so far you're doing good!

Issue #2293, opened 2015-12-27
Testability strategy


ayr-ton:
Hey there.

I was digging into the code and saw a excellent piece of work from all the contributors everywhere.

However, something caught my attention. The code base is growing every day but there's no Unit Tests mention at the CONTRIBUTING.md file. As Go has a good built-in support for automated tests (https://golang.org/pkg/testing/) I would like to ask if you guys have plans to design a testing strategy as something required for future contributions. 

It could increase confidence in recommending the project for production use as well, as I think gogs could replace a lot of solutions out there.

Also, I can see there's a few modules with tests https://github.com/gogits/gogs/search?p=1&q=testing&utf8=%E2%9C%93 but the coverage still small.

If you :+1: this idea, we can discuss here the strategy.

Unknwon:
Yes, unit tests can only be applied to small part of Gogs's codebase.

I'm waiting to finish APIs to achieve more on testing.

Unknwon:
Also, I suggest move discussion to http://forum.gogs.io/

Unknwon:
ping @ayr-ton ?

ayr-ton:
Pong @Unknwon 
I will be moving the discussion to the forum and then I will be posting the results here :)

ayr-ton:
Related post:
http://forum.gogs.io/topic/21/testability-strategy

tebanep:
Happened exactly the same to me, when looking for tests. Hope an strategy can be set so the confidence in the program increases. Though, I'm using gogs in production right now without problems. Thanks for it!

PR #2948, opened 2016-04-08
[Proposal of testing approach] Write tests for issue API

andreynering:
Context: https://github.com/gogits/gogs/issues/2293

I wrote few tests (just 4), because I want to know if you guys agree with the general approach. More could be easily added for other APIs (or existing tests improved).

### Approach

- We are now testing against a SQLite database (It could be PostgreSQL, or MySQL. I choose SQLite for portability)
- Fixtures are loaded in the test database before each router test
- In the future, we might need to write scripts that generate Git repositories in the `gogs-repositories` folder, in order to test code that traverse Git repos.

### DIFF
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -33,6 +33,10 @@ const (
 	HTTP  Scheme = "http"
 	HTTPS Scheme = "https"
 	FCGI  Scheme = "fcgi"
+
+	RUN_MODE_PROD = "prod"
+	RUN_MODE_DEV  = "dev"
+	RUN_MODE_TEST = "test"
 )
 
 type LandingPage string
@@ -55,6 +59,8 @@ var (
 	AppSubUrlDepth int // Number of slashes
 	AppPath        string
 	AppDataPath    string
+
+	AppRunMode string
 
 	// Server settings
 	Protocol           Scheme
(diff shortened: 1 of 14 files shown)


modules/setting/setting.go
```diff
@@ -33,6 +33,9 @@ const (
 	HTTP  Scheme = "http"
 	HTTPS Scheme = "https"
 	FCGI  Scheme = "fcgi"
+
+	RUN_MODE_PROD = "prod"
+	RUN_MODE_TEST = "test"
 )
```
mhartkorn: Did you leave out `RUN_MODE_DEV` on purpose?

andreynering: I leave it out because it wouldn't be used by now. But I think I could add it for maybe future use.

andreynering:
Tip: tests compiles a lot faster without `-cover` and `-race` flags. You probably won't want to use that for development. I took many minutes to discover why it was so slow.

andreynering:
@Unknwon Any feedback here?

Unknwon:
I think it's not a good time to spend time on integrating testing yet.

andreynering:
@Unknwon OK. I hope Gogs get a strong test suite in the future, even that a different approach would be choose.

Unknwon:
@andreynering I think Gogs should have as much as unit tests as possible, combine with integrating tests. But thoughts change fast, so I don't want people keep putting time on testing.

Issue #1496, opened 2015-08-17
Switch from HTML to plain text mails


nd2s:
Currently mails are HTML only, which is very annoying and unnecessary. Loads of clients don't render them and my spamassassin even moved them straight into the junk folder (and it was right to do so, see report below).

I think HTML mails are evil and totally not necessary in a development tool like this. So instead of fixing them I'd suggest to directly switch to plain text mails.

Spamassassin report for the record (tested on password reset mail):

	* -1.0 ALL_TRUSTED Passed through trusted hosts only via SMTP
	*  0.0 URIBL_BLOCKED ADMINISTRATOR NOTICE: The query to URIBL was blocked.
	*       See http://wiki.apache.org/spamassassin/DnsBlocklists#dnsbl-block
	*      for more information.
	*      [URIs: example.com]
	*  1.3 HTML_IMAGE_ONLY_24 BODY: HTML: images with 2000-2400 bytes of words
	*  1.1 MIME_HTML_ONLY BODY: Message only has text/html MIME parts
	*  0.0 HTML_MESSAGE BODY: HTML included in message
	*  2.0 MIME_HEADER_CTYPE_ONLY 'Content-Type' found without required MIME
	*      headers
	*  2.0 TO_NO_BRKTS_HTML_ONLY To: misformatted and HTML only

Unknwon:
Thanks your feedback!

By reading this section in your error report: http://wiki.apache.org/spamassassin/DnsBlocklists#dnsbl-block

I believe it has nothing to do with HTML thing.

But, it is good idea to add a config option whether use HTML or plain text e-mail.

nd2s:
The section you quoted was irrelevant. Spamassassin assigns points to mails. If a mail has >x points (4 in my case) it is marked spam. The DNS thing caused 0 points, this was just some info.

Relevant parts are after that: TO_NO_BRKTS_HTML_ONLY, MIME_HTML_ONLY (gogs mails have no plaintext part at all, they HAVE to have that, even if you are sending additional HTML), and so on...

Unknwon:
> Relevant parts are after that: TO_NO_BRKTS_HTML_ONLY, MIME_HTML_ONLY (gogs mails have no plaintext part at all, they HAVE to have that, even if you are sending additional HTML), and so on...

Ah... OK, I see! Sorry I've never used something similar before. :joy:

But still, I hold on opinion that add a config option about whether use HTML or plain text version e-mails.

nd2s:
[...]

PR #2954, opened 2016-04-10
Do not write HTML in text/plain mail part

### LINKED ISSUE #2928: text/plain email notifications encoded as HTML
All notifications sent (upon registration and on event) arrive as multipart containing a text/plain and a text/html content-typed elements. But both elements are the same payload: HTML.

Example:
```
Date: Tue, 05 Apr 2016 06:34:58 -0400                                           
From: gogitservice@gmail.com                                                    
To: strk@xxx.net                                                             
Subject: kalxas added you to kalxas/OSGeoLive                                   
                                                                                
<!DOCTYPE html>                                                                 
<html>                                                                          
<head>                                                                          
        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />   
        <title>kalxas added you to kalxas/OSGeoLive</title>                     
</head>                                                                         
                                                                                
<body>                                                                          
        <p>You are now a collaborator of this repository.</p>                   
        <p>                                                                     
                ---                                                             
                <br>                                                            
                View it on Gogs: <a href="https://try.gogs.io/kalxas/OSGeoLive">
kalxas/OSGeoLive</a>                                                                    </p>                                                                    
</body>                                                                         
</html>        
```

Note that this only affects MUA that are configured to prefer text/plain over text/html.
The multipart header:
```
  I     1 <no description>                     [multipa/alternativ, 7bit, 1.1K] 
  I     2 ├─><no description>                 [text/plain, quoted, utf-8, 0.4K] 
  I     3 └─><no description>                  [text/html, quoted, utf-8, 0.4K] 
```

strk:
Fixes #2928

### DIFF
--- a/modules/mailer/mailer.go
+++ b/modules/mailer/mailer.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/setting"
+	"github.com/jaytaylor/html2text"
 )
 
 type Message struct {
@@ -26,14 +27,18 @@ type Message struct {
 }
 
 // NewMessageFrom creates new mail message object with custom From header.
-func NewMessageFrom(to []string, from, subject, body string) *Message {
+func NewMessageFrom(to []string, from, subject, htmlbody string) *Message {
 	msg := gomail.NewMessage()
 	msg.SetHeader("From", from)
 	msg.SetHeader("To", to...)
 	msg.SetHeader("Subject", subject)
 	msg.SetDateHeader("Date", time.Now())
-	msg.SetBody("text/plain", body)
-	msg.AddAlternative("text/html", body)
+	msg.SetBody("text/html", htmlbody)
+	body, err := html2text.FromString(htmlbody)
+	if err != nil {
+		// TODO: report error ?
+	}
+	msg.AddAlternative("text/plain", body)
 
 	return &Message{
 		Message: msg,
--- a/modules/mailer/mailer.go
+++ b/modules/mailer/mailer.go
@@ -33,12 +33,14 @@ func NewMessageFrom(to []string, from, subject, htmlbody string) *Message {
 	msg.SetHeader("To", to...)
 	msg.SetHeader("Subject", subject)
 	msg.SetDateHeader("Date", time.Now())
-	msg.SetBody("text/html", htmlbody)
 	body, err := html2text.FromString(htmlbody)
 	if err != nil {
 		// TODO: report error ?
 	}
-	msg.AddAlternative("text/plain", body)
+	msg.SetBody("text/plain", body)
+
+	// TODO: avoid this (use a configuration switch?)
+	msg.AddAlternative("text/html", htmlbody)
 
 	return &Message{
 		Message: msg,
--- a/modules/mailer/mailer.go
+++ b/modules/mailer/mailer.go
@@ -36,11 +36,12 @@ func NewMessageFrom(to []string, from, subject, htmlbody string) *Message {
 	body, err := html2text.FromString(htmlbody)
 	if err != nil {
 		// TODO: report error ?
+		msg.SetBody("text/html", htmlbody)
+	} else {
+		msg.SetBody("text/plain", body)
+		// TODO: avoid this (use a configuration switch?)
+		msg.AddAlternative("text/html", htmlbody)
 	}
-	msg.SetBody("text/plain", body)
-
-	// TODO: avoid this (use a configuration switch?)
-	msg.AddAlternative("text/html", htmlbody)
 
 	return &Message{
 		Message: msg,


strk:
Uses the library suggested by @Jonny007-MKD -- keeps the current multipart body but avoids html in the text/plain part.

SirCmpwn:
Just

send

plain

text

Say NO to HTML emails!

strk:
@SirCmpwn being my first contribution to the codebase, I don't want to change too much of the code. Current templates are html, eventual users templates would be html too, not including HTML would mean not including the only manually authored template from the user.

I agree about not sending HTML mail, but if backward compatibility has to be retained there would need to be an option and a way for admins to opt-in for text rather than html mail.

SirCmpwn:
User authored templates can be plaintext too. *Do not send HTML emails*

strk:
@SirCmpwn I'm looking forward for your PR

strk:
I reordered the code so it's now just a matter of removing the .AddAlternative line to avoid HTML completely. But maybe it still needs a configuration variable ?

nd2s:
No one needs backwards compatibility in emails and it certainly doesn't need unnecessary external dependencies. Removing HTML tags is just plain ugly. The only correct way is to write plain text templates.

strk:
The html2text code doesn't just strip html tags, but also preserve links by explicitly showing the target url. I agree directly authoring templates in text only would be preferrable, but I found the mailer code being passed HTML code not coming from any template, which is what discouraged me from taking the path of assuming the input was plain text.

Jonny007-MKD:
As we are in such a fundamental discussion: how about writing the mail templates in markdown? It's target is to be readable in plaintext and the dependency for converting out to html is already there :)

nd2s:
@strk I meant the whole approach is ugly, not the output.

strk:
I agree the approach is ugly.
I'd like to see HTML completely removed from template to html body.
Only as my first PR I did not want it to become too big.
This is just a tiny improvement aimed at fixing an otherwise BROKEN
mail encoding.

I'll welcome any other PR ensuring no HTML ever gets to be sent.
If this code can serve as a start, use it, otherwise start another
branch and I'll be happy to help as I can.

Jonny007-MKD:
@strk: I agree with you :)

strk:
For the record: code hard-coding HTML for notifications is in mail.go, function SendIssueNotifyMail (see #2988)

modules/mailer/mailer.go
```diff
@@ -26,14 +27,20 @@ type Message struct {
...
 	msg := gomail.NewMessage()
 	msg.SetHeader("From", from)
 	msg.SetHeader("To", to...)
 	msg.SetHeader("Subject", subject)
 	msg.SetDateHeader("Date", time.Now())
+	body, err := html2text.FromString(htmlbody)
+	if err != nil {
+		// TODO: report error ?
+	}
 	msg.SetBody("text/plain", body)
```
bkcsoft: Should probably not set `text/plain` if there was an error...

bkcsoft:
This is probably the best option for now. In the future I'd like to have the template being markdown (`text/template`). The included comment is in markdown already, so it makes sense that the template is so as well.

For those that want `text/html` (future global/user-option) we could just render it to HTML and set the relevant body as well. But `text/plain` should *always* be plain markdown. :smiley: 

 @strk @Unknwon I've reviewed the code and it looks fine except for that one comment :smile: 

Pre-emptive passive-aggressiveness: **IF** someone have opinions against markdown in `text/plain`: since the user-provided content is already markdown with no proper way to "strip" it, it doesn't seem like a valid opinion :trollface:

nd2s:
@bkcsoft Did you just seriously suggest to put markdown in the `text/plain` part or was that sarcasm of some sort?

bkcsoft:
@nd2s Yes I did, did you read the entire comment or did you stop at "markdown"? I **don't** want to render markdown, just send it as-is as plain-text... the last paragraph explains why...

nd2s:
This is even more retarded than sending HTML-only mails. I think I'll have to switch to gitlab soon.

bkcsoft:
@nd2s This is what GitHub is doing (and GitLab btw, so have fun migrating :laughing: )

The email-output would look something like this: (took a random issue-mail)
```
When adding many tags there are white spaces between them on the issue page. Non selected tags are hidden but there's a blank space that persists.

![tags_spaces](https://cloud.githubusercontent.com/assets/1147214/15358975/5b8755de-1d07-11e6-89c5-6602cd0ac513.png)

- Gogs version (or commit ref): 0.9.13.0319
- Git version: 2.6.3.windows.1

Tcl/Tk version 8.6.4
- Operating system: Windows 7 x64
- Database:
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
  - [ ] No
  - [ ] Not relevant
```

Which is what I'd expect any issue email-system to output...

nd2s:
No this is **not** what Github is doing. I am getting Github plain text mails als the time (a few secs ago for example). And they are standard text mail, not markdown code. Links are normal links not this weird `[]()` shit.

You can use `*bold*`, `_underlined_` and `/italics/` in plain text mails, but this has nothing to do with markdown.

nd2s:
@bkcsoft Just look in #1496, there even is a screenshot of a Github mail including a link.

bkcsoft:
That snippet **is** a github-email.

If you're referring to this image, the link at the end is appended by github, not the comment (which you can see in _this_ mail :smile: )

bkcsoft:
Note that I'm not saying that the template itself should contain any Markdown, just that it could

nd2s:
Then this was a misunderstanding. I'm saying that Github's mail templates are not markdown as you can see on the screenshot (normal link, no `[]()` stuff), nor are the ones of any other site I would know. And that it makes no sense to send markdown code to users.

Just "inserting" some markdown from a comment into the plain text mail template, is ok of course (if it can't be avoided). But every code in the template itself is poison. As is converting HTML to plain text automatically.


And I just want to note that there are 5 thumbs-up and 1 thumbs-down on the commend that says HTML templates should be removed (and this ticket has 6 subscribers). The HTML templates really should go, they are just unnecessary bloat of the code base.

bkcsoft:
Heh, I guessed there was a misunderstanding somewhere :smiley:

The template I have in mind would look like this: (converted from `templates/mail/notify/mention.tmpl`)
```
@{{.ActUserName}} mentioned you:

{{.Content}}

---
View it on Gogs: {{.AppUrl}}{{.IssueLink}}
```

So in the end it would end up just like GitHub

Side-note: I'd probably change it so the mention-part is "optional" and the template can be re-used by other parts (where you aren't mentioned):
```
{{if .IsMentioned }}
@{{.ActUserName}} mentioned you:
{{end}}
```

nd2s:
Not sure if reuse makes sense with that few templates (that probably differ quite a lot). But I don't know the current ones.

But please use the signature style I mentioned earlier:

```
-- 
View it on Gogs: {{.AppUrl}}{{.IssueLink}}
```

Signature starting with dast-dash-whitespace-linebreak. This makes some email clients (like Thunderbird) detect the signature and lighten the text a bit.

bkcsoft:
Doesn't look like gogs has a non-mention template actually... And IMO it makes sense to have as few templates as possible to reduce the risk of similar templates ending up having different content all together :smile:

> please use the signature style I mentioned earlier

Obviously it would follow a convention that mail-clients can handle :smiley: 

> there are 5 thumbs-up and 1 thumbs-down on the commend that says HTML templates should be removed

I prefer HTML, so the preferably there would be an option for turning this on/off. One per user (I like it, you hate it), and one system-wide (bandwidth :unamused: )

bkcsoft:
ping @strk by request

strk:
Thank you, 无闻 -- welcome back !

--strk;

Unknwon:
https://try.gogs.io/ has been updated.

PR #2780, opened 2016-03-07
Add AddCollaborator API Endpoint

roblabla:
Adds the `PUT /repos/:username/:reponame/collaborators/:collaborator` endpoint, based on https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator, allowing the API to add collaborators to a repository.

I messed up #2775 (it was opened on the wrong branch).

**EDIT**: now implements `permission` like github's api, taking one of three arguments, "pull", "push" or "admin".
I can squash into one commit if that is prefered.

PR #2967, opened 2016-04-12
Allow admin to clone and push to all repositories

### LINKED ISSUE #2452: Admin user should be able to clone/push to all repositories
I think the admin user should be able to clone and push to all repositories without being in a organisation and team.

This is somewhat related to this issue: https://github.com/gogits/gogs/issues/1146

roblabla:
Fixes #2452. Also allows admin to see the private repos on the user's profile. 

Tested with SSH login.

### DIFF
--- a/models/access.go
+++ b/models/access.go
@@ -67,7 +67,11 @@ func accessLevel(e Engine, u *User, repo *Repository) (AccessMode, error) {
 		return mode, nil
 	}
 
-	if u.Id == repo.OwnerID {
+	else if u.IsAdmin {
+		return ACCESS_MODE_ADMIN, nil
+	}
+
+	else if u.Id == repo.OwnerID {
 		return ACCESS_MODE_OWNER, nil
 	}
 
--- a/routers/user/profile.go
+++ b/routers/user/profile.go
@@ -93,7 +93,7 @@ func Profile(ctx *context.Context) {
 		}
 	default:
 		var err error
-		ctx.Data["Repos"], err = models.GetRepositories(u.Id, ctx.IsSigned && ctx.User.Id == u.Id)
+		ctx.Data["Repos"], err = models.GetRepositories(u.Id, ctx.IsSigned && (ctx.User.IsAdmin || ctx.User.Id == u.Id))
 		if err != nil {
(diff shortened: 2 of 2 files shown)


bkcsoft:
The general concensus in that issue seemed to be no push-rights though. Otherwise looks good :smiley:

roblabla:
Well, that's a bit strange. The admin already has access to lots and lots of things (Like the settings panel). Why restrict push access ?

bkcsoft:
The reason for not allowing push-rights is explained in the comments for that Issue :)

roblabla:
I don't agree with the line of reasoning presented there. As I explained, the admin already has "too much access". To me, the administrator should be allowed to do whatever he wants, that's the whole point of that role.

I need that level of access for a bot that's running on my gogs instance (it automatically pushes changes to READMEs). Having to do a AddPrivilege->Push->RemovePrivilege is a bit of a bother. Not terrible, just annoying. Furthermore, so long as #2780 isn't merged, I can't even acquire those privileges through the API.

models/access.go
```diff
@@ -67,6 +67,10 @@ func accessLevel(e Engine, u *User, repo *Repository) (AccessMode, error) {
 		return mode, nil
 	}
 
+	if u.IsAdmin {
+		return ACCESS_MODE_ADMIN, nil
+	}
+
 	if u.Id == repo.OwnerID {
```
Unknwon: Please merge the if condition.

models/access.go
```diff
@@ -67,6 +67,10 @@ func accessLevel(e Engine, u *User, repo *Repository) (AccessMode, error) {
...
 		return ACCESS_MODE_OWNER, nil
 	}
```
Unknwon: Use `else if` should be better.

models/access.go
```diff
@@ -67,7 +67,11 @@ func accessLevel(e Engine, u *User, repo *Repository) (AccessMode, error) {
 		return mode, nil
 	}
 
-	if u.Id == repo.OwnerID {
+	else if u.IsAdmin {
+		return ACCESS_MODE_ADMIN, nil
+	}
+
```
Unknwon: Please make sure it compiles.

PR #2969, opened 2016-04-12
Fix remove folder issues, including initialization failling

andreynering:
- Prevent panic on creating notice if database is not available
- Prevent incorrect folder on Windows (`/` instead of `\`)

### DIFF
--- a/models/admin.go
+++ b/models/admin.go
@@ -52,6 +52,11 @@ func (n *Notice) TrStr() string {
 
 // CreateNotice creates new system notice.
 func CreateNotice(tp NoticeType, desc string) error {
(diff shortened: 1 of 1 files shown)


andreynering:
ping @Unknwon 

I improved this PR fixing a problem I was having with initialization.

Gogs was failing to initialize [with this log](https://gist.githubusercontent.com/andreynering/ac3105767fedf7bf0176911f04f26347/raw/09fc0bc24f78352778bc23df6e080bdeb314c531/gogs.log).

This happened because Gogs was trying to create a notice while the database reference was `nil`. I saw this both on Linux and Windows.

Unknwon:
Thanks a lot!

Issue #2972, opened 2016-04-13
Parameter mishandling on DSN during /install 


tcarrio:
Relevant info:

- Gogs version (or commit ref): **0.9.13**
- Git version: **1.7.1**
- Operating system: **Red Hat Enterprise Linux 6.5**
- Database: 
  - [X] PostgreSQL
  - [X] MySQL
  - [ ] SQLite
- Can you reproduce the bug at http://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [X] Not relevant
- Log gist: [https://gist.github.com/tcarrio/36f3d5fbe9368a98ce390efcb151cabb](https://gist.github.com/tcarrio/36f3d5fbe9368a98ce390efcb151cabb)

## Description
The installation process will fail at hostname:port/install if attempting to add parameters to the DSN. This is the suggested method in order to bypass the `old_passwords=1` setting in the `my.cnf` however. The error given for attempting to connect with no parameters  on the database name as `gogs` is:

`Database setting is not correct: this user requires old password authentication. If you still want to use it, please add 'allowOldPasswords=1' to your DSN. See also https://github.com/go-sql-driver/mysql/wiki/old_passwords`

But when modifying the DSN so that the database name entry is now `gogs?allowOldPasswords=1`, a different error is thrown stating that the parameter setting is incorrect:

`Database setting is not correct: invalid bool value: 1?charset=utf8`

This is likely due to the parameter `charset=utf8` being automatically added by the backend without checking whether there were other parameters attached to the DSN. If the function checks for existing parameters (easily done using a check for `?` in the database name), it can more intelligently append parameters and replace the `?` character with `&` since there are parameters previous to it. 

I may be taking a look at the source code later tonight to fork and fix myself, but figured I would toss in an issue now since I'm still busy. 
...

tcarrio:
I've located the issue and I'll be submitted a pull request later this evening.

tcarrio:
Issue resolved in referenced commit

PR #2974, opened 2016-04-14
getEngine() now handles DB parameters (#2972)

tcarrio:
For reference, Issue #2972. 

Beforehand, any DSN parameters on MySQL or Postgres databases would cause an error due to the concatenated strings not checking if parameters already exists. 

getEngine() now checks for existing parameters and correctly address the additional strings being concatenated for the databases according to whether a parameter exists. User may enter `database name` with or without parameters now. 

This helps for certain databases that require specific configurations, such as MySQL databases using `old_passwords=1` with the `allowOldPasswords=1` argument being passed to the DSN.

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -97,14 +97,18 @@ func LoadConfigs() {
 
 func getEngine() (*xorm.Engine, error) {
 	cnnstr := ""
+	var Param string = "?"
+	if strings.Contains(DbCfg.Name,Param) {
+		Param="&"
+	}
 	switch DbCfg.Type {
 	case "mysql":
 		if DbCfg.Host[0] == '/' { // looks like a unix socket
-			cnnstr = fmt.Sprintf("%s:%s@unix(%s)/%s?charset=utf8&parseTime=true",
-				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name)
+			cnnstr = fmt.Sprintf("%s:%s@unix(%s)/%s%scharset=utf8&parseTime=true",
+				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
 		} else {
-			cnnstr = fmt.Sprintf("%s:%s@tcp(%s)/%s?charset=utf8&parseTime=true",
-				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name)
+			cnnstr = fmt.Sprintf("%s:%s@tcp(%s)/%s%scharset=utf8&parseTime=true",
+				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
 		}
 	case "postgres":
 		var host, port = "127.0.0.1", "5432"
@@ -115,8 +119,8 @@ func getEngine() (*xorm.Engine, error) {
 		if len(fields) > 1 && len(strings.TrimSpace(fields[1])) > 0 {
 			port = fields[1]
 		}
-		cnnstr = fmt.Sprintf("postgres://%s:%s@%s:%s/%s?sslmode=%s",
-			url.QueryEscape(DbCfg.User), url.QueryEscape(DbCfg.Passwd), host, port, DbCfg.Name, DbCfg.SSLMode)
+		cnnstr = fmt.Sprintf("postgres://%s:%s@%s:%s/%s%ssslmode=%s",
+			url.QueryEscape(DbCfg.User), url.QueryEscape(DbCfg.Passwd), host, port, DbCfg.Name, Param, DbCfg.SSLMode)
 	case "sqlite3":
 		if !EnableSQLite3 {
 			return nil, fmt.Errorf("Unknown database type: %s", DbCfg.Type)


tcarrio:
Thought I should add that the affected databases were MySQL **and** PostgreSQL. Both had appended paremeters with no checks on the passed DSN. Tests were run using MySQL, but the DSN formatting is equivalent across both.

tcarrio:
This fix was created since the codebase does have DSN parameter functionality even though it is suggested by gogs to append them. 

`Database setting is not correct: this user requires old password authentication. If you still want to use it, please add 'allowOldPasswords=1' to your DSN. See also https://github.com/go-sql-driver/mysql/wiki/old_passwords`

Is the error message at the top of gogs when connecting to a database using old passwords. This issue exists for any database requiring parameters passed to the data source name via the database driver URI.

bkcsoft:
That error-message is most likely generated by go-sql-driver/mysql and not gogs itself. (Mainly mentioning this so there's no confusion...)

models/models.go
```diff
@@ -97,14 +97,18 @@ func LoadConfigs() {
 
 func getEngine() (*xorm.Engine, error) {
 	cnnstr := ""
+	var Param string = "?"
+	if strings.Contains(DbCfg.Name,Param) {
+		Param="&"
+	}
```
bkcsoft: These above 4 lines can be removed by doing `?charset=utf8&parseTime=true%s` instead below which makes more sense.

models/models.go
```diff
@@ -97,14 +97,18 @@ func LoadConfigs() {
...
 	switch DbCfg.Type {
 	case "mysql":
 		if DbCfg.Host[0] == '/' { // looks like a unix socket
-			cnnstr = fmt.Sprintf("%s:%s@unix(%s)/%s?charset=utf8&parseTime=true",
-				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name)
+			cnnstr = fmt.Sprintf("%s:%s@unix(%s)/%s%scharset=utf8&parseTime=true",
+				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
```
bkcsoft: see above

models/models.go
```diff
@@ -97,14 +97,18 @@ func LoadConfigs() {
...
 		} else {
-			cnnstr = fmt.Sprintf("%s:%s@tcp(%s)/%s?charset=utf8&parseTime=true",
-				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name)
+			cnnstr = fmt.Sprintf("%s:%s@tcp(%s)/%s%scharset=utf8&parseTime=true",
+				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
```
bkcsoft: see above

models/models.go
```diff
@@ -115,8 +119,8 @@ func getEngine() (*xorm.Engine, error) {
 		if len(fields) > 1 && len(strings.TrimSpace(fields[1])) > 0 {
 			port = fields[1]
 		}
-		cnnstr = fmt.Sprintf("postgres://%s:%s@%s:%s/%s?sslmode=%s",
-			url.QueryEscape(DbCfg.User), url.QueryEscape(DbCfg.Passwd), host, port, DbCfg.Name, DbCfg.SSLMode)
+		cnnstr = fmt.Sprintf("postgres://%s:%s@%s:%s/%s%ssslmode=%s",
+			url.QueryEscape(DbCfg.User), url.QueryEscape(DbCfg.Passwd), host, port, DbCfg.Name, Param, DbCfg.SSLMode)
```
bkcsoft: see above

models/models.go
```diff
@@ -115,8 +119,8 @@ func getEngine() (*xorm.Engine, error) {
...
 	case "sqlite3":
 		if !EnableSQLite3 {
 			return nil, fmt.Errorf("Unknown database type: %s", DbCfg.Type)
```
bkcsoft: Warn if `Param` is set in this mode? Since otherwise it's silently ignored (not very nice...)

tcarrio:
As of now the setup page for gogs only has a Database textfield where the user inputs the DSN. 

This lets the user input the DSN as they would with any other database connector without gogs throwing errors. Before it had no logic to discern whether the DSN contained parameters and therefore caused the DSN to be malformed if there were any added by the user. This was a shortsighted in its initial design as the error thrown by `go-sql-driver/mysql` mentions a malformed DSN.

The current implementation takes the DSN textfield and automatically appends parameters. This is fine until the user wants to add their own, where the [query](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax) appended to their DSN, signified by the `?`, will cause the automatically appended `?` in the `models/models.go` to short.

There are obviously several different way of handling the query portion of the DSN in `models/models.go`, however this method will simply check the input DSN for a query, and if it exists, will instead add further parameters in `models/models.go` with the correct delimiter. 

`Param` is a somewhat poor variable name, is signifies whether a parameter existed in the DSN, and contains either the query character (`?`) or delimiter (`&`) accordingly. 

I don't see how the DSN could be generated with your suggestion of using `?charset=utf8&parseTime=true%s`, without splitting the DSN using a `?` delimiter to pull the query portion of the DSN, which would actually require more processing than this proposed solution. 

Example of current solution:

Using:

* `URI: database.example.com:3306`
* `DSN: exampledb`

```go
# models/models.go
# cnnstr = fmt.Sprintf("%s:%s@unix(%s)/%s?charset=utf8&parseTime=true",
# DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name)
cnnstr = "user:password@database.example.com/exampledb?charset=utf8&parseTime=true"
```

which works fine, but once you include parameters in your DSN (`exampledb?testAttribute=true`):

```go
cnnstr = "user:password@database.example.com/exampledb?testAttribute=true?charset=utf8&parseTime=true"
```

While URI's containing multiple `?`s does not make the URI invalid, however subsequent entries of `?` are viewed as text, so the `testAttribute=true` becomes `testAttribute=truecharset=utf8`, which is not a valid entry for the database and also drops your `charset=utf8` argument. 

Now my solution, which checks the input DSN for a query (`?`), allows the user to input completely valid DSN entries while allowing `models/models.go` to still append the parameters it needs. 

`DSN: exampledb?testAttribute=true`

```go
# because DbCfg.Name contains "?":

cnnstr = "user:password@database.example.com/exampledb?testAttribute=true&charset=utf8&parseTime=true"

If you would like to split the DSN on `?` indefinitely instead of using a separate query/delimiter string that is based on the users DSN, it would work of course, but it's not as efficient.

models/models.go
tcarrio: I don't believe it requires warning, `Param` is only changed to "&" if the user has input a DSN that uses (or even requires) parameters.

Unknwon:
Any updates after @bkcsoft 's comments? Seems no?

tcarrio:
There were no updates since no one replied to my explanation for the change. 

It seems there's a misunderstanding of what has been changed in this PR.

This only changes the way the URI is generated, so that existing parameters are still used to maintain consistency with the current code while allowing the user to add their own, which may be necessary for certain databases depending on their configuration. 

> bkcsoft:
> Warn if Param is set in this mode

There are **already** arguments attached to each of the databases, that is what caused the issue in the first place since gogs does not correctly generate the DSN URI. Regardless of user-attached parameters, gogs automatically added `?sslmode=%s` to PostgreSQL database URIs and `%scharset=utf8&parseTime=true` to MySQL database URIs. 

This PR fixes the ability to add parameters to databases that require them.

Unknwon:
Thanks. Just want to make sure, there is no possibilities that database name could contain `?`?

tcarrio:
No, [URI](https://en.m.wikipedia.org/wiki/Uniform_Resource_Identifier)s use `?` to signify the start of the query, such as `?charset=utf8&parseTime=true`, and are not allowed in the [DSN](https://en.m.wikipedia.org/wiki/Data_source_name).  This change will allow more functionality in the user creation of the DSNs in the configuration of gogs.

Unknwon:
OK, thanks!

PR #2980, opened 2016-04-15
[Feature/WIP] Confirm on editing forms (fixes #2881)

### LINKED ISSUE #2881: ask user to confirm when he/she leaves wiki edit page if the page not saved
I'm editing a wiki page but didn't save it, and hit the backspace outside the edit box by an accident, then I lost all my work.

suggest to save wiki to a local storage like cookie periodic during edit. and ask user to recover when he/she comes back

another way for this issue may be: ask user to confirm when he/she leaves the edit page. this maybe easier to implement.

bkcsoft:
All edited (and saved) forms in Gogs will now ask you if you really want to leave the unchanged form.

Fixes #2881 and Updates SimpleMDE to 1.10.

(I'll clean this up before actually merging...)


mhartkorn:
Why did you create a new `.js` file? That's an additional request for all page loads.

bkcsoft:
jquery.are-you-sure is the one doing all the works.
This works on any form anywhere in gogs with a single line of code (3 actually), hence the global inclusion.

As for the extra request, there's a bunch of js/css-files already included (even when not used in that page), if keeping requests low is an issue it's a separate issue and in my opinion not a very relevant one with http/2 :smile: (actually http/1.1 keeps the connection open for subsequest requests if the browser/server has support for this)

Basically the choice was between areYouSure or writing (and maintaining) our own implementation. I opted for including areYouSure.

PR #2990, opened 2016-04-18
Independ change to src tree to automate rpm creation for CentOS 7

tjyang:
Hi Unknown,

I added a "contrib" dir with 4 files to automate the rpm creation.
It is still raw and not upto standard for submitting to Fedora repo. 
But is it is good enough for me for my own project to start with.

Hope others find this useful.


hhenkel:
I think it is a good idea to add the stuff necessary to build an rpm. I just used that and was able to build an rpm with that.

However I'm wondering why you took the approach to compile externally of the rpmbuild. From what I've seen you got golang as a build dependency which seems unnecessary how this is currently working. I'm also not sure if it is a good idea to require mariadb/mysql-server ias one could also choose postgresql, sqlite or tidb.

As I most of the time build from git source I altered some things to fetch the version on the fly and added the short version of the git head to the package. Mainly I'm too lazy to alter the version in two files, therefore I guess that would be an improvement.

tjyang:
> However I'm wondering why you took the approach to compile externally of the rpmbuild. 
Because the change is to aim to be low impact to existing to source tree. one of the todo is to insert gogs in spec file's build tag. 

> I'm also not sure if it is a good idea to require mariadb/mysql-server ias one could also choose postgresql, sqlite or tidb.
True. this part can be changed. The dependency should be at rpm build time.

So far I yet to convince @Unknwon  that packaging in general is important to the Gogs project.

Unknwon:
I think it would be better use the same approach that Jenkins uses, which creates a repository called `packaging` puts all such things together.

tjyang:
>I think it would be better use the same approach that Jenkins uses, which creates a repository called packaging puts all such things together.
Yes, I like to use Jenkins' directory layout. (change of "packager" dir path need packager.io folk's comment). 
I can submit another pull request if we all agree.

Unknwon:
Feel free to send a PR under `rpm` subdirectory on https://github.com/gogits/packaging

tjyang:
Sorry, there is a miss communication here. I meant following directory layout. Not a separate repo under gogits. This way we can have top level main makefile to call packaging/rpm/Makefile to build the rpm package.

gogs
├── cmd
├── conf
├── contrib
├── docker
├── Dockerfile
├── Dockerfile.rpi
├── glide.lock
├── glide.yaml
├── gogs
├── gogs-0.9.20.tar.gz
├── gogs.go
├── LICENSE
├── Makefile
├── models
├── modules
├── packager
├── packaging
         -rpm
         -msi
├── public
├── README.md
├── README_ZH.md
├── routers
├── scripts
└── templates
`
`

Unknwon:
No, I do not want to include these files in the main repo.

tjyang:
Since "packager" directory for rpm package  by packager.io is created already. 
May I know what is your reasoning  of not creating "packaging" directory ?

Unknwon:
To be honest, I do not understand what do these files do in the PR.

And following the practice of Jenkins, put all files in a separate repository seems better solution to me. The `packager` is just for `packager.io`, not general purpose `pakcaging`.

tjyang:
Ok, now I understand you want to protect your codebase. I will submit PR to gogits/packaging repo.  That means  I would need to checkout src first to build and then check out gogits/packaging repo for rpm building.  Not ideal but acceptable.

tjyang:
Native packaging effort redirect to https://github.com/gogits/packaging/pull/1

PR #2992, opened 2016-04-19
Fixes #2669: Allows alphanumeric issue style for external issue tracker links

### LINKED ISSUE #2669: Support differnt shape issue tracker IDs with RegEx
In our issue tracker of choice (JIRA) we've got several project for different teams (Quality, Support, Development). These project have different "issue id formats" like QM-55, SUP-879 or DEV-357 they should redirect to  a URL like these https://_URL_/browse/DEV-357. So it would be great to take up the idea of @rojaro (in issue #890) again and implement a RegEx based issue ID identification.

cstroe:
Allow to use alphanumeric issue mentions (e.g. ABC-1234) for external links.  This works specifically with JIRA and Bloodhound.

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -576,6 +576,9 @@ settings.external_wiki_url_desc = Visitors will be redirected to URL when they c
 settings.issues_desc = Enable builtin lightweight issue tracker
 settings.use_external_issue_tracker = Use external issue tracker
 settings.tracker_url_format = External Issue Tracker URL Format
+settings.tracker_issue_style = External Issue Tracker Naming Style:
+settings.tracker_issue_style.numeric = Numeric <span class="ui light grey text">(1234)</span>
+settings.tracker_issue_style.alphanumeric = Alphanumeric <span class="ui light grey text">(ABC-123, DEFG-234, ...)</span>
 settings.tracker_url_format_desc = You can use placeholder <code>{user} {repo} {index}</code> for user name, repository name and issue index.
 settings.pulls_desc = Enable pull requests to accept public contributions
 settings.danger_zone = Danger Zone
--- a/models/repo.go
+++ b/models/repo.go
@@ -170,6 +170,7 @@ type Repository struct {
 	EnableIssues          bool `xorm:"NOT NULL DEFAULT true"`
 	EnableExternalTracker bool
 	ExternalTrackerFormat string
+	ExternalTrackerStyle  string
 	ExternalMetas         map[string]string `xorm:"-"`
 	EnablePulls           bool              `xorm:"NOT NULL DEFAULT true"`
 
@@ -254,6 +255,13 @@ func (repo *Repository) ComposeMetas() map[string]string {
 			"user":   repo.MustOwner().Name,
 			"repo":   repo.Name,
 		}
+		switch repo.ExternalTrackerStyle {
+		case markdown.ISSUE_NAME_STYLE_ALPHANUMERIC:
+			repo.ExternalMetas["style"] = markdown.ISSUE_NAME_STYLE_ALPHANUMERIC
+		default:
+			repo.ExternalMetas["style"] = markdown.ISSUE_NAME_STYLE_NUMERIC
+		}
+
 	}
 	return repo.ExternalMetas
 }
--- a/modules/markdown/markdown.go
+++ b/modules/markdown/markdown.go
@@ -22,6 +22,11 @@ import (
 	"github.com/gogits/gogs/modules/setting"
 )
 
+const (
+	ISSUE_NAME_STYLE_NUMERIC = "numeric"
+	ISSUE_NAME_STYLE_ALPHANUMERIC = "alphanumeric"
+)
+
 var Sanitizer = bluemonday.UGCPolicy()
 
 // BuildSanitizer initializes sanitizer with allowed attributes based on settings.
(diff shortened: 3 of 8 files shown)


bkcsoft:
Why so specific? Why not allow `[a-zA-Z0-9-]` ?

andreynering:
:+1: for the unit tests.

Though, we should discuss what testing lib we will use in Gogs. I used [testify](https://github.com/stretchr/testify) on https://github.com/gogits/gogs/pull/2948. You are using [ginkgo](https://github.com/onsi/ginkgo) here. Unknows's [ini](https://github.com/go-ini/ini) use [goconvey](https://github.com/smartystreets/goconvey), which I assume he prefers.

Idealy we should choose one, or we will endup with many different test libs.

cstroe:
@bkcsoft: I wanted to make a very strict regex, the minimum possible, to reduce false positives on detecting the issue mentions.

JIRA specifies that the project shortname (the prefix of the issue) can be at most 10 characters.  Limiting it to only uppercase letters is also with the intent or reducing false positives.  It also allows issue mentions to have a distinct visual style and stand out from normal text.

Using the regex you gave would have the side effect of matching hyphenated words, and other non-issue words.  For example, abc-def and 123-456 would be converted, where they would not form a valid link.  

I have to admit that my aim is very JIRA-centric, so if there are other issue trackers that you would like to support, please let me know.

@andreynering: I don't have a.problem converting the tests to another framework.  I prefer the BDD style, as it self documents the tests, so that intent is quickly readable.

cstroe:
@andreynering: I converted the tests to goconvey to avoid introducing another dependency.

andreynering:
@cstroe Thank you, my vote is for goconvey, too. :+1: 

Since @Unknwon is the boss here, he will decide this, and about your PR, too. Let's wait.

cstroe:
Screenshots, why not.

Updated repo settings:
gogs-repo-settings

Issue replace in action:
gogs-external-tracker-replace

conf/locale/locale_en-US.ini
```diff
@@ -576,6 +576,9 @@ settings.external_wiki_url_desc = Visitors will be redirected to URL when they c
 settings.issues_desc = Enable builtin lightweight issue tracker
 settings.use_external_issue_tracker = Use external issue tracker
 settings.tracker_url_format = External Issue Tracker URL Format
+settings.tracker_issue_style = External Issue Tracker Issue Style:
```
Unknwon: Maybe `External Issue Tracker Naming Style`?

models/repo.go
```diff
@@ -254,6 +255,14 @@ func (repo *Repository) ComposeMetas() map[string]string {
 			"user":   repo.MustOwner().Name,
 			"repo":   repo.Name,
 		}
+		switch repo.ExternalTrackerStyle {
+		case "numeric": fallthrough
```
Unknwon: I can't remember how exactly `fallthrough` works, but I think it fallthroughs to `case "alphanumeric":` and is not what we intend to?

I suggest just remove this case, let `default` handles it.

modules/markdown/markdown.go
```diff
@@ -208,22 +210,30 @@ func cutoutVerbosePrefix(prefix string) string {
...
 		if metas == nil {
-			rawBytes = bytes.Replace(rawBytes, m, []byte(fmt.Sprintf(`%s<a href="%s/issues/%s">%s</a>`,
-				space, urlPrefix, m[1:], m)), 1)
+			link = fmt.Sprintf(`<a href="%s/issues/%s">%s</a>`, urlPrefix, m[1:], m)
 		} else {
 			// Support for external issue tracker
-			metas["index"] = string(m[1:])
-			rawBytes = bytes.Replace(rawBytes, m, []byte(fmt.Sprintf(`%s<a href="%s">%s</a>`,
-				space, com.Expand(metas["format"], metas), m)), 1)
+			if metas["style"] == "numeric" {
```
Unknwon: I think here we should check `"alphanumeric"` and leave `else` for `"numeric"`, because that's the default one.

cstroe:
@Unknwon : Made the changes you asked.

models/repo.go
```diff
@@ -254,6 +255,13 @@ func (repo *Repository) ComposeMetas() map[string]string {
 			"user":   repo.MustOwner().Name,
 			"repo":   repo.Name,
 		}
+		switch repo.ExternalTrackerStyle {
+		case "alphanumeric":
```
Unknwon: Should we use constantS like `ISSUE_NAME_STYLE_ALPHANUMERIC` and `ISSUE_NAME_STYLE_NUMERIC` to avoid spell it every time we use?

Unknwon:
Would mind cleanup the commits to a single one?

cstroe:
@Unknwon: I put the constants in the `markdown` package.  Is there a better place to put them so that we can access them in templates too?  Is that worth it?

Unknwon:
> I put the constants in the markdown package.

I think that's the best place for now.

Unknwon:
It's now live on https://try.gogs.io

i00:
Think it would be better if you allowed for custom items (regex and a description) in the config, so you can easily add more and then choose from what one you want for the repo in the interface...

..My issue tracker uses guid's

Unknwon:
@i00 That would be the next step, but custom thing can be miss used a lot and cause some strange issues.

i00:
@Unknwon - seems to me that this is kind of a band-aid solution if it is implemented for a specific case only, I am sure that most developers would know regex, should let them choose in the repo settings.

Unknwon:
@i00 would be happy to review a PR for that :)

i00:
Would be great thanks, I don't mean having an interface for doing this btw (in the config file for eg)... just some place where you can specify on a repo, by repo basis what one of the built in version numbering systems to use....
And by "built in" I mean ones that are defined in the config file (so we can add our own).

mqudsi:
This does not seem to catch issue numbers placed in `[]`, such as `[PROJ-1234]`, which is a very common requirement for referencing issues everywhere except on GitHub.

cstroe:
@mqudsi When rendering the issue numbers placed in `[]`, are the square brackets preserved in the rendered link?

mqudsi:
No, as an example, [PROJ-123] should become http://issuetracker/PROJ-123

While I personally have no need for it, I have seen issue trackers that would end up with a URL like http://issuetracker/proj/123 which would require the tokenization of the key into separate alphanumeric+numeric, btw.

cstroe:
It might be worth revamping how we currently translate issue mentions into their tracker URLs, just to have a fully flexible translation system.  But I'm not going to tackle that unless there's a compelling use case for building such a flexible system.

cstroe:
Also, for rendering `http://issuetracker/proj/123`, that's possible by using the numeric style of mentions, because the tracker URL template is defined per repository (as long as there's a one to one relationship between tracker project and repository).

Issue #2993, opened 2016-04-19
PDF previewing


tforgione:
[This post explains](https://github.com/blog/1974-pdf-viewing) explains how github allow users to see a versionned pdf file directly on the website.

There is an example [here](https://github.com/nasa/pvslib/blob/5e52ed24e8dafab5ee1de8626f6fcd721a292547/vectors/doc/intersection-paper-eberly.pdf).

It would be nice to have such a feature on Gogs !

bkcsoft:
This is pretty simple. Claimed ^.^

bkcsoft:
@tforgione feel free to checkout #3000 and see if it works

Unknwon:
This is claimed to be implemented by merging #3000 , please help test on `develop` branch or https://try.gogs.io.

tforgione:
I installed the develop version of gogs on my pc, and added a pdf on a test repository.

The pdf preview works successfully !

Unknwon:
@tforgione Thanks your confirmation!

PR #3000, opened 2016-04-20
[Feature] pdf preview #2993

bkcsoft:
Implements PDF Previewer as per #2993 

Obligatory Screenshot:
2016-04-20-132321_1596x861_scrot


tforgione:
Yep, this works correctly for me !

In fact, it's even better than github's method since they render the pdf to a canvas, and therefore, no search or copying from pdf is available !

bkcsoft:
@Unknwon Thoughts?

Unknwon:
Looks awesome, but 371 file changes?

lunny:
Maybe pdf.js could be a git submodule on gogs so that to keep it update?

lenovouser:
Maybe even consider using NPM for managing your JavaScript dependencies instead of submodules.

andreynering:
My opinion is we should let the browser take care of it.

Firefox actually use pdf.js to handle any PDF file. Chrome has it's own viewer, too. Others falls back to what editor is available in the OS (e.g. Adobe Reader, Foxit, whatever).

Anyway it's better than including 371 files (mostly binary), since the browser already handles it.

So, I think just including an iframe with the PDF URL is enough.

bkcsoft:
@Unknwon loads of locale and cmaps-files :unamused: 

@lunny hate submodules but it is indeed an idea
@lenovouser nope, not even once :rage:

@andreynering the browser can't take care of it if we want it inside the file-list (see the corresponding Issue). if is already using an iframe though (easier and better IMO (since contained JS-execution)), and you just reminded me that this always runs in Offline Mode, even if unchecked (should use https://mozilla.github.io/pdf.js/web/viewer.html when not in Offline Mode...)

bkcsoft:
@lenovouser and in any case that would be bower, not npm, since npm is for node-modules while bower is for browser-modules

Unknwon:
This is ... hard to merge.

bkcsoft:
@Unknwon I could remove the locales and cmaps (risking breaking PDFs...) and we're down to 4-5 files...

Unknwon:
What do `cmaps` do?

bkcsoft:
@Unknwon https://blog.idrsolutions.com/2012/05/understanding-the-pdf-file-format-embedded-cmap-tables/

Unknwon:
How many files left if remove locales?

bkcsoft:
there are 169 cmaps, 105 locales, 6 js-files, 1 css, 1 html and 84 images (icons for buttons)

bkcsoft:
the entire file-tree: https://github.com/bkcsoft/gogs/tree/feature/pdf-preview/public/plugins/pdfjs-1.4.20

Unknwon:
I think remove locales could be the first step.

bkcsoft:
@Unknwon done

Unknwon:
>https://blog.idrsolutions.com/2012/05/understanding-the-pdf-file-format-embedded-cmap-tables/

From what I understand from this article, the CMAP tables are used for copy/paste/search? But display would be correct?

bkcsoft:
It's also possible to do it like Github... with something like this... http://mozilla.github.io/pdf.js/examples/learning/prevnext.html

bkcsoft:
The problem with doing it like Github is keeping track of multiple canvas-elements :trollface:

Unknwon:
How bad it is without including CMAP tables/

bkcsoft:
Seems to be working... 

2016-04-27-024308_1596x861_scrot

routers/repo/view.go
```diff
@@ -92,9 +91,12 @@ func Home(ctx *context.Context) {
 
 			_, isTextFile := base.IsTextFile(buf)
 			_, isImageFile := base.IsImageFile(buf)
+			_, isPdfFile := base.IsPdfFile(buf)
```
Unknwon: `IsPDFFile` feels better? Same as below.

bkcsoft:
@Unknwon Done

modules/base/tool.go
```diff
@@ -522,3 +522,11 @@ func IsImageFile(data []byte) (string, bool) {
 	}
 	return contentType, false
 }
+
+func IsPdfFile(data []byte) (string, bool) {
```
Unknwon: Hmm?

Unknwon:
Also, clean up to 1 commit would be the best.

bkcsoft:
Done. And I was just keeping the history while fiddling around :trollface:

Unknwon:
Thanks!

PR #3008, opened 2016-04-23
init script call gogs with wrong params

raywill:
we should run gogs like
```
gogs run web
```

instead of
```
gogs web
```

Otherwise we would get error like below in log file.
```
[rh@CloudY gogs]# gogs web
Usage:
  gogs run COMMAND [options]
  gogs scale TYPE=NUM
  gogs logs [--tail|-n NUMBER]
  gogs config:get VAR
  gogs config:set VAR=VALUE
  gogs configure
  gogs reconfigure
```

### DIFF
--- a/scripts/init/centos/gogs
+++ b/scripts/init/centos/gogs
@@ -49,7 +49,7 @@ DAEMON_OPTS="--check $NAME"
 start() {
   cd ${GOGS_HOME}
   echo -n "Starting ${SERVICENAME}: "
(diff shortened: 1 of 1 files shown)


bkcsoft:
Please refer to this: https://github.com/gogits/gogs/pull/2322

tboerger:
:-1: as this is specific to packager.io

bkcsoft:
Closing as invalid

Issue #2940, opened 2016-04-06
[FR] git-daemon-export-ok


ghost:
Would it be possible to add an option to ensure presence of `git-daemon-export-ok` file for public repositories and its absence for private ones? The presence of this file guides the `git-daemon` used to provide `git://` protocol as to which repositories to actually expose.

AFAICT, using `git-daemon` to expose repos read-only in parallel with gogs works, but there is no distinction for private / public ones, so it exposes either none of them (cause the file is not present), or all of them (if run with `--export-all`).

strk:
See also #2950 for another problem related to Gogs making no distinction between private / public repos.

strk:
Looking at the code the support for adding/removing `git-export-ok` should be added to https://github.com/gogits/git-module, I suggest you file an issue there @herby

herby:
@strk Why there? It depends on the private/public setting of the gogs itself, it is not very much of a git operation... ping @Unknwon.

strk:
Because that's the module used to interact with the git repo.
So that module should expose a function to add/remove the git-daemon-export-ok
and Gogs itself should be using that function.

Ideally the module would also expose functions to set the repository ownership and permissions, to deal with #2950

herby:
I see.

Dňa 16. apríla 2016 19:10:37 CEST používateľ Sandro Santilli <notifications@github.com> napísal:

Unknwon:
Hmm... any reference for `git-daemon-export-ok`, did not know about this feature of git before.

herby:
It's feature of `git daemon`, https://git-scm.com/book/en/v2/Git-on-the-Server-Git-Daemon.

无闻 wrote:

Unknwon:
Thanks for the link. I think we need to let Gogs support `git://` protocol first?

herby:
I don't know what do you mean by that. I asked you via comment on faq if I can run git-daemon in parallel and you said it was ok. This would just tell it which o es to expose and which not.
If you mean to support git:// protocol directly by gogs code, that would be completely different issue.

strk:
I agree there's no need for Gogs to support git:// protocol itself
(although it could be useful to _allow_ adding an indication of git://
access method when available by external means).

On the other hand, having access mediated by different servers is
always a risk, unless all servers carefully agree on what kind
of access level to grant to the resources (I'm thinking about the
"require sign in" setting of Gogs for public repos, for example).

The usage of file-based configurations (like git-export-ok) is a good
[...]

PR #3046, opened 2016-05-01
Add "support" for git-daemon-export-ok (#2940)

bkcsoft:
PR for #2940 

This PR adds "support" for `git-daemon`, I write "support" in quotes because it does it stupidly (i.e. no error-checking outside what is required for Gogs not to crash :smiling_imp:)

I did it this way because actually checking for errors will be a major headache (revert updates, return a nice error-message to the WebUI, translations for said error, etc, etc, etc) and this is not really Gogs job (I would have made a 3rd daemon that polls gogs via rest-api every now and then and update the files appropriately... as you're already running 2 services (gogs + git-daemon) you could just as well run 3...)

Update: Now checks and logs errors, but still doesn't care about them :trollface:


bkcsoft:
Ooh! Gogs doesn't no checks for this file, so when this PR is merged and installed, you still have to add all the files in the correct place, or change the settings twice (to get the original setting back)

herby:
@bkcsoft 

3rd daemon is something I did not think about. But it has two disadvantages: it lags and can stop working silently. Especially the second is problematic and having it integrated into gogs itself seems to be the right thing.

The fact that I would need to add files manually initially does not bother me (though, if this gets merged I would want an option added in admin panel "regenerate" list to check all repos and add/remove the file appropriately, as it does for other things like .ssh/authorized_keys).

herby:
As for error checking, I would probably close it forcefully and ignore any error happening withing `Close` (no need to check if there were error while opening).

bkcsoft:
Anything that stops silently without notifying anyone isn't configured correctly :laughing:

Jokes aside, 3rd daemon is error-prone yes. I'm not sure if gogs already have a update-repo button in the admin section, in that case this should already work, otherwise I *might* add said button at a later time, though I'm in no rush :smile:

bkcsoft:
About errors, awesome! I'm used to languages that segfault if you close a
broken filedescriptor :laughing:

herby:
I am ok with "in good will" solution to add/remove the file as well as with no rush to "reapply to all" in admin panel.

Kim Carlbäcker wrote:
> that stops silently without notifying anyone isn't configured

herby:
Ah. Sorry. If that's the case, ignore me. I have never thought of such possibility, I live in higher languages world where closing file produces error not unconditional crash.

Kim Carlbäcker wrote:

bkcsoft:
There's no issue, golang produces errors only (unlike C). Will fix that and
push a new one :smile:

models/repo.go
```diff
@@ -1273,6 +1273,19 @@ func updateRepository(e Engine, repo *Repository, visibilityChanged bool) (err e
 			}
 		}
 
+		// Create/Remove git-daemon-export-ok for git-daemon...
+		daemonExportFile := strings.Join([]string{repo.RepoPath(), `git-daemon-export-ok`}, "/")
```
Unknwon: Why not use `path.Join`

bkcsoft: Correct you are! Must've been tired when I wrote that. Thank god for code reviews :laughing:

bkcsoft:
@Unknwon fixed the `strings.Join => path.Join` and removed some unnecessary error-checking

models/repo.go
```diff
@@ -1273,6 +1273,17 @@ func updateRepository(e Engine, repo *Repository, visibilityChanged bool) (err e
...
 		}
 
+		// Create/Remove git-daemon-export-ok for git-daemon...
+		daemonExportFile := path.Join(repo.RepoPath(), `git-daemon-export-ok`)
+		if repo.IsPrivate {
+			// NOTE: Gogs doesn't actually care about this file so we don't do any error-checking :D
+			os.Remove(daemonExportFile)
+		} else {
+			// NOTE: Gogs doesn't actually care about this file so we don't do any error-checking :D
+			f, _ := os.Create(daemonExportFile)
```
Unknwon: Maybe just call `ioutil.WriteFile` with 0 bytes? And Make an error log `log.Error` if any error occurs.

bkcsoft: Any reason not to use `os.Create`? they perform the same task, just that `os.Create` is self-documenting while `ioutil.WriteFile` 0 bytes isn't (IMO). Since what we're doing is _creating_ a file, not writing to it :smile: 

I'll fix the error-checking though. Might as well make the admin aware of the fact that something is wrong :laughing:

Unknwon: You're right, add a notice for error should be enough.

bkcsoft: I'll add error-logging for `os.Remove` as well, since most admins wanna know if a (now) private repo might still be public... :worried:

Unknwon:
Please sign the CLA.

models/repo.go
```diff
@@ -1275,6 +1275,22 @@ func updateRepository(e Engine, repo *Repository, visibilityChanged bool) (err e
 			}
 		}
 
+		// Create/Remove git-daemon-export-ok for git-daemon...
+		daemonExportFile := path.Join(repo.RepoPath(), `git-daemon-export-ok`)
+		if repo.IsPrivate {
+			// NOTE: Gogs doesn't actually care about this file so we don't do any error-checking :D
```
Unknwon: This comment is somewhat, inappropriate now?

models/repo.go
```diff
@@ -1275,6 +1275,22 @@ func updateRepository(e Engine, repo *Repository, visibilityChanged bool) (err e
...
+			if err = os.Remove(daemonExportFile); err != nil {
+				log.Error(0, "Failed to remove %s: %s", daemonExportFile, err.Error())
```
Unknwon: Use 4.

models/repo.go
```diff
@@ -1275,6 +1275,22 @@ func updateRepository(e Engine, repo *Repository, visibilityChanged bool) (err e
...
+			}
+		} else {
+			// NOTE: Gogs doesn't actually care about this file so we don't do any error-checking :D
+			if f, err := os.Create(daemonExportFile); err != nil {
+				log.Error(0, "Failed to create %s: %s", daemonExportFile, err.Error())
```
Unknwon: Use 4, not 0.

bkcsoft:
Done. No clue where to sign the CLA :trollface:

models/repo.go
```diff
@@ -1275,6 +1275,20 @@ func updateRepository(e Engine, repo *Repository, visibilityChanged bool) (err e
 			}
 		}
 
+		// Create/Remove git-daemon-export-ok for git-daemon...
+		daemonExportFile := path.Join(repo.RepoPath(), `git-daemon-export-ok`)
+		if repo.IsPrivate {
+			if err = os.Remove(daemonExportFile); err != nil {
```
Unknwon: Would this return error if the file does not exist?

bkcsoft: It will return a `*PathError` with the message `File does not exist`, but I'd rather have false positives than not getting `file can't be removed because permissions`-errors :unamused:

Unknwon: You can use `com.IsExist` to check first. `if repo.isprivate && com.isexist`, etc.

bkcsoft: that'd work, though checking if a file exists before removing it is bad practice... since it can be added/removed in between the check and the removal...

also this would break
```
if repo.IsPrivate && com.IsExist {
 [...]
} else { // if file doesn't exist, this block is ran even if it's private...
  os.Create(...)
}
```

So in that case I'd have to do:
```
if repo.IsPrivate {
  if com.IsExist {
    if os.Remove(...)
[...]
```
Which doesn't look very nice :sweat:

Unknwon: Two `else if`s.

bkcsoft: you mean like this?
```
if repo.isprivate && com.isexist {
 os.Remove
} else if !repo.isPrivate {
 os.Create
}
```

Unknwon:
```
if repo.isprivate && !com.isexist {
 if os.Create
} else if !repo.isPrivate && com.isexist {
 if os.Remove
}
```

The `isPrivate` is the latest value, just realized you did it oppositely?

bkcsoft: The file is created if it's **public** so the inner blocks should be reversed.
```
if repo.IsPrivate && com.IsExist {
  if os.Remove
} else if !repo.IsPrivate && !com.IsExist {
  if os.Create
}
```
but otherwise I "agree" that we need `com.IsExist` on both... (don't really agree since it's bad practice, but fair enough)

bkcsoft: As for the whole "it throws errors if it already exists or is already removed", people shouldn't change files manually (why I'm adding this :trollface: ) so it shouldn't happen and therefore isn't realyl an issue... it's might throw errors on the first update, but you can just ignore those errors... Just my 2 cents

bkcsoft:
Fixed file-checking :smiley:

Unknwon:
Thanks!

PR #3048, opened 2016-05-02
Fix listing team members #3031

### LINKED ISSUE #3031: Non-admin user cannot list teams
- Gogs version (or commit ref):5ec8ef023057c8250ae9fab6a2fa26951eec848b
- Git version:
- Operating system:Linux
- Database:
  - [X] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
  - [ ] No
  - [X] Not relevant
- Log gist:

## Description

For an organization we added multiple teams and sorted users into them. When a regular user is logged in (regular as in non-Gogs-admin) and they open the origanization page, the "People" list is filled with all organization members, but the team list is always empty. Not even the teams, that the user is a member of, are listed.

As a user, I would expect to see the list of the teams I am a member of, at least.

bkcsoft:
Hopefully fixes #3031

### DIFF
--- a/models/org.go
+++ b/models/org.go
@@ -443,7 +443,7 @@ func RemoveOrgRepo(orgID, repoID int64) error {
 // that the user with the given userID has access to.
 func (org *User) GetUserRepositories(userID int64) (err error) {
 	teams := make([]*Team, 0, org.NumTeams)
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

Issue #3027, opened 2016-04-27
[Proposal] Units of Repositories


lunny:
I would like to see Gogs is modulable to resolve some questions. We can define one module I named it Unit. If you have more suitable name, please comment.

``` Go
type UnitType int

const (
    UnitCode UnitType = iota
    UnitIssues
    UnitPRs
    UnitCommits
    UnitReleases
    UnitWiki
    UnitSettings
)

type Unit struct {
    UnitType
    NameKey string
    URI string
    DescKey string
    Sequence int
}
```

Now we have 6 units. And of course we maybe have more units in future such as Pluse, Graphs just like github's.

``` Go
var (
    Units = map[UnitType]Unit{
        {
            UnitCode,
            "repo.code",
            "/",
            "repo.code_desc",
            0,
        },
        {
            UnitIssues,
            "repo.issues",
            "/issues",
            "repo.issues_desc",
            1,
        },
        {
            UnitPRs,
            "repo.pulls",
            "/pulls",
            "repo.pulls_desc",
            2,
        },
        {
            UnitCommits,
            "repo.commits",
            "/commits/master",
            "repo.commits_desc",
            3,
        },
        {
            UnitReleases,
            "repo.releases",
            "/releases",
            "repo.releases_desc",
            4,
        },
        {
            UnitWiki,
            "repo.wiki",
            "/wiki",
            "repo.wiki_desc",
            5,
        },
        {
            UnitSettings,
            "repo.settings",
            "/settings",
            "repo.settings_desc",
            6,
        }
    }
)
```

Then we can add one new field on Repository struct

``` Go
type Repository struct {
...
Units []UnitType `xorm:"json"`
...
}
```

One repository could be assigned one or more units when it was created. And the old repositories could be assgined all the current units. When user visit repository page, firstly we could show units according the repository.

That is the first step.

The second we can add units on Access and Organization, Team,

``` Go
type Access struct {
...
RepoID int64
UserID int64
Units []UnitType `xorm:"json"`
}
```

So for one person who visit repository it could only visit all allowed units. 
We can define Team units, so when someone be added to this team, it will copy the team's units to this user's repository access.

``` Go
type Team struct {
...
Units []UnitType `xorm:"json"`
...
```

I think the effect of this proposal is small. We can complete it on 2 to 3 commits.

bkcsoft:
This has been discussed at lengths over here https://github.com/gogits/gogs/issues/2438 . Might wanna continue there instead :)

[...]

PR #3049, opened 2016-05-02
Added Units for repo, access and team. First step of three on #3027.

lunny:
For the first step of #3027, we added units for repository, access and team. And it's completely compatible with old repos. And there are no any changes on User's UI.

The second, we can change repo's settings page to let user to choose which units admin to want. Code, Issues,  Pulls, Commits, Releases, Wiki, Settings. So the repos' home page will dynamically show the units according the repo's settings.

The last step, we will change repo's access and team's access page to dynamically show the units according the user's access.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ gogs.sublime-project
 gogs.sublime-workspace
 /release
 vendor
+.vscode
--- a/models/access.go
+++ b/models/access.go
@@ -55,6 +55,7 @@ type Access struct {
 	UserID int64 `xorm:"UNIQUE(s)"`
 	RepoID int64 `xorm:"UNIQUE(s)"`
 	Mode   AccessMode
+	Units  []UnitType `xorm:"json"`
 }
 
 func accessLevel(e Engine, u *User, repo *Repository) (AccessMode, error) {
--- a/models/access.go
+++ b/models/access.go
@@ -167,6 +167,7 @@ func (repo *Repository) refreshAccesses(e Engine, accessMap map[int64]AccessMode
 			UserID: userID,
 			RepoID: repo.ID,
 			Mode:   mode,
+			Units:  UnitTypes,
 		})
 	}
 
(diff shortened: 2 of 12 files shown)


Unknwon:
Thanks your PR!

But please keep `migration.go` in its own package... it's not a good thing to mix it with models as it could contain legacy code and `models` changes over time.

lunny:
Hi, I have to move `migration.go` to models folder. Because `migration.go` will reference some struct (for this PR is `Access`, `Team`, `Repository` and `UnitType`) defined on `models` and `models` will reference `migrate`. It will result in recycle reference.

Unknwon:
Make new independent copy of structs, there is no grantee the struct will remain the same in the future, therefore the migration would fail.

lunny:
OKay. Another question, I think migrations could be execute before or after `xorm.Sync2`.  So I have changed the migrations. Do you have any idea about that?

Unknwon:
What is the advantage doing migration before sync tables?

lunny:
In fact, all of the before migration is exeuted before sync tables. In this case, we need set the default units to some fields after sync tables (added some units field on team, access,repository).

Unknwon:
Hmm... I don't know what you mean..

lunny:
See https://github.com/gogits/gogs/blob/master/models/models.go#L183
```Go
func NewEngine() (err error) {
	if err = SetEngine(); err != nil {
		return err
	}

	if err = migrations.Migrate(x); err != nil {
		return fmt.Errorf("migrate: %v", err)
	}

	if err = x.StoreEngine("InnoDB").Sync2(tables...); err != nil {
		return fmt.Errorf("sync database struct error: %v\n", err)
	}

	return nil
}
```
Migration will be executed before sync2.

Unknwon:
Yes, but why is that?

Unknwon:
OK... I got it. What is the problem right now?

lunny:
I will move back migrations to sub folder and PR again.

lunny:
I sent a new PR #3336 since the codes changed a lot. Please code review that. @Unknwon I will close this PR.

Issue #1268, opened 2015-05-11
LDAP user sync


Lafriks:
Would be great if there was an option to sync (automatically create/enable/disable) users with LDAP on regular interval, ex daily.

Lafriks:
Any info on when or if it will be implemented, it is only thing that keeps me back to finally start using it in my team

Unknwon:
Thanks your feedback!

I'm not very familiar with LDAP, does it provide anyway to get all users to sync?

Lafriks:
Just as you query for one user attributes you can query multiple users that are in specific domain/group only without specific name. With same base query and filter like this for example (&(objectCategory=person)(objectClass=user)) but this filter should be as setting (if there is no one already)

Unknwon:
@Lafriks Thanks, how does the results would be look like?

cc @SergioBenitez

jaydio:
Basically all that needs to happen is to send a request to the ldap server using the **User Filter**. Once the result was received iterate through the list of users and disable all local user accounts that are are **not** present within in the list. As for updating/syncing local accounts with the information available in ldap; this should be triggered - (1) every time the user logs in (2) interval-based by sending the same query again, iterate through the list of users and then either create or update local user accounts as required for all available entries. For accounts handled through ldap I'd also declare the **email** and **user name** fields as read-only so the user won't be able to change these. This could be triggered using an additional boolean within the configuration (e.g. LDAP_DETAILS_READ_ONLY).

Unknwon:
@jaydio thanks your tip!

Lafriks:
Implemented in pull request #3062

venessaj:
Hi! I want to ask a question regarding this feature.

I have input all the mandatory fields on : admin setting > authentications for the LDAP simple auth, but whenever I tried to login with my ldap account, it always said user doesn't exists. Is there any way to sync the user from my ldap directory in gogs? need your help

PR #3062, opened 2016-05-05
For LDAP user sync #1268 

lafriks:
Still have to create cron job for regular interval sync

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -325,6 +325,11 @@ ARGS =
 RUN_AT_START = true
 SCHEDULE = @every 24h
 
+; Synchronize external user data
+[cron.sync_external_users]
+SCHEDULE = @every 24h
+UPDATE_EXISTING = true
+
 [git]
 MAX_GIT_DIFF_LINES = 10000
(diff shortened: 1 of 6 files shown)


Lafriks:
Created new cron job with option to update existing user data or just creating new users.
Probably one thing to think about if option not to update users is set probably it would still be needed to deactivate users that are not in LDAP server anymore. Currently it only deactivates users if UpdateUsers option is set to true

ptman:
Have you thought about updating data on login?

Lafriks:
Do you mean updating only current users name/surname/email that is logging in or all users?

ptman:
Only current

barracks510:
Can you incorporate an option to disable the removal of old users?

Lafriks:
Old users are not removed, just disabled

Lafriks:
Any progress on merging this?

lafriks:
?

Unknwon:
image

strk:
@ptman would you be up to port this to Gitea ?

lafriks:
Pull request have been ported to Gitea, so if you are interested in it get latest changes from there

Issue #3077, opened 2016-05-10
when use https protocol ,it cann't trigger the webhook


everhopingandwaiting:
- Gogs version (or commit ref):
- Git version: 0.9.13.0318
- Operating system:  rpi2
- Database:
  -  SQLite

- Log gist:
```
2016/04/27 07:30:52 [...ts/gogs/cmd/serve.go:128 handleUpdateTask()] [E] Fail to trigger task: Head http://localhost:3000/graduation_project/lib-springmvc-mybatis/tasks/trigger?branch=new_UI&secret=1870c418ca91dab8b95c6595437ba9e7: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
2016/04/27 16:16:05 [T] Trigger task: http://localhost:3000/Huazai/huazai/tasks/trigger?branch=master&secret=4fa94abacc392120570cfdfba6bf9748
2016/04/27 16:16:05 [...ts/gogs/cmd/serve.go:128 handleUpdateTask()] [E] Fail to trigger task: Head http://localhost:3000/Huazai/huazai/tasks/trigger?branch=master&secret=4fa94abacc392120570cfdfba6bf9748: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
2016/04/29 12:35:47 [T] Trigger task: http://localhost:3000/SoLoRhythm/lib-springmvc-mybatis/tasks/trigger?branch=new_UI&secret=d87a0ec885a376c7caebfd29e9327da0
2016/04/29 12:35:48 [...ts/gogs/cmd/serve.go:128 handleUpdateTask()] [E] Fail to trigger task: Head http://localhost:3000/SoLoRhythm/lib-springmvc-mybatis/tasks/trigger?branch=new_UI&secret=d87a0ec885a376c7caebfd29e9327da0: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
2016/05/10 09:50:33 [T] Trigger task: http://localhost:3000/graduation_project/doctohtml/tasks/trigger?branch=master&secret=1870c418ca91dab8b95c6595437ba9e7
2016/05/10 09:50:33 [...ts/gogs/cmd/serve.go:128 handleUpdateTask()] [E] Fail to trigger task: Head http://localhost:3000/graduation_project/doctohtml/tasks/trigger?branch=master&secret=1870c418ca91dab8b95c6595437ba9e7: malformed HTTP response "\x15\x03\x01\x00\x02\x02"
[...]

PR #3078, opened 2016-05-10
  change  setting.go -> LocalUrl default value  to MustString(string(Protocol) + "://localhost:" + HttpPort + "/")

everhopingandwaiting:
change  setting.go -> LocalUrl  value default to  `sec.Key("LOCAL_ROOT_URL").MustString(Protocol + HttpAddr + HttpPort + "/") `  

###  when lost to congfig the app.ini file, if user use https protocol , webhook can not work !

issue #3077  

### DIFF
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -331,7 +331,7 @@ func NewContext() {
 	Domain = sec.Key("DOMAIN").MustString("localhost")
 	HttpAddr = sec.Key("HTTP_ADDR").MustString("0.0.0.0")
 	HttpPort = sec.Key("HTTP_PORT").MustString("3000")
-	LocalURL = sec.Key("LOCAL_ROOT_URL").MustString("http://localhost:" + HttpPort + "/")
+	LocalURL = sec.Key("LOCAL_ROOT_URL").MustString(string(Protocol) + "://localhost:" + HttpPort + "/")
 	OfflineMode = sec.Key("OFFLINE_MODE").MustBool()
 	DisableRouterLog = sec.Key("DISABLE_ROUTER_LOG").MustBool()
 	StaticRootPath = sec.Key("STATIC_ROOT_PATH").MustString(workDir)


modules/setting/setting.go
```diff
@@ -320,17 +322,20 @@ func NewContext() {
...
 		KeyFile = sec.Key("KEY_FILE").String()
 	} else if sec.Key("PROTOCOL").String() == "fcgi" {
 		Protocol = FCGI
+		protocol = "fcgi"
 	}
 	Domain = sec.Key("DOMAIN").MustString("localhost")
 	HttpAddr = sec.Key("HTTP_ADDR").MustString("0.0.0.0")
 	HttpPort = sec.Key("HTTP_PORT").MustString("3000")
-	LocalURL = sec.Key("LOCAL_ROOT_URL").MustString(Protocol + HttpAddr + HttpPort + "/")
+	LocalURL = sec.Key("LOCAL_ROOT_URL").MustString( protocol + HttpAddr + HttpPort + "/")
```
Lafriks: I think instead of HttpAddr you should use Domain as LocalUrl with all default values will be http://0.0.0.0:3000/ and that is not valid url

everhopingandwaiting: oh,  I  known , final  I  did  this  changes for  those  code ;
```
protocol = "http"
 	if sec.Key("PROTOCOL").String() == "https" {
 		Protocol = HTTPS
+		protocol = "https"
 		CertFile = sec.Key("CERT_FILE").String()
 		KeyFile = sec.Key("KEY_FILE").String()
 	} else if sec.Key("PROTOCOL").String() == "fcgi" {
 		Protocol = FCGI
+		protocol = "fcgi"
 	}
 	Domain = sec.Key("DOMAIN").MustString("localhost")
 	HttpAddr = sec.Key("HTTP_ADDR").MustString("0.0.0.0")
 	HttpPort = sec.Key("HTTP_PORT").MustString("3000")
-	LocalURL = sec.Key("LOCAL_ROOT_URL").MustString("http://localhost:" + HttpPort + "/")
+	LocalURL = sec.Key("LOCAL_ROOT_URL").MustString( protocol + HttpAddr + HttpPort + "/")
 	
```

Lafriks: But it should be:
LocalURL = sec.Key("LOCAL_ROOT_URL").MustString(protocol + Domain + HttpPort + "/")

Also I don't think fcgi:// as protocol will be correct and usable

everhopingandwaiting: Thanks  for  you  advice ,    this  fcgi  also  I  do not  known  how  it  works!  then   I  will  change the HttpPort  to  Domain.

everhopingandwaiting:
@Lafriks  took your advice!

Unknwon:
Thanks your PR!

We used to use `Domain`, but that's not always the case.

In terms of the error you have right now, I think you should use:

```go
string(Protocol) + "://localhost:" + HttpPort + "/"
```

Also `protocol` is not needed.

everhopingandwaiting:
@Unknwon   yes  sir

Unknwon:
Please,

1. Confirm it is working right now.
2. Clean up to 1 commit.

bkcsoft:
This won't work right? As `LOCAL_ROOT_URL` is set in `conf/app.ini` hence will not fallback to `.Must()` (if I've understood it correctly). So `conf/app.ini` has to be updated as well.

Unknwon:
Ping @everhopingandwaiting

everhopingandwaiting:
alao i had reported it on discus  two month ago,haha

PR #3079, opened 2016-05-11
Update docker/build.sh script to use glide & make

0rax:
- docker/build.sh will now use glide to fetch dependencies
- glide is built from source to keep compatibility with arm (no pre-prebuilt binary for arm)
- docker/build.sh will also now use the provided Makefile. 
  It will generate an error when trying to get git build hash as we do
  not ship the 88mo .git directory during the build (should not cause
  any problem as the variable it sets was not set previously)
- Updated gosu to 1.9 in `Dockerfile` & `Dockerfile.rpi`
- Tested on OSX (Docker for Mac), Raspberry Pi 2 (ArchLinux) and Raspberry Pi 3 (HypriotOS / Raspbian Jessie)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,5 @@
 .git
 .git/**
-conf
-conf/**
 packager
 packager/**
 scripts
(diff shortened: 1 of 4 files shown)


Unknwon:
Thanks!

Issue #3069, opened 2016-05-06
"gogs dump" should take a destination argument


majewsky:
```
$ sudo -u gogs -g gogs /usr/share/gogs/gogs dump --config /srv/gogs/conf/app.ini
2016/05/06 18:02:03 Creating tmp work dir: /tmp/gogs-dump-187221175
2016/05/06 18:02:03 Dumping local repositories.../srv/gogs/repo
2016/05/06 18:02:47 Fail to dump local repositories: write /tmp/gogs-dump-187221175/gogs-repo.zip: no space left on device
```

`/tmp` is usually a tmpfs on modern distributions and thus limited to 50% the size of the RAM, which in my case is just 0.5 GiB. A `--output/-o` argument should be allowed to specify a different output directory.

bkcsoft:
what you're asking for is a `temp directory argument`, `--output/-o` would be where `gogs-dump-<date>.zip` ends up :smile:

msoedov:
Sorry a bit late response. I agree with @bkcsoft comments on the PR that `--tempdir/-t` would be more accurate.

Unknwon:
This is claimed to be fixed by merging #3086, please help test on `develop` branch.

PR #3086, opened 2016-05-12
Add output flag for  `gogs dump`

msoedov:
#3069

### DIFF
--- a/cmd/dump.go
+++ b/cmd/dump.go
@@ -28,7 +28,8 @@ It can be used for backup and capture Gogs server image to send to maintainer`,
 	Action: runDump,
 	Flags: []cli.Flag{
 		stringFlag("config, c", "custom/conf/app.ini", "Custom configuration file path"),
-		boolFlag("verbose, v", "show process details"),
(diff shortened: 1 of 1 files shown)


cmd/dump.go
```diff
@@ -28,7 +28,8 @@ It can be used for backup and capture Gogs server image to send to maintainer`,
 	Action: runDump,
 	Flags: []cli.Flag{
 		stringFlag("config, c", "custom/conf/app.ini", "Custom configuration file path"),
-		boolFlag("verbose, v", "show process details"),
+		boolFlag("verbose, v", "Show process details"),
+		stringFlag("output, o", os.TempDir(), "Destination dir path"),
```
bkcsoft: `tempdir, t` makes more sense since that's what you're actually changing
(not sure if this needs a short-hand version though...) @Unknwon

Unknwon:
Thanks for the PR!

Any reasons that you need to specify temp dir by yourself?

bkcsoft:
@Unknwon because `/tmp` is to small on some systems :)

Unknwon:
OK, make sense!

PR #3093, opened 2016-05-13
Add comment note

andreynering:
As asked here: https://github.com/gogits/gogs/issues/1738#issuecomment-218917124

/cc @Unknwon


Unknwon:
Thanks!

PR #3095, opened 2016-05-13
Added webkit-font-smoothing

deanpcmad:
Here I've added webkit font smoothing to make it look nicer on Mac's

Before:

After:


bkcsoft:
Stylesheet changes is done in .less-files, .css is generated

deanpcmad:
Ah ok, I'll close this and re-open it after changing the less files

PR #3104, opened 2016-05-19
Add ServerName to tls.Config in LDAP auth

ptman:
From https://godoc.org/crypto/tls#Config

    ServerName is used to verify the hostname on the returned
    certificates unless InsecureSkipVerify is given. It is also included
    in the client's handshake to support virtual hosting unless it is
    an IP address.

This is needed for certificate validation without InsecureSkipVerify.


ptman:
@bkcsoft @Unknwon Trivial change, easy to merge.

Unknwon:
Thanks!

PR #3109, opened 2016-05-20
[WIP] Restore command

bkcsoft:
This PR adds a `gogs restore`-command for restoring from `gogs dump`.

Currently a WIP since it does **no sanity-checking what so ever**. Using this on an already installed instance **will break it**. Running it on a clean instance should work just fine (I'm using it to migrate servers).

Things that needs to be done:
- [ ] Sanity-checks (e.g. 'do you _really_ want to override your entire database?')
- [ ] Fix datetime-bug

Nice things to have?
- [ ] Partial restores? (e.g. only database for when mysql corrupts your files *again*)
- [ ] Side-loading configfiles? (e.g. migrating between database-backends)

Please suggest more things to the list, since I've probably forgotten something :laughing: 

EDIT: Just noticed that it doesn't restore datetime-values (effectively setting them to `0001-01-01T00:00:00Z` :unamused: )

### DIFF
--- a/cmd/restore.go
+++ b/cmd/restore.go
@@ -26,7 +26,7 @@ It can be used for backup and capture Gogs server image to send to maintainer`,
 	},
 }
 
-func runRestore(ctx *cli.Context) {
+func runRestore(ctx *cli.Context) error {
 	if !ctx.IsSet("file") {
 		log.Fatalf("No dump-file set...")
 	}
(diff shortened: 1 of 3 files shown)


bkcsoft:
@Unknwon Why did you close this one?

Unknwon:
I think it is not worth investigating anymore?

amb-it-ion:
Why not? Is there an alternative restore command?

Unknwon:
@amb-it-ion currently Gogs does not have agreement on the format that is dumped, so would be hard to have a restore command at current stage.

rambabusaravanan:
@Unknwon so you mean, there is no way to restore gogs.
If so, it is impossible to migrate gogs from one machine to another machine?

PR #3115, opened 2016-05-21
Add support for federated avatars

### LINKED ISSUE #3105: Add support for federated avatars
See https://www.libravatar.org/

strk:
Fixes #3105

The code would be better cleaned up by someone else.
There's a duplication (which was already in the code before my intervention) between modules/base/tools.go and modules/user.go

The libravatar module is my first module. Not so big so it could eventually be included, if needed.


strk:
I just found out that in the "profile" page the big avatar is a link to "change your avatar" pointing at Gravatar, that part should also be changed to be dependent on configuration.

strk:
With the latest two commits I made "libravatar" the default GravatarSource.
libravatar automatically falls back to Gravatar so avatars would be by default fetched in this order:

 1. Federated avatar (via DNS lookup)
 2. libravatar.org account 
 3. gravatar host

Those who want to use "duoshuo" can still do that, in which case the order would be:

 1. Federated avatar (via DNS lookup)
 2. duoshuo account
 3. -- I dunno how dushuo works, but it's up to them to make the next move --

strk:
Travis failure for go 1.6 seems to have nothing to do with the PR

strk:
For the record, the duplication code is also gone now

strk:
rebased and removed bindata regeneration (which is in conflict with changes in develop branch, I think).

@Unknwon would it make sense to remove the bindata.go from the repository ?

strk:
With some testing I found that the code really needs some caching, or showing the explore/user page can jump up from 3ms to 1000ms...

strk:
So I'm experimenting with a DNS lookup cache within the Libravatar module and times are back down to 3ms for a view of a page full of users.

strk:
So with the new commit, and an updated strk.kbt.io/projects/go/libravatar module, a 1-day lookup cache is implemented for each domain, fixing the performance problem

strk:
I confirm even in production the time to render user page goes down from 14000 ms to 5ms

strk:
One thing I'd want to add to this PR is changing public/img/avatar_default.jpg to not be the gravatar one (why should it be?) but rather some gopher ...

strk:
And now a custom default avatar is also available, ain't it cute ? :)

I've derived it from the Gogs logo

https://github.com/strk/gogs/blob/55421ab42f24076adfca05cc22da661908abaa7f/public/img/avatar_default.png

bkcsoft:
@strk Nice job, few things though:
1. Locales are croud-sources from CrowdIn, only en_US should be changed in the repo :)
2. Why did you change the default to `libravatar` when you're seeing perforance hits? caching still requires it to be fetched once (with said hit)
3. How fast does libravatar notice a change of avatar with the caching turned on? (not entirely familiar with libravatar, only that it falls back on gravatar in some cases)

Lastly, since most people who have Gravatar-support turned on (like me) will continue to use gravatar, changing the default to libravarar (adding a slight latency hit) isn't very nice, even if it would still work as intended (just slightly slower) :confused:

strk:
> 1. Locales are croud-sources from CrowdIn, only en_US should be changed in the repo :)

I noticed the actual sources are constants, why is en_US special ?

> 2. Why did you change the default to `libravatar` when you're seeing perforance hits? caching still requires it to be fetched once (with said hit)

The lookup is always performed with the code in this PR.
GravatarSource only affects the fallback if the user email domain does not expose
an "avatar" service. If you're concerned about performance there should be a setting
to completely disable federated avatars.

I changed the default because libravatar.org, contrary to the previous gravatar.com
default, is a community effort, is the default for libravatar client libraries and
does contain instructions about how to run your own avatar service.

> 3. How fast does libravatar notice a change of avatar with the caching turned on? (not entirely familiar with libravatar, only that it falls back on gravatar in some cases)

Caching is hard-coded to be 1 day, as 1 day is the recommended minimum
for the TTL of the DNS record.

> Lastly, since most people who have Gravatar-support turned on (like me) will continue to use gravatar, changing the default to libravarar (adding a slight latency hit) isn't very nice, even if it would still work as intended (just slightly slower) :confused: 

What Libre Avatar (libravatar) adds, is the ability for users to host
their own avatars. Try setting "strk@kbt.io" as your avatar email and
see the effects of it (it won't hit libravatar.org at all!).

bkcsoft:
> I noticed the actual sources are constants, why is en_US special ?

https://github.com/gogits/gogs/wiki/Contributing-Code#those-we-do-not-accept

> What Libre Avatar (libravatar) adds, is the ability for users to host their own avatars. Try setting "strk@kbt.io" as your avatar email and see the effects of it (it won't hit libravatar.org at all!).

Try setting it to `kim.carlbacker@gmail.com` and see if there's a performance or time penalty :) I (and most likely anyone currently using Gravatar) will not move to another avatar-source, since every other service out there (that supports it) is using Gravatar.  
And I for one can't change the DNS-record for gmail.com :laughing: 

> Caching is hard-coded to be 1 day, as 1 day is the recommended minimum
for the TTL of the DNS record.

Having worked with DNS record to some extend, I do know that the TTL can be fetched with the result, wouldn't this be a better way? For when the record doesn't exist, you can default it to 1 day.

> The lookup is always performed with the code in this PR. GravatarSource only affects the fallback if the user email domain does not expose an "avatar" service. If you're concerned about performance there should be a setting to completely disable federated avatars.

This was my point. Changing the default, to something that for most people will add a penalty, isn't very user-friendly :smile: 

> I changed the default because libravatar.org, contrary to the previous gravatar.com default, is a community effort, is the default for libravatar client libraries and does contain instructions about how to run your own avatar service.

I'm not contesting the use (or value) of Libre Avatar (seems awesome, and I like decentralized things :smiley:), just the change of the defaults. Changing defaults isn't considered very nice since it can possibly break things, or add penalties, for people using defaults :unamused: 

In a perfect world, Gogs would have a statistics service where we can see (opt-in and completely anonymous of course) what settings people use. Then we can easier see what would break (or slow down), by changing defaults. :smiley:

strk:
> https://github.com/gogits/gogs/wiki/Contributing-Code#those-we-do-not-accept

Ok, force-pushed with the non-en_US locale files untouched.

> Try setting it to `kim.carlbacker@gmail.com` and see if there's a performance or time penalty :) I (and most likely anyone currently using Gravatar) will not move to another avatar-source, since every other service out there (that supports it) is using Gravatar.  

Everyone is also using github, why bothering with Gogs ?

> And I for one can't change the DNS-record for gmail.com :laughing: 

Yep, that's a problem, and the reason why libravatar.org will fallback
to gravatar.com for those which are unknown.

The time penalty is one DNS lookup per day, for the google.com SRV record.
I guess it could be removed by making each user decide how to fetch
the avatar (local, fixed url, via service).

> Having worked with DNS record to some extend, I do know that the TTL can be fetched with the result, wouldn't this be a better way? For when the record doesn't exist, you can default it to 1 day.

To be honest I hadn't found a way to query the SOA, if you want to help,
the cache is in strk.kbt.io/projects/go/libravatar, send a PR ! :)

I might try again in some future time, but it'll be outside of Gogs
(and inside the library).

> This was my point. Changing the default, to something that for most people will add a penalty, isn't very user-friendly :smile: 

I don't like free software default pointing to non-free software services.
And it's particularly annoying for that non-free software service to be referenced in hard-coded strings/templates (and even in the default_avatar.jpg -- also fixed by this PR).

Anyway, as long as there's an option to opt-out of pushing users to a non-free service
I'll be happy. Right now doing that would have required refactoring the code to
avoid the hard-coded strings, so I decided to rather change the hard-coded strings
to point to the free replacement instead.

Unknwon:
Thanks for the PR!

I agree with @bkcsoft which default behavior should not be changed unless there is a serious bug or concern.

strk:
I think we should distinguish between the "federated avatar" support
(which enables or not the DNS lookup) and the "avatar source".

The rationale for using libravatar.org as a default "avatar source"
would be that of not pushing users toward a proprietary service.
It would add NO overhead on the server side, but might result
in a double request from the browser IFF the the has of a given email
is not known by libravatar.org (one request to libravatar.org, another
on being redirected by libravatar.org to gravatar.com).
Sending users to libravatar.org for "updating their avatar" would be a way
to make users aware of the project and of the possibility they have to
serve their own avatar.

Would you be ok to change the default "avatar source", independently
of the "federated avatar" support ?

My main problem with this is that gravatar.com things are hard-coded
in the sources (templates, and default avatar). If they weren't, I'd
be ok with just changing the setting myself, but if I just change
the setting from app.ini to use "libravatar" while the "update your avatar"
text in template still send users to "gravatar.com" it is not a good
service for the libre world :)

Unknwon:
At any rate, I think everything about `libravatar` should have its own settings, and an extra one for enable `libravatar`.

strk:
At the bare minimum we'd need a setting to enable or disable
federated avatars (the DNS lookup).

That, or turn avatars providing into an interface and only decide
which implementation to use from the setting (via "nickname")

Unknwon:
> That, or turn avatars providing into an interface and only decide
which implementation to use from the setting (via "nickname")

Sounds better to me.

strk:
On May 30, 2016 2:51:32 PM CEST, "无闻" <notifications@github.com> wrote:
>> That, or turn avatars providing into an interface and only decide
>which implementation to use from the setting (via "nickname")
>
>Sounds better to me.
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/gogits/gogs/pull/3115#issuecomment-222486024

Do you think you could draft such interface ? Would the default avatar still be ok as a gopher ? It would be nice to merge the agreed portions until a more structured approach is ready (I don't know when I'll have time for this)
--strk;

Unknwon:
1. After a rethink, make a new interface might be overkill. Because all we need is allowing users set another special value `libravatar` and handles correspondingly.
2. Could you put the code `strk.kbt.io/projects/go/libravatar` on GitHub? Custom domain feels unstable to me.
3. Change default image is not quite what I perfer...

strk:
On Wed, Jun 01, 2016 at 01:11:15AM -0700, 无闻 wrote:
> 1. After a rethink, make a new interface might be overkill. Because all we need is allowing users set another special value `libravatar` and handles correspondingly.

In the GravatarSource, you mean ?
My PR is currently mapping 'libravatar' simply as
https://cdn-sec.libravatar.org, with no indication about
whether federation is active or not.

Federation DNS lookup only really makese sense as a default so that
new users just magically see their avatar pop up w/out having to
express any special preference. If they have to express a preference
then why not directly entering the URL to their preferred avatar ?
It would take no DNS lookup even.

Or by "users" you meant "instance administrators" ?

> 2. Could you put the code `strk.kbt.io/projects/go/libravatar` on GitHub? Custom domain feels unstable to me.

That goes in the opposite direction of "federation" :)
I'm appreciating the fact that "golang" clones dependent
projects locally, making "stability" a non-issue after
the initial cloning. 

Besides, the "import url" just needs a metadata in the HTML
pointing to the actual repository, so you could as well use
your own proxy to it, if you prefer (check out the html
of strk.kbt.io/projects/go/libravatar).

> 3. Change default image is not quite what I perfer...

What would you prefer ? I hope you agree that the gravatar.com logo 
is an inappropriate default image ?

Unknwon:
>If they have to express a preference
then why not directly entering the URL to their preferred avatar ?
It would take no DNS lookup even.

If users/admins choose to set `GRAVATAR_SOURCE = libravatar`, then it is the same set it as `https://cdn-sec.libravatar.org` , or then can simply give a more specific value.

>That goes in the opposite direction of "federation" :)
I'm appreciating the fact that "golang" clones dependent
projects locally, making "stability" a non-issue after
the initial cloning.

>Besides, the "import url" just needs a metadata in the HTML
pointing to the actual repository, so you could as well use
your own proxy to it, if you prefer (check out the html
of strk.kbt.io/projects/go/libravatar).

I'm afraid I would not accept PR based on that...

>What would you prefer ? I hope you agree that the gravatar.com logo
is an inappropriate default image ?

I do think gravatar.com logo is more appropriate default image than gopher in this case.

strk:
I'm not sure I'm understanding what your suggestion is being, with reguard to the libravatar setting. Isn't it what's being done here: 
https://github.com/strk/gogs/blob/2abecb47c959b70042042b30a2821e4241310679/modules/setting/setting.go#L456

About import path, is it github.com or nothing ? Or is it specifically against my own domain ?
Would gopkg.in also do, for example ?

What makes gravatar.com logo more appropriate as a default avatar image than a gopher ?
Would libravatar.org logo as good as the gravatar.com one ?

Unknwon:
>https://github.com/gogits/gogs/pull/3115/files#diff-60f5fca6d29eafbb30eb645a0cd4041fR468

This part I don't understand what's for? 

>About import path, is it github.com or nothing ? Or is it specifically against my own domain ?
Would gopkg.in also do, for example ?

gopkg.in eventually redirect to GitHub.

strk:
On Wed, Jun 01, 2016 at 03:12:37PM -0700, 无闻 wrote:
> >https://github.com/gogits/gogs/pull/3115/files#diff-60f5fca6d29eafbb30eb645a0cd4041fR468
> This part I don't understand what's for? 

It attempts to extrapolate Libravatar configuration from
the GRAVATAR_SOURCE setting. Basically an hack to avoid
adding more configuration settings (which would call for encoding
the specific setting in some form of dynamic object).

The extracted settings are:

 1. Whether to use HTTP or HTTPS for fetching the image
 2. Which base url to use if DNS lookup finds no "avatar" service
    (ie: what to use while not federated)

Note that the the fallback url is not a suggested setting from
the libravatar.org website. I've added support for it to do allow
people using "duoshuo" (for example) to never hit libravatar.org or
gravatar.com but directly duoshuo.com when DNS lookup yelds no
domain support for federated avatars.

> >About import path, is it github.com or nothing ? Or is it specifically against my own domain ?
> Would gopkg.in also do, for example ?
> 
> gopkg.in eventually redirect to GitHub.

Ok, wrong example :)
How about golang.org ?
What makes a domain "stable" or not, to you ?

--strk;

Unknwon:
Thanks for the explanation!

I think this is all we need to in the settings: https://github.com/gogits/gogs/pull/3115/files#diff-60f5fca6d29eafbb30eb645a0cd4041fR459

And a more general description on locale file: https://github.com/gogits/gogs/pull/3115/files#diff-514b45a5b35ffc3c85b9e04ed2e0c649L223

Maybe `Change your avatar on your service website`?

Unknwon:
>Ok, wrong example :)
How about golang.org ?
What makes a domain "stable" or not, to you ?

golang.org down? It basically means you can't even install Go, and Gogs, and everything.

strk:
On Thu, Jun 02, 2016 at 12:17:53PM -0700, 无闻 wrote:
> Thanks for the explanation!
> 
> I think this is all we need to in the settings: https://github.com/gogits/gogs/pull/3115/files#diff-60f5fca6d29eafbb30eb645a0cd4041fR459

Sorry, I don't understand what you're suggesting. Are you saying we
just add "libravatar" as an indication of admin will to use
the federated avatar service, while not changing the default ?

This would also mean tightly coupling "federated avatar"
with libravatar.org, right ?

> And a more general description on locale file: https://github.com/gogits/gogs/pull/3115/files#diff-514b45a5b35ffc3c85b9e04ed2e0c649L223
> Maybe `Change your avatar on your service website`?

And would link where ?
Currently templates/user/profile.tmpl has an hard-coded target
url (gravatar.com in master branch, libravatar.org in my branch).

--strk;

strk:
On Thu, Jun 02, 2016 at 12:18:27PM -0700, 无闻 wrote:

> golang.org down? It basically means you can't even install Go, and Gogs, and everything.

Alright, another suboptimal example.
What about gnu.org, would you consider that stable ?
And goo.gl ?
And what about gogs.io ? (I actually wonder why gogs isn't go-get'able
from gogs.io ...)

Also, is it availability of the import path that concerns
you or of the git repository ?

--strk;

Unknwon:
>Sorry, I don't understand what you're suggesting. Are you saying we
just add "libravatar" as an indication of admin will to use
the federated avatar service, while not changing the default ?
>This would also mean tightly coupling "federated avatar"
with libravatar.org, right ?

Yes and yes.

> And would link where ?

Maybe should Just remove the tooltip while not using Gogs's own avatar system.

>...

No wishes on continuing that topic.

strk:
I'd rather add a setting to "Enable Federated Avatar Lookup" completely independent of the other settings, taking care of the lookup. It's really a oneliner, btw, doesn't need an external library (so we also get past the kbt.io hate party).

Unknwon:
>I'd rather add a setting to "Enable Federated Avatar Lookup" completely independent of the other settings, taking care of the lookup

OK, make sense...

strk:
I might give that a try (not sure when) but see also #3164 as an helpful step in that direction.

strk:
BTW, I've pushed my library to try.gogs.io/strk/go-libravatar, so it's closer to the main code :P

strk:
@Unknwon I've tried adding the "Use Federated Avatar" checkbox but got stuck on getting it honoured (just the checkbox, no code using it yet). Could you help with that ?

strk:
So as of 70a837e7c87313452d179079666df57b724c67ad the checkbox is effectively giving the option of using federated avatar lookup to each user on the system. There is no global option (there should probably be one, for organizations), and it is possible for a user to set both "Use federated avatar" and "Use custom avatar" which are semantically mutually exclusive instead...

strk:
@Unknwon what do you think about the approach ? Does it make sense for the setting to be per-user instead of systemwide ?

strk:
With latest commits the "AvatarForm" got mutually exclusive choice between "Lookup Avatar by Email" and "Use Custom Avatar". What's left ?

models/user.go
```diff
@@ -260,7 +261,17 @@ func (u *User) RelAvatarLink() string {
 
 		return "/avatars/" + com.ToStr(u.Id)
 	}
-	return setting.GravatarSource + u.Avatar
+
+	var url string
+
```
Unknwon: Remove empty line here.

models/user.go
```diff
@@ -260,7 +261,17 @@ func (u *User) RelAvatarLink() string {
...
+	if ( u.UseFederatedAvatar && setting.LibravatarService != nil ) {
+		url, _ = setting.LibravatarService.FromEmail(u.AvatarEmail)
+	}
+	if url == "" {
+			return base.AvatarLink(u.AvatarEmail)
```
Unknwon: Have you ran `gofmt`?

modules/setting/setting.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/gogits/gogs/modules/bindata"
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/user"
+
+	"strk.kbt.io/projects/go/libravatar"
```
Unknwon: Please host to a GitHub repo.

modules/setting/setting.go
```diff
@@ -460,6 +463,17 @@ func NewContext() {
 	if OfflineMode {
 		DisableGravatar = true
 	}
+	if ( ! DisableGravatar ) {
+		LibravatarService = libravatar.New()
+		// TODO: initialize libravatar just once, at configuration time
```
Unknwon: What does this `TODO` suppose to do?

strk:
@Unknwon this was squash-rebased against develop branch. It's a single commit now reworking the user setting page to have mutually-exclusive  Custom/ByEmail avatar handling and adding a "federated" checkbox in the "ByEmail" side. Open for review.

routers/user/setting.go
```diff
@@ -92,10 +92,6 @@ func SettingsPost(ctx *context.Context, form auth.UpdateProfileForm) {
 	ctx.User.Email = form.Email
 	ctx.User.Website = form.Website
 	ctx.User.Location = form.Location
-	if len(form.Gravatar) > 0 {
```
Unknwon: Why remove this?

modules/auth/user_form.go
```diff
@@ -93,19 +93,25 @@ type UpdateProfileForm struct {
...
 
 func (f *UpdateProfileForm) Validate(ctx *macaron.Context, errs binding.Errors) binding.Errors {
 	return validate(errs, ctx.Data, f, ctx.Locale)
 }
 
-type UploadAvatarForm struct {
-	Enable bool
+const (
+	AVATAR_LOCAL string = "local"
+	AVATAR_BYMAIL string = "bymail"
```
Unknwon: Seems never been used?

Unknwon:
Please attach a screenshot to new user settings page.

models/user.go
strk: On Mon, Jun 27, 2016 at 02:34:13AM -0700, 无闻 wrote:
> > @@ -260,7 +261,17 @@ func (u *User) RelAvatarLink() string {
> >  
> >  		return "/avatars/" + com.ToStr(u.Id)
> >  	}
> > -	return setting.GravatarSource + u.Avatar
> > +
> > +	var url string
> > +
> > +	if ( u.UseFederatedAvatar && setting.LibravatarService != nil ) {
> > +		url, _ = setting.LibravatarService.FromEmail(u.AvatarEmail)
> > +	}
> > +	if url == "" {
> > +			return base.AvatarLink(u.AvatarEmail)
> 
> Have you ran `gofmt`?

Yes. Does it not work on already-committed code ?

routers/user/setting.go
strk: On Mon, Jun 27, 2016 at 02:36:08AM -0700, 无闻 wrote:
> > @@ -92,10 +92,6 @@ func SettingsPost(ctx *context.Context, form auth.UpdateProfileForm) {
> >  	ctx.User.Email = form.Email
> >  	ctx.User.Website = form.Website
> >  	ctx.User.Location = form.Location
> > -	if len(form.Gravatar) > 0 {
> 
> Why remove this?

The avatar settings are in a different form, not
in Update.ProfileForm anymore

modules/auth/user_form.go
strk: On Mon, Jun 27, 2016 at 02:36:35AM -0700, 无闻 wrote:
> >  }
> >  
> >  func (f *UpdateProfileForm) Validate(ctx *macaron.Context, errs binding.Errors) binding.Errors {
> >  	return validate(errs, ctx.Data, f, ctx.Locale)
> >  }
> >  
> > -type UploadAvatarForm struct {
> > -	Enable bool
> > +const (
> > +	AVATAR_LOCAL string = "local"
> > +	AVATAR_BYMAIL string = "bymail"
> 
> Seems never been used?

routers/org/setting.go: form.Source = auth.AVATAR_LOCAL

AVATAR_BYMAIL is currently never used as !AVATAR_LOCAL
is currently enough to know it's AVATAR_BYMAIL instead,
being there only 2 options. In case this grows, there
could be more values (for example when looking up avatar
by URL instead, via OpenID).

--strk;

strk:
TODO item and newspace removed. Rebased to develop branch. Here's the screenshot:

screenshot2

About import path: I don't intend to maintain a mirror on github, but gitlab (contrary to github) supports an automated mirroring tool, so if a corporation URL is really needed for you I could setup a mirror on gitlab.com. Would that help you ?

models/user.go
strk: On Mon, Jun 27, 2016 at 02:33:45AM -0700, 无闻 wrote:
> > @@ -260,7 +261,17 @@ func (u *User) RelAvatarLink() string {
> >  
> >  		return "/avatars/" + com.ToStr(u.Id)
> >  	}
> > -	return setting.GravatarSource + u.Avatar
> > +
> > +	var url string
> > +
> 
> Remove empty line here.

Done

modules/setting/setting.go
strk: On Mon, Jun 27, 2016 at 02:34:36AM -0700, 无闻 wrote:
> > @@ -25,6 +25,8 @@ import (
> >  	"github.com/gogits/gogs/modules/bindata"
> >  	"github.com/gogits/gogs/modules/log"
> >  	"github.com/gogits/gogs/modules/user"
> > +
> > +	"strk.kbt.io/projects/go/libravatar"
> 
> Please host to a GitHub repo.

Would a gitlab mirror do ?
go get gitlab.com/strk/go-libravatar

Only downside is it'll be only updated hourly

--strk;

modules/setting/setting.go
strk: On Mon, Jun 27, 2016 at 02:35:05AM -0700, 无闻 wrote:
> > @@ -460,6 +463,17 @@ func NewContext() {
> >  	if OfflineMode {
> >  		DisableGravatar = true
> >  	}
> > +	if ( ! DisableGravatar ) {
> > +		LibravatarService = libravatar.New()
> > +		// TODO: initialize libravatar just once, at configuration time
> 
> What does this `TODO` suppose to do?

Left-over, removed on squash-rebase

strk:
Looking at the screenshot, I think it could be worth putting the "avatar" form in its own page, accessible from a link in the left tab, under Profile. What do you think ?

models/user.go
Unknwon: The indentation does not look like it is output from `gofmt`.

modules/setting/setting.go
Unknwon: I'm sorry... but no...

strk: Are you really going to refuse using anything that's not on github.com ?
Or do you have a list of "whiltelisted" hosting places to pick from ?

Unknwon:
Basically GitHub and BitBucket.

models/user.go
strk: On Mon, Jun 27, 2016 at 06:00:21AM -0700, 无闻 wrote:
>
> The indentation does not look like it is output from `gofmt`.

You were right, I was just calling "go fmt" with wrong arguments.
Passing it each of the files I modified as arguments (multiple calls)
did change formatting. I've pushed the result.

Note that running `git fmt ./...` changes even more files, but those
I didn't push, not being files that were changed in my branch.

strk:
Latest commit adds go-libravatar under vendor/, to resolve the availability issue.
Previous commit redesign the form to have Avatar settings in their own form page.

Unknwon:
Said many times, this PR will not be accepted with current import path, nor put under `vendor`, that's for very purpose of vendor, and ignored in `.gitignore` by purpose currently.

strk:
I reverted the vendor/ inclusion, but was is exactly the "purpose of vendor" ?

strk:
With the latest commit the "federated avatar" setting became a systemwide setting, and per-user setting was removed. I think it makes more sense, given the DNS lookup overhead  concern. It defaults to off, and can be turned on at install time (or via custom/conf/app.ini using the FEDERATED_AVATAR setting)

strk:
@Unknwon : github import path is also in place now, ready to merge

modules/setting/setting.go
```diff
@@ -25,6 +25,16 @@ import (
...
 	"github.com/gogits/gogs/modules/user"
+
+	// The go-libravatar dependency is being
+	// imported from an unsupported github mirror
+	// due to Unknwon not trusting the stability
+	// of its official domain. See
+	// https://github.com/gogits/gogs/pull/3115#issuecomment-222922681
+	// Official import path is left commented-out for documentation
+	// purpose
+	//"strk.kbt.io/projects/go/libravatar"
```
Unknwon: You can put these comments in your GitHub description, but not here...

strk: > You can put these comments in your GitHub description, but not here...

Comments removed

modules/setting/setting.go
```diff
@@ -25,6 +25,9 @@ import (
 	"github.com/gogits/gogs/modules/bindata"
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/user"
+
+	// should be "strk.kbt.io/projects/go/libravatar"
+	"github.com/strk/go-libravatar"
```
Unknwon: No comments needed at all and please put after line 23 to group external imports together.

strk: Done

modules/setting/setting.go
jhasse: I think it's about minimizing possible points of failure.

strk: Vendoring would minimize them even further, w/out stamping a
proprietary software based service brand on top of free software
developers works

strk:
Rebased and made very sure nothing leaks about signs of life outside of github.com.
What's left to do here, for the PR to be accepted ?

strk:
Rebased again and re-tested. 
@bkcsoft could you give this a review ?

I'm attaching new screenshot after splitting the profile and avatar forms:

screenshot from 2016-07-21 19 14 48

And the admin panel, under configuration:

screenshot from 2016-07-21 19 16 45

Unknwon:
This thread is too long which is not ideal for communicate with inline comments, could you close this one and create a new PR?

strk:
Continues to #3320

PR #3121, opened 2016-05-23
Fixes #3006: Markdown misinterpreting issue URLs

### LINKED ISSUE #3006: Markdown stripping parameters from URLs when the URL contains the word 'issues'
- Gogs version (or commit ref): 0.9.13.0319
- Git version: 1.9.1
- Operating system: Ubuntu 14.04.4 LTS
- Database:
  - [x] PostgreSQL
  - [x] Yes https://try.gogs.io/ultralisk/ultratest/issues/2

## Description

I pasted a link into an issue comment which happened to have a similar structure to a gogs URL, gogs treated the URL like a reference link as well as stripped the parameters that the linked contained.

...

cstroe:
Markdown module is now more strict about what URLs it interprets as issue and commit URLs.

The main change is that it now checks to see if the URL starts with the `setting.AppUrl` before treating it as an issue or commit URL.

### DIFF
--- a/modules/markdown/markdown_test.go
+++ b/modules/markdown/markdown_test.go
@@ -1,24 +1,26 @@
 package markdown_test
 
 import (
-	"testing"
-	. "github.com/smartystreets/goconvey/convey"
 	. "github.com/gogits/gogs/modules/markdown"
+	. "github.com/smartystreets/goconvey/convey"
+	"testing"
 
+	"bytes"
 	"github.com/gogits/gogs/modules/setting"
+	"github.com/russross/blackfriday"
 )
 
 func TestMarkdown(t *testing.T) {
-	var (
-		urlPrefix = "/prefix"
-		metas map[string]string = nil
-	)
-	setting.AppSubUrlDepth = 0
+	Convey("Rendering an issue mention", t, func() {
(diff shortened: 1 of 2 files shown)


modules/markdown/markdown_test.go
```diff
@@ -50,18 +52,18 @@ func TestMarkdown(t *testing.T) {
 					So(string(RenderIssueIndexPattern([]byte(testCases[i]), urlPrefix, metas)), ShouldEqual, testCases[i+1])
 				}
 			})
-			Convey("It should not render issue without leading space", func () {
+			Convey("It should not render mention without leading space", func() {
```
Unknwon: Why change to `mention`?

Unknwon:
Could you clean up to 1 commit?

And one minor comment, otherwise LGTM.

modules/markdown/markdown_test.go
cstroe: You're right, that should be an issue mention.  A mention could be a commit mention, an issue mention, or a username mention.

cstroe:
@Unknwon, all done

Unknwon:
Thanks!

Issue #3110, opened 2016-05-20
private repo can be cloned with empty user/password


dcale:
I'm running Gogs Version: 0.9.15.0323 on Ubuntu. I changed a previously public repo to a private one and can still clone if I know the URL (using the http scheme). Git will prompt me for a username/password but by simply hitting "enter" twice the request passes through and I can clone the repository.

dcale:
changed title

terrywh:
0.9.13.0318 same problem, this is serious.

kinnarr:
I have the same behavior, if my account has an access token. After deleting this, cloning with empty username and password doesn't work any more.

kinnarr:
I found the source for this behavior and will submit a pull request later today

Unknwon:
This is claimed to be fixed by merging #3136, please help test on `develop` branch or https://try.gogs.io.

PR #3136, opened 2016-05-26
Fixes security issue #3110

kinnarr:

Fixes security issue #3110 by checking if sha token is empty and returning error in this case

### DIFF
--- a/models/error.go
+++ b/models/error.go
@@ -280,6 +280,18 @@ func (err ErrAccessTokenNotExist) Error() string {
 	return fmt.Sprintf("access token does not exist [sha: %s]", err.SHA)
 }
 
+type ErrAccessTokenEmpty struct {
+}
+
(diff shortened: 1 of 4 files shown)


roblabla:
Eh, that's weird. Doesn't that mean that there is an empty AccessToken in the database ? I'm pretty sure that's not normal.

kinnarr:
No. It's the way, how xorm queries the db. If there is no query (or an empty query, because sha is empty), it returns the first entry in the table.

bkcsoft:
Looks good to me :+1:

Unknwon:
Seems having conflict?

kinnarr:
@Unknwon fixed

Unknwon:
It would be even better to cleanup into 1 commit :D

Unknwon:
Thanks!

Issue #2988, opened 2016-04-18
Issue comment notification mail do not contain comment author


strk:
Mail notifications for issue comments do not contain comment author name, but do contain full comment. It'd be more useful to have the commenter name in there too.

Beside, the issue comment notification body is hard-coded in mail.go (SendIssueNotifyMail) so the content cannot be tweaked using templates.

strk:
I was just hit by this again, and found my previously reported issue before filing a new one.
Is a self upvote valid ? :)

Unknwon:
Can you give an example.... ?

strk:
Your comment arrived with `From: 无闻 <notifications@github.com>` so I guess it could be considered as including the comment author (in the From field). Closing. Will re-file with an example if it doesn't include the author.

PR #3137, opened 2016-05-26
add user DisplayName in mail From when notifying on issue

saimonn:
the From: header in mails sent on issue messages notification was always set to the default setting.MailService.From parameter.

This PR adds a more speaking From: header, letting the mail address unchanged.

### DIFF
--- a/modules/mailer/mail.go
+++ b/modules/mailer/mail.go
@@ -126,11 +126,12 @@ func SendIssueNotifyMail(u, owner *models.User, repo *models.Repository, issue *
 		return tos, nil
 	}
 
+	from := fmt.Sprintf("\"%s\" <%s>", u.DisplayName(), setting.MailService.From)
 	subject := fmt.Sprintf("[%s] %s (#%d)", repo.Name, issue.Name, issue.Index)
 	content := fmt.Sprintf("%s<br>-<br> <a href=\"%s%s/%s/issues/%d\">View it on Gogs</a>.",
 		markdown.RenderSpecialLink([]byte(strings.Replace(issue.Content, "\n", "<br>", -1)), owner.Name+"/"+repo.Name, repo.ComposeMetas()),
 		setting.AppUrl, owner.Name, repo.Name, issue.Index)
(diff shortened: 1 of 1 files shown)


saimonn:
fixes  #2988 and

saimonn:
@bkcsoft or @Unknwon , any chances to see this PR reviewed or merged ?

modules/mailer/mail.go
```diff
@@ -126,11 +126,12 @@ func SendIssueNotifyMail(u, owner *models.User, repo *models.Repository, issue *
 		return tos, nil
 	}
 
+	from := fmt.Sprintf("\"%s\" <%s>", u.DisplayName(), setting.MailService.From)
```
Unknwon: `setting.MailService.From` is already in format `"name" <email>` https://github.com/gogits/gogs/blob/master/conf/app.ini#L184

saimonn:
@Unknwon  May be we should restrict `setting.MailService.From`  to the "Addr-Spec" format (`email@domain`, see RFC#5322, chapter 3.4.1.)

This would allow gogs to set the user name in the `From:` header. At the moment we have no information on the User when we receive a mail issue notification from gogs. We have to open the link to know who commented.

If you have any thoughts on how we should display the User name in mails, I'm open to discussion.

Unknwon:
If we're going to change default behavior, then it is a big change, and not backward compatible...

saimonn:
You are right, this would not be backward compatible.

We could have an additional setting like `setting.MailService.NotifyFrom`, restricted to  to the "Addr-Spec" format and use it when it exists in `SendIssueNotifyMail` prefixed with the Author name. The `SendIssueNotifyMail` would still behave the old way if this setting is not set.

@Unknwon  What do you think about this idea ?

Unknwon:
Why not just the user name whoever triggers the notification? I think that is part of the standard too?

```go
msg := NewMessageFrom(tos, u.DisplayName(), subject, content)
```

Unknwon:
Please sign the CLA.

PR #3138, opened 2016-05-26
Use markdown in HTML of an issue comment mail

thehowl:
At the moment, issue comments are being emailed and displayed as following:

Not having markdown parsed is quite ugly if you're not using text/plain, so with this issue markdown is parsed before being sent as an email. Also, I added a parameter to NewMessage (`bodyPlain`), which is the body that will be sent as `text/plain` (and the one that will be displayed by mail clients which do not correctly display HTML mails).


Unknwon:
Please fix conflict...

thehowl:
Done.

modules/mailer/mail.go
```diff
@@ -52,7 +51,7 @@ func SendUserMail(c *macaron.Context, u *models.User, tpl base.TplName, code, su
 		return
 	}
 
-	msg := NewMessage([]string{u.Email}, subject, body)
+	msg := NewMessage([]string{u.Email}, subject, body, "")
```
Unknwon: Looks like you're always passing empty string as `bodyPlain`, then why do you need this?

thehowl: I'm not always passing an empty string. Check [this](https://github.com/gogits/gogs/pull/3138/files#diff-ca7ee5dc0cfb6781d976e6928f5a9c3eR128).

modules/mailer/mail.go
```diff
@@ -126,11 +125,16 @@ func SendIssueNotifyMail(u, owner *models.User, repo *models.Repository, issue *
...
+	var (
+		subject  = fmt.Sprintf("[%s] %s (#%d)", repo.Name, issue.Name, issue.Index)
+		gogsLink = fmt.Sprintf("%s%s/%s/issues/%d", setting.AppUrl, owner.Name, repo.Name, issue.Index)
+		content  = fmt.Sprintf("%s<br>&mdash;<br><a href=\"%s\">View it on Gogs</a>.",
+			markdown.Render([]byte(issue.Content), repo.RepoLink(), repo.ComposeMetas()), gogsLink)
+		// Markdown is meant to look good even when it can't be parsed into HTML, that's why it's being
+		// written directly in the email for the text/plain body (fallback version).
+		contentPlain = fmt.Sprintf("%s\n\n---\nView it on gogs: %s", issue.Content, gogsLink)
+		msg          = NewMessage(tos, subject, content, contentPlain)
+	)
```
Unknwon: Thanks for pointing out, the way code is done looks confusing, I recommend use `varName := xxx` instead.

Unknwon:
Please sign the CLA.

PR #3174, opened 2016-06-07
Re-work MAX_DIFF_LINES: supress diff per file, not the whole diff.

andreynering:
Today Gogs supresses the whole diff when it is too big. I dislike it, because I can't see anything about the commit other than the message. I have no tip about what files changed.

This PR makes Gogs supress diffs per file, like GitHub does. It looks like this (I hardcoded this case):

screenshot at 2016-06-06 21 39 45

So now `MAX_DIFF_LINES` sets the max lines per file (I changed the default to 500).

I am open to suggestion if someone dislike this change.

To-Do:
- [x] ~~Also supress by the number of files, since we can have a huge number of small files, which also makes the browser hangs.~~ (Done).

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -690,6 +690,8 @@ diff.show_unified_view = Unified View
 diff.stats_desc = <strong> %d changed files</strong> with <strong>%d additions</strong> and <strong>%d deletions</strong>
 diff.bin = BIN
 diff.view_file = View File
+diff.file_supressed = File diff supressed because it is too large
+diff.too_many_files = Some files were not shown because too many files changed in this diff
 
 release.releases = Releases
 release.new_release = New Release
--- a/models/git_diff.go
+++ b/models/git_diff.go
@@ -48,10 +48,10 @@ const (
 )
 
 type DiffLine struct {
-	LeftIdx       int
-	RightIdx      int
-	Type          DiffLineType
-	Content       string
+	LeftIdx  int
+	RightIdx int
+	Type     DiffLineType
+	Content  string
 }
 
 func (d *DiffLine) GetType() int {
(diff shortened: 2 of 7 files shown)


andreynering:
Updated. The defaults are:

- File has more than 500 lines -> supress
- File has line(s) with more than 500 characteres (like a JS or CSS bundle, or bindata) -> supress
- Diff has more than 100 files changed -> show only firsts 100 files + a message informing some files are not being shown

These 3 options are cofigurable on `app.ini`.

conf/locale/locale_en-US.ini
```diff
@@ -687,6 +687,8 @@ diff.show_unified_view = Unified View
 diff.stats_desc = <strong> %d changed files</strong> with <strong>%d additions</strong> and <strong>%d deletions</strong>
 diff.bin = BIN
 diff.view_file = View File
+diff.file_supressed = File diff supressed because it is too big
```
Unknwon: Maybe should use `too large`?

models/git_diff.go
```diff
@@ -48,10 +48,10 @@ const (
 )
 
 type DiffLine struct {
-	LeftIdx       int
-	RightIdx      int
-	Type          DiffLineType
-	Content       string
```
Unknwon: Is it result from `gofmt`?

Unknwon:
Just two minor comments, otherwise LGTM.

models/git_diff.go
andreynering: Yes

conf/locale/locale_en-US.ini
andreynering: @Unknwon Good catch. Fixed.

Unknwon:
Thanks!

PR #3176, opened 2016-06-07
Add a default repository permission to orgs

roblabla:
This patch allows the org admin to set the default access mode of the Org's member to all of the org's Repositories.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -382,6 +382,9 @@ func runWeb(ctx *cli.Context) error {
 					m.Post("/slack/:id", bindIgnErr(auth.NewSlackHookForm{}), repo.SlackHooksEditPost)
 				})
 
+				m.Combo("/member_privileges").Get(org.SettingsMemberPrivileges).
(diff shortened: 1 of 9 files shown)


roblabla:
hmm. I think a better way to do this might be to leverage the existing "team" infrastructure. Having an invisible "default" team with all users/repos that the toggle would act on. What I currently have breaks the "Collaborative Repositories" section, and probably a few other things.

I'll try to implement this.

Unknwon:
Thanks your PR but does not seem a good idea...

roblabla:
Erm, yeah ? That's exactly the use-case for it. It allows an org to give permissions on all of its repositories (including private) to all its members. Github has a similar feature : https://github.com/organizations/gogits/settings/member_privileges . The default, obviously, is to give no permission at all.

The use-case is, in my company, every employee is allowed to access all repositories. Before this PR, I had a "members" team that contained all repositories and all members. However, it meant that I had to handle that manually, and it was a hassle. Now, I just allow every member of the org to have "read" permission, and that's it.

Unknwon:
OK, will take a deep look.

Unknwon:
Since now we're able add organization member as repository collaborator, I'm closing this.

PR #3200, opened 2016-06-21
Fix empty space surrounding hidden issue labels

chriswatt:
Hidden labels on an issue still have some padding, meaning labels move slightly further down as the amount of labels increase.

Before
labelsbefore

After
labelsafter


Unknwon:
Thanks!

PR #3203, opened 2016-06-22
Fix capitalisation of repo-name in news

SjonHortensius:
use 'official' repo.Name instead of incoming repoName; to enforce correct capitalisation


Unknwon:
Thanks!

PR #3207, opened 2016-06-24
Fix negative issue count set by Check Repository Statistics

roblabla:

num_issues and num_closed_issues don't contain the PR-issues.

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -1832,14 +1832,14 @@ func CheckRepoStats() {
 
 	// ***** START: Repository.NumClosedIssues *****
 	desc := "repository count 'num_closed_issues'"
-	results, err := x.Query("SELECT repo.id FROM `repository` repo WHERE repo.num_closed_issues!=(SELECT COUNT(*) FROM `issue` WHERE repo_id=repo.id AND is_closed=?)", true)
+	results, err := x.Query("SELECT repo.id FROM `repository` repo WHERE repo.num_closed_issues!=(SELECT COUNT(*) FROM `issue` WHERE repo_id=repo.id AND is_closed=? AND is_pull=?)", true, false)
 	if err != nil {
 		log.Error(4, "Select %s: %v", desc, err)
 	} else {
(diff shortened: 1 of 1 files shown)


BurakDev:
Nice catch @roblabla

tkausl:
The whole checking if it has to be updated and then updating it one by one is unnecessary. We could replace the whole block with just one execution of
```
UPDATE `repository` SET num_closed_issues = (SELECT COUNT(*) FROM `issue` WHERE repo_id = `repository`.`id` AND is_closed = 1 AND is_pull = 0)
```

BurakDev:
@tkausl Is your query compatible with all database backend ? Don't forget that

andreynering:
What do you guys think of not actually using this column, but count for the open issues when needed? So, no need to update it manually.

tkausl:
@andreynering I already thought about that. Not sure why it was implemented like this, probably performance reasons.

andreynering:
@tkausl Count queries are fast with proper index, I think performance wouldn't be a problem.

Unknwon:
Thanks!

Unknwon:
@andreynering It would be ideal, but since we can now keep this column value correct, I think it would be not be necessary.

Unknwon:
@tkausl: @BurakDev pointed out one part of the reasons, another reason it would be nice to show which repository is been updated for debugging purpose. And since we only do this on start time and the loop is usually small (most of time 0), it would not hurt performance at all in my opinion.

FGM:
Any suggestion about the best way to fix it once the database contains the incorrect values ? I currently have "Issues: -3" on Gogs 0.9.28.0527. Or will it automagically be fixed when the update is applied from the debian package ?

Unknwon:
@FGM 

`the debian package ?` is out of our hands and have been failed to build latest Gogs for quite a while, but once you use latest Gogs connect to your existing database, it fixes automatically.

FGM:
@Unknwon thank for the confirmation.

PR #3208, opened 2016-06-25
Use a gopher as default avatar (rather than the gravatar logo)

strk:
Also changes the avatar from a jpeg to a png, to allow for
transparent background. The indexed png is also smaller in size.

Note that at the moment the default avatar is only used when
the user requested a custom avatar and the custom avatar file
is not found (should never happen).

In the future the default avatar could be used as a default
return when by-mail avatar lookups fail too (both gravatar
and libravatar support passing a default)


bkcsoft:
Looks good to me :)

@Unknwon

strk:
maybe a mustache could be added too :)

Unknwon:
Please attach a screenshot for a default avatar.

strk:
You can see it in the "Files Changed" tab.
https://raw.githubusercontent.com/gogits/gogs/1a20edf49115c7f818d942eb0f22361f03b4e1ee/public/img/avatar_default.png

Unknwon:
I want to see in action how it looks like in the whole page...

strk:
It wasn't easy as the only case in which the "default avatar" is currently shown is when a "CustomAvatar" is set in the database but the corresponding file is missing. In absence of errors a random avatar is instead used for users not uploading or deleting their custom avatar.

screenshot

Unknwon:
OK, lGTM.

PR #3209, opened 2016-06-25
Do not assume avatar needs be changed from gravatar.com

strk:
Do not assume avatar needs be changed from gravatar.com
    
Always send user to settings screen to change avatar.
Drops "change_custom_avatar" localized message, keeps "change_avatar"
for the generic one.


bkcsoft:
Looks good to me

@Unknwon thoughts?

Unknwon:
Thanks!

PR #3227, opened 2016-07-02
The pruning for the synchronized mirrors is optative 

psaavedra:
**Feature:**
The pruning for the synchronized mirrors is a option now. Default value: enable_prune = true

**Motivation:**
As far as I know, currently the mirroring is being executed with **--prune** option. This cause the local branches deletion each time the mirror is sync with the remote/origin repository. The idea is to allow mirrors which local branches.

**Limitations of the patch:**
- Explicit update instruction for the new `Mirror` table column named `enabled_prune`
- Added two new i18n text chains. Only translated to EN and ES


Unknwon:
Please run `go fmt` to format code...

Unknwon:
Please resolve conflict...

models/repo.go
```diff
@@ -656,8 +658,12 @@ func GetMirror(repoId int64) (*Mirror, error) {
 }
 
 func updateMirror(e Engine, m *Mirror) error {
-	_, err := e.Id(m.ID).Update(m)
```
psaavedra: to be honest, these are my first steps with Go and XORM. For some reason this update() ignores the new column added in the Mirror model (enable_prune) for this reason I included the extra update instruction.

I could suggest a review for a more avances Go programmer .

Unknwon: `xxxx.AllCols().xxx`

bkcsoft:
@Unknwon what about the spanish translation? Shouldn't that one be removed?

Unknwon:
Hi, can you cleanup to 1 commit?

psaavedra:
New pull request: https://github.com/gogits/gogs/pull/3246

Unknwon:
You could actually just do `git push --force` after cleanup...

PR #3235, opened 2016-07-05
Return avatar link as absolute url

strk:
Fixes relative urls coming from api/v1

See https://github.com/drone/drone/issues/1701

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -263,11 +263,11 @@ func (u *User) RelAvatarLink() string {
 	return setting.GravatarSource + u.Avatar
 }
 
-// AvatarLink returns user gravatar link.
+// AvatarLink returns user avatar link.
 func (u *User) AvatarLink() string {
 	link := u.RelAvatarLink()
 	if link[0] == '/' && link[1] != '/' {
-		return setting.AppSubUrl + link
+		return strings.TrimSuffix(setting.AppUrl, "/") + link
 	}
 	return link
 }


models/user.go
```diff
@@ -263,11 +263,11 @@ func (u *User) RelAvatarLink() string {
 	return setting.GravatarSource + u.Avatar
 }
 
-// AvatarLink returns user gravatar link.
+// AvatarLink returns user avatar link.
 func (u *User) AvatarLink() string {
 	link := u.RelAvatarLink()
 	if link[0] == '/' && link[1] != '/' {
-		return setting.AppSubUrl + link
+		return strings.TrimSuffix(setting.AppUrl, "/") + link
```
bkcsoft: `path.Join` instead?

tboerger: `path.Join` destroys the `//` within `http://` or `https://`, e.g. https://play.golang.org/p/BH7RapbPzC ;)

tboerger: Does `setting.AppUrl` include `setting.AppSubUrl`?

bkcsoft: `AppUrl` contains `Domain+AppSubUrl` so this should be correct

bkcsoft:
Looks Good To Me :smiley:

bkcsoft:
@Unknwon the only negative thing I can thing of there is backwards compatibility

strk:
On Tue, Jul 05, 2016 at 02:31:55PM -0700, Kim Carlbäcker wrote:
> @Unknwon the only negative thing I can thing of there is backwards compatibility

Note that the function can currently return either relative or
absolute url, depending on user setting (UseCustomAvatar or not).

The patch makes it always absolute.

Can you really think of use cases in which a caller would base
some decision on the absolute-vs-relative return ?

--strk;

bkcsoft:
@strk Makes sense. I stand by my LGTM :+1:

models/user.go
Unknwon: Maybe `setting.AppUrl + link[1:]` is more efficient?

strk: I dubt the difference would be perceptible by the human eye,
and surely I've no time to profile it. Feel free to tweak
as you like, as long as it works.

Unknwon: Yeah, please change to that.

strk: I just remembered the rationale for my approach:
if AppUrl is in the form "https://try.gogs.io"
your version would result in a bogus url
(ie: "https://try.gogs.ioavatar/3")

So my code is trimming the AppUrl instead, knowing that
the avatar link starts with a slash (as per prior check).

Unknwon: No, that is not possible: https://github.com/gogits/gogs/blob/master/modules/setting/setting.go#L317

strk: I see, well the code still takes the path of least unpredictability.
It is locally known that the "link" variable starts with a slash,
and it is known that an URL would be valid either with and w/out
and ending slash, so it's just safer to avoid the risk.

In absence of a dedicated testsuite, I'd rather avoid risks.
Someone changing the code in setting.go could hardly guess
some code in user.go was dependent on AppUrl having an enforced
ending slash.

Go strings have known length, so the additional cost of TrimSuffix
over chopping the initial portion of relative avatar link is that
of comparing a character for each slash found in AppUrl.

Is that really worth the time we're spending on discussing this ?

strk:
Is this PR not being merged due to lack of agreement on the `setting.AppUrl + link[1:]` construct ?

Unknwon:
Thanks!

andreynering:
Was this PR suppose to modify the URL in `` tags also?

If yes, this PR broke the avatar images when you access Gogs from a different URL.

Let's say you have `AppUrl` set to your external URL:

```ini
;app.ini

AppUrl = https://gogs.domain.com
```

But while in your internal network you access it by a IP or machine name:

```
http://<IP-ADDRESS>:<PORT>/
```

For internal network the avatar is now broken, because ` https://gogs.domain.com` is not accessible from the internal network.

Also, if someone had to use the API from the inside, the avatar URL would also not be correct.

strk:
It was not the primary intention, which was instead returning an absolute url from the API response,
but I guess both the API and the templating are using the same call to obtain the avatar link, thus possibly breaking the use case you report.

I won't comment on the specific use case of URLs being valid or not depending on provenence.

andreynering:
Maybe we could make the URL relative in `` tags, and send both on API:

```json
{
  "relative_avatar": "/avatar/3",
  "absolute_avatar": "https://domain.com/avatar/3"
}
```

lstahlman:
The API could just return an absolute URL based on the URL used to make the API request.

Unknwon:
https://github.com/gogits/gogs/issues/3348

PR #3151, opened 2016-05-31
Fixes #1071 - Web editor for repository files (Create new, edit, upload, rename, delete)

richmahn:
NOTE: This also requires https://github.com/gogits/git-module/pull/12 for git-module for handling branches

NOTE: Discussion has already taken place on previous pull request #3092 (closed)

NOTE: I renamed font-awesome, semantics and simplemde just so we don't need the version numbers in path to them (they have version in their headers) so that in the templates/base/head.tmpl we don't need to change it, as that causes problems with custom/templates/base/head.tmpl if it has been copied there. I left the version number in the jquery library though.

Adds create new, edit, rename, uplad and delete feature for files

CodeMirror/SimpleMDE (same package) for editor
Moves workingPool out of the wiki.go file so it can be used for repos as well
Uses Dropzone to upload files
Feel free to code review and give constructive criticism. My first addition with Go Lang. Thanks.

PR #3241, opened 2016-07-07
On showing diff/file, use the tab_width specified on .editorconfig, if any

### LINKED ISSUE #3182: Allow for displaying tabs as 4 spaces
Gogs currently shows tabs as `\t`, which is rendered as eight spaces on every browser.

Given that displaying tabs as four spaces is the default in many editors and environments, could there be an option to display them as such?

andreynering:
Closes #3182

---

In the future, after #3151 being merged, we can also make the editor use the specs in `.editorconfig`.

### DIFF
--- a/modules/context/repo.go
+++ b/modules/context/repo.go
@@ -6,10 +6,12 @@ package context
 
 import (
 	"fmt"
+	"io/ioutil"
 	"path"
 	"strings"
 
 	"github.com/Unknwon/com"
+	"gopkg.in/editorconfig/editorconfig-core-go.v1"
 	"gopkg.in/macaron.v1"
 
 	"github.com/gogits/git-module"
@@ -67,6 +69,28 @@ func (r *Repository) IsWriter() bool {
 // HasAccess returns true if the current user has at least read access for this repository
 func (r *Repository) HasAccess() bool {
 	return r.AccessMode >= models.ACCESS_MODE_READ
+}
+
+// GetEditorconfig returns the .editorconfig definition if found in the
+// HEAD of the default repo branch.
+func (r *Repository) GetEditorconfig() (*editorconfig.Editorconfig, error) {
+	commit, err := r.GitRepo.GetBranchCommit(r.Repository.DefaultBranch)
+	if err != nil {
+		return nil, err
+	}
+	treeEntry, err := commit.GetTreeEntryByPath(".editorconfig")
+	if err != nil {
+		return nil, err
+	}
+	reader, err := treeEntry.Blob().Data()
+	if err != nil {
+		return nil, err
+	}
+	data, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return nil, err
+	}
+	return editorconfig.ParseBytes(data)
 }
 
 func RetrieveBaseRepo(ctx *Context, repo *models.Repository) {
--- a/modules/template/template.go
+++ b/modules/template/template.go
@@ -17,6 +17,7 @@ import (
 
 	"golang.org/x/net/html/charset"
 	"golang.org/x/text/transform"
+	"gopkg.in/editorconfig/editorconfig-core-go.v1"
 
 	"github.com/gogits/gogs/models"
 	"github.com/gogits/gogs/modules/base"
@@ -108,6 +109,15 @@ func NewFuncMap() []template.FuncMap {
 		"FilenameIsImage": func(filename string) bool {
 			mimeType := mime.TypeByExtension(filepath.Ext(filename))
 			return strings.HasPrefix(mimeType, "image/")
+		},
+		"TabSizeClass": func(ec *editorconfig.Editorconfig, filename string) string {
+			if ec != nil {
+				def := ec.GetDefinitionForFilename(filename)
+				if def.TabWidth > 0 {
+					return fmt.Sprintf("tab-size-%d", def.TabWidth)
+				}
+			}
+			return "tab-size-8"
 		},
 	}}
 }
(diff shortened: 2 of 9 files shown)


modules/template/template.go
```diff
@@ -104,6 +105,15 @@ func NewFuncMap() []template.FuncMap {
...
 			return setting.ThemeColorMetaTag
 		},
+		"TabSizeClass": func(ec *editorconfig.Editorconfig, filename string) string {
+			if ec != nil {
+				def := ec.GetDefinitionForFilename(filename)
+				if def.TabWidth > 0 {
+					return fmt.Sprintf("tab-size-%d", def.TabWidth)
+				}
+			}
+			return "tab-size-8"
```
silverwind: Maybe default to 4 spaces? I think it's a lot more common.

andreynering: I will let this open for discussion, but in theory 8 is the default for tabs.

What we can do is allowing one change the default in `app.ini`, as well disable the `editorconfig` functionality all together. I think it would be a good idea, in case someone dislike this. What do you think?

silverwind: While I would welcome such options, I think they'd be better placed in the user's settings panel instead of `app.ini`, because it's mostly a personal preference. Maybe through a new 'Display' settings pane.

Anyways, I'm fine if you want to keep it at 8. It's still the default in editors like `vim`, and I'm not really affected by it as i use `.editorconfig` in all my projects.

modules/context/repo.go
```diff
@@ -68,6 +72,36 @@ func (r *Repository) HasAccess() bool {
...
 
+// GetEditorconfig returns the .editorconfig definition if found in the
+// HEAD of the default repo branch. If any error happens (like if the file
+// doesn't exists) it returns nil
+func (r *Repository) GetEditorconfig() *editorconfig.Editorconfig {
+	if r.editorConfig != nil {
+		return r.editorConfig
+	}
+
+	commit, err := r.GitRepo.GetBranchCommit(r.Repository.DefaultBranch)
```
Unknwon: Should we do something with the errors?

andreynering: I updated with a little refactoring.

I am now returning the errors for future use, but ignoring them in this case, because we don't want to break anything if the file doesn't exists/can't be parsed.

Maybe your idea would be printing them in the log?

Unknwon:
You can check error with https://github.com/gogits/git-module/blob/master/error.go#L30

andreynering:
Done

public/less/_repository.less
```diff
@@ -1326,3 +1326,36 @@
 		width: 100%!important;
 	}
 }
+
+.tab-size-1 {
```
Unknwon: Something like this? https://github.com/andreynering/gogs/blob/460907f09fc88cf9e67b095e298e2f4037ddd19b/public/less/_base.less#L369

andreynering: That worked. Updated.

modules/template/template.go
rugk: +1 for 4 spaces by default. I really think that's common as the main discussion is always "4 spaces vs tabs". If anyone has any stats it would be very useful, but otherwise I would subjectively use 4 spaces by default.

andreynering: @rugk This page has some statistics: https://ukupat.github.io/tabs-or-spaces/

I thought on 8 because it is backward compatible: its how it would be shown on Gogs today.

I will let @Unknwon decide this.

rugk: Ah okay, don't know the previous behaviour of Gogs, but if it is indeed 8 this is a good argument for 8 spaces by default.
And thanks for the stats: Unfortunately there is no "any language" diagram.

silverwind: the argument for 8 is that, while archaic, it's the default on Github too.
the argument for 4 is that it's more common.

there could also be made an argument for 2 spaces, it's predominant in a few languages, so ultimatively, there should be an option. maybe in `app.ini` as a start, but I don't see it as a blocker on this PR.

rugk: > the argument for 8 is that, while archaic, it's the default on Github too.

That's wrong. E.g. [editing this ini file](https://github.com/gogits/docs/edit/master/TOC.ini) it is 2 spaces.

silverwind: No idea why it's two there. I see 8 for repos without `.editorconfig` in both viewer and editor, [example](https://github.com/silverwind/Combusted/blob/master/Combusted.lua), [example](https://github.com/sindresorhus/public-ip/blob/master/test.js).

rugk: Ah, all right. I see it there too. GitHub obviously does it depending on the file type or... - does it even analyse the current file?

rugk: It seems it analyses the current file as I tried it with different Markdown files.

silverwind: So it's more than just `.editorconfig`, interesting.

rugk: So GitHub is no argument for 8 spaces and personally I would prefer the default for Gogs to be 4 spaces or something as intelligent as GitHub does.

silverwind: Took a closer look. Default tab-width is still 8:

https://github.com/silverwind/docs/blob/patch-1/TOC.ini

It's just that the ACE editor defaults to 2-space indent on files not containing a tab character. As soon as a tab is there, it goes to 8-wide tabs indentation, overridden through `.editorconfig`.

Unknwon:
Please sign the CLA.

Unknwon:
Having conflict.

andreynering:
@Unknwon Conflicts fixed

Unknwon:
Thanks!

PR #3246, opened 2016-07-08
The pruning for the synchronized mirrors is optative

psaavedra:
**Feature:**
The pruning for the synchronized mirrors is a option now. Default value: enable_prune = true

**Motivation:**
As far as I know, currently the mirroring is being executed with --prune option. This cause the local branches deletion each time the mirror is sync with the remote/origin repository. The idea is to allow mirrors which local branches.

**Related Pull Requests**

* https://github.com/gogits/gogs/pull/3227 (Closed)

**Changes:** 

Executed go fmt

getEngine() not handles DB parameters (#2972) (#2974)

Uses .AllCols() for Update in updateMirror()

Spanish traslation removed

Fixed a wrong way to ommit the --prune option in process.ExecDir() for MirrorUpdate function


models/models.go
```diff
@@ -97,14 +97,18 @@ func LoadConfigs() {
 
 func getEngine() (*xorm.Engine, error) {
 	cnnstr := ""
+	var Param string = "?"
```
Unknwon: Hmm?... what happened to this file?

psaavedra: Good point, I didn't see this change till now. This is that is not a explicit change done by me. Let sync again with your develop branch.

psaavedra: That change has disappeared after a re-merge.

Unknwon:
Thanks!

PR #3254, opened 2016-07-11
Missed quote in struct definition 

fsotomsk:
Missed quote in repo_form.go


Unknwon:
Please read https://github.com/gogits/gogs/wiki/Contributing-Code

Issue #3140, opened 2016-05-27
Add repo timestamps to API output


jmptbl:
We DO NOT take questions or config/deploy problems on GitHub, please use our forum: https://discuss.gogs.io

Please take a moment to search that an issue doesn't already exist.

For bug reports, please give the relevant info:

- Gogs version (or commit ref): 0.9.25.0506
- Git version: 2.1.4
- Operating system: Debian 8
- Database:
  - [ ] PostgreSQL
  - [X] MySQL
  - [ ] SQLite
- Can you reproduce the bug at http://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [X] Not relevant
- Log gist:

## Description

API calls that retrieve repo information does not include creation and update timestamps despite these details being in the SQL table.  Can it be added to the API output, please?  Example call:

`curl -H 'Authorization: token XXX' https://site/api/v1/repos/orgname/reponame
{"id":98,"owner":{"id":28,"username":"orgname","full_name":"orgname","email":"orgemail@example.com","avatar_url":"https://secure.gravatar.com/avatar/7aad197485430a2feacd2bd5a460be78"},"full_name":"orgname/reponame","private":false,"fork":false,"html_url":"https://site/orgname/reponame","clone_url":"https://site/orgname/reponame.git","ssh_url":"git@site:orgname/reponame.git","permissions":{"admin":true,"push":true,"pull":true}}`

lstahlman:
I can pick this up.  The new response will include the properties created_at and updated_at.
Example:
`{  
  "id":2,
  "owner":{  
    "id":1,
    "username":"test",
    "full_name":"",
    "email":"test@localhost",
    "avatar_url":"https://secure.gravatar.com/avatar/30fa571d2f85bb1c5a5b574d5c5c5005"
  },
  "full_name":"test/Test-Repo-1",
  "description":"Some description",
  "private":false,
  "fork":false,
  "html_url":"http://localhost:3000/test/Test-Repo-1",
  "clone_url":"http://localhost:3000/test/Test-Repo-1.git",
  "ssh_url":"test@localhost:test/Test-Repo-1.git",
  "stars_count":1,
  "forks_count":0,
  "watchers_count":1,
  "open_issues_count":0,
 `**`"created_at":"2016-07-09T08:15:47-07:00",
  "updated_at":"2016-07-11T11:45:40-27:00",`**`
  "permissions":{  
    "admin":true,
    "push":true,
    "pull":true
  }
[...]

PR #3255, opened 2016-07-11
Add timestamps to repository api response

lstahlman:
Adds created_at and updated_at properties.  See #3140 
This change is dependent on gogits/go-gogs-client#34


Unknwon:
Thanks!

PR #3256, opened 2016-07-12
Ignore Response Body for Slack Hooks #3169

### LINKED ISSUE #3169: Rocket.Chat webhook: slack failed with: {"success":true}
- Gogs version:  0.9.28.0527
- Git version: 2.7.4
- Operating system: Ubuntu 16.04
- Database:
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
  - [ ] No
  - [x] Not relevant
- Log: 

"[...rc/runtime/asm_arm.s:990 goexit()] [E] slack failed with: {"success":true}
[T] Hook delivery failed: 1053b9b9-2c6f-48c6-8c40-cba5c420ace2"

Hello,

I am using Rocket.Chat and have configured a webhook in "slack" format in Gogs (since Rocket.Chat does support this without much configuration).

Even though the webhook works fine, Gogs thinks it failed, because the body response is "{"success":true}", instead of the "ok" that Gogs seems to expect.

I've been looking at Slack's webhook documentation, and it does not specify what response Slack will return. Instead it only says: "[..] in most cases you'll receive a "HTTP 200 OK" response with a JSON body indicating that your message posted successfully, [...]".

So I feel Gogs should just look at HTTP 200 OK (like it does on the custom format code path), since assuming the response is always going to be the text "ok" seems problematic if Slack changes this.
(As they have not defined the response format, they probably don't expect people to rely on this exact format either).

Anyway, still works, just thought it might confuse people.

bkcsoft:
This fixes #3169 

As per https://api.slack.com/incoming-webhooks ("Handling errors") the response body is undefined and should therefore not be checked.

### DIFF
--- a/models/webhook.go
+++ b/models/webhook.go
@@ -584,14 +584,6 @@ func (t *HookTask) deliver() {
 		return
 	}
 	t.ResponseInfo.Body = string(p)
-
-	switch t.Type {
-	case SLACK:
-		if t.ResponseInfo.Body != "ok" {
-			log.Error(5, "slack failed with: %s", t.ResponseInfo.Body)
-			t.IsSucceed = false
-		}
-	}
 }
 
 // DeliverHooks checks and delivers undelivered hooks.


barracks510:
I've tested this on Gogs v0.9.40. And it works. Thanks!

Unknwon:
Thanks!

PR #3266, opened 2016-07-13
Add Japanese Font(more)

Laica-Lunasys:
Adding japanese Font. already added this PR https://github.com/gogits/gogs/pull/2816.
but, it does not applied navbar and some object.


Unknwon:
Thanks!

PR #3267, opened 2016-07-13
Fixes #3263 - Change for Fontawesome icons to not conflict withs sema…

### LINKED ISSUE #3263: FontAwesome icons are not showing up, all are flags
This is not only happening on my development copy as of 7/12, but also on try.gogs.io. It appears any button/link that uses a class of "icon fa" and then fa-<icon_name> appears as a flag, because semantic's css file gets priority with this:

```
i.icon.fa:before {
    content: "\f2b4";
}
```
Which is a flag. Not sure why this started happening.

richmahn:
Semantic UI seems to have put in a CSS for "icon fa". Removed the "fa" from the Watching/Not Watching icon


Unknwon:
Thanks!

PR #3268, opened 2016-07-13
Changed Turkish language name to 'Türkçe'

atakangktepe:
Turkish language name changed to 'Türkçe' from 'Türk' because 'Türk' meaning is 'Turkish People' in Turkey but 'Türkçe' meaning is 'Turkish Language'.


Unknwon:
Thanks!

PR #2895, opened 2016-03-27
Made the issue stats query more secure with parameterized placeholders

tboerger:
Another try to get fine for https://github.com/gogits/gogs/issues/2892, but I need confirmation again that this works like expected.

PR #3269, opened 2016-07-13
Fixes #3180: 500 error when attempting to view issues of type 'Mentioning you'

### LINKED ISSUE #3180: 500 error when attempting to view issues of type 'Mentioning you'
**Gogs version:** 0.9.28.0527
**Git version:** 1.7.10.4
**OS:** Debian 7
**Database:** MySQL 5.6.29
**Can reproduce at http://try.gogs.io:** Yes
**Reproduce URL:** https://try.gogs.io/dannherd/sdfsddsf/issues?type=mentioned&sort=&state=open&labels=0&milestone=0&assignee=0
**Description:** Whenever I attempt to view issues of type 'Mentioning you', it crashes with a 500 error.
**Log line:** 2016/06/08 12:28:15 [...outers/repo/issue.go:185 Issues()] [E] Issues: %v: Error 1052: Column 'is_closed' in where clause is ambiguous

lstahlman:
This resolves #3180 by adding table aliases to the Issues SQL statement.

Once the 500 error was gone, I discovered another bug where the open/closed counts always showed 0 when viewing the same type of issues.  This was also fixed by adding table aliases to the Issues Stats statement.

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -506,55 +506,55 @@ func Issues(opts *IssuesOptions) ([]*Issue, error) {
 	sess := x.Limit(setting.IssuePagingNum, (opts.Page-1)*setting.IssuePagingNum)
 
 	if opts.RepoID > 0 {
-		sess.Where("issue.repo_id=?", opts.RepoID).And("is_closed=?", opts.IsClosed)
+		sess.Where("issue.repo_id=?", opts.RepoID).And("issue.is_closed=?", opts.IsClosed)
 	} else if opts.RepoIDs != nil {
 		// In case repository IDs are provided but actually no repository has issue.
 		if len(opts.RepoIDs) == 0 {
 			return make([]*Issue, 0), nil
 		}
-		sess.In("repo_id", base.Int64sToStrings(opts.RepoIDs)).And("is_closed=?", opts.IsClosed)
+		sess.In("issue.repo_id", base.Int64sToStrings(opts.RepoIDs)).And("issue.is_closed=?", opts.IsClosed)
 	} else {
 		sess.Where("issue.is_closed=?", opts.IsClosed)
 	}
 
 	if opts.AssigneeID > 0 {
-		sess.And("assignee_id=?", opts.AssigneeID)
+		sess.And("issue.assignee_id=?", opts.AssigneeID)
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR!

We removed those by purpose a while ago, can you confirm everything works after this change?

lstahlman:
Ah, I see I basically undid some changes you made for #2895 in commit 0325bec283db28ee6a464f2d1884df0c48977373.

It looks like xorm automatically adds table aliases to statements when a join is involved (there is a needTableName() function in xorm/statement.go that returns true only if the statement contains a join).

The bug you fixed in 0325bec283db28ee6a464f2d1884df0c48977373 was due to the statements referencing table aliases that weren't set.  The "Mentioning you" view + issue counts probably worked when the others didn't, so by taking out all the aliases the defect flipped (all other views worked except "Mentioning you").

Unknwon:
So... what do we do?

lstahlman:
I did not see any regressions on my test environment, but hold off on merging until I do a bit more testing.

lstahlman:
I think for the Issues function the best way forward is to set the table alias manually and use it for the rest of the process.  That way there is no question whether the alias exists and it is also more clear in the code where each referenced column is coming from.

I'm testing this change in my environment today.

Unknwon:
The commiter on your system does not seem right, this will lose your credit on commits.

models/issue.go
```diff
@@ -504,57 +504,58 @@ func Issues(opts *IssuesOptions) ([]*Issue, error) {
 	}
 
 	sess := x.Limit(setting.IssuePagingNum, (opts.Page-1)*setting.IssuePagingNum)
+	sess.Alias("issue")
```
Unknwon: What magic does this do?

lstahlman: This explicitly sets the table alias rather than letting xorm decide whether it's necessary (see my previous comment on why I think it's good to do this).

Unknwon: OK, but we've added `issue` alias manually below already, don't we?

lstahlman: Nevermind, the Alias function does not need to be called.  I will take that out.

Unknwon:
Thanks!

PR #3275, opened 2016-07-16
api: delete repository webhooks

barracks510:
Allows the deletion of a webhook from a repository at the
/:user/:repo/hooks/:id endpoint.

Solves drone/drone#1603.

### DIFF
--- a/routers/api/v1/repo/hook.go
+++ b/routers/api/v1/repo/hook.go
@@ -96,6 +96,15 @@ func CreateHook(ctx *context.APIContext, form api.CreateHookOption) {
 	ctx.JSON(201, convert.ToHook(ctx.Repo.RepoLink, w))
 }
 
+func DeleteHook(ctx *context.APIContext) {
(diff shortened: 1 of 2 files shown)


routers/api/v1/repo/hook.go
```diff
@@ -96,6 +96,19 @@ func CreateHook(ctx *context.APIContext, form api.CreateHookOption) {
 	ctx.JSON(201, convert.ToHook(ctx.Repo.RepoLink, w))
 }
 
+func DeleteHook(ctx *context.APIContext) {
+	if err := models.DeleteWebhook(ctx.ParamsInt64(":id")); err != nil {
+		if models.IsErrWebhookNotExist(err) {
```
Unknwon: There is no possibility to return this type of error, you only need to response 500 when error occurs.

barracks510: Fixed and rebased.

Unknwon:
Thanks!

PR #3276, opened 2016-07-17
Fixes #3228: Error on changing repo name when the repo is a mirror

### LINKED ISSUE #3228: 500 error after rename repository because of path change
There is a warning that path will change during rename a repository, but after repo name is changed the user shouldn't get served an 500 error. User should be redirected to the new url instead of the old one.

lstahlman:
Proposed fix for #3228.  ctx.Repo.Mirror is refreshed following a repo name change.

### DIFF
--- a/routers/repo/setting.go
+++ b/routers/repo/setting.go
@@ -105,6 +105,15 @@ func SettingsPost(ctx *context.Context, form auth.RepoSettingForm) {
 		}
 
 		if repo.IsMirror {
+			if isNameChanged {
+				var err error
+				ctx.Repo.Mirror, err = models.GetMirror(repo.ID)
+				if err != nil {
+					ctx.Handle(500, "RefreshRepositoryMirror", err)
+					return
+				}
+			}
+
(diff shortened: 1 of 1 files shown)


routers/repo/setting.go
```diff
@@ -105,6 +105,14 @@ func SettingsPost(ctx *context.Context, form auth.RepoSettingForm) {
 		}
 
 		if repo.IsMirror {
+			if isNameChanged {
+				var err error
+				ctx.Repo.Mirror, err = models.GetMirror(repo.ID)
+				if err != nil {
+					ctx.Handle(500, "RefreshRepositoryMirror", err)
```
Unknwon: Need `return` after this.

lstahlman: Amended to include the return statement. Are you ok with "RefreshRepositoryMirror" as the title for the error handling?

Unknwon: Yeah, it's fine.

Unknwon:
Thanks!

PR #3286, opened 2016-07-19
Added coding style modes to SimpleMDE

richmahn:
This PR is needed for the Web Edit feature, but made separate since it is just an update to the SimpleMD (built on Code Mirror) to have the loadmode.js and the mode JS files for editing different coding style modes.


Unknwon:
Thanks, just curious, would everything still work if we upgrade simplemde to [1.11.2](https://github.com/NextStepWebs/simplemde-markdown-editor/releases/tag/1.11.2) by this chance?

richmahn:
@Unknwon As long as you leave loadmode.js and the modes directory, all should work.... we could make that modes directory separate so we don't have to worry about losing it. Should i do that? It actually comes from an extension for CodeMirror.

richmahn:
@Unknwon Just to log the decisions I had made, I actually at one point had both the CodeMirror plugin, and the SimpleMD plugin being used in the gogs.js, and it would load the proper one based on the extension, SimpleMD for .md and CodeMirror for all else. Yet that wasn't very practical. I hoped to find a way to extend CodeMirror to have a good Markdown editor, but that failed. Then I figured out that since SimpleMD includes the CodeMirror library, it best in my opinion, to just keep using SimpleMD and call with in its interface the standard CodeMirror editor when not a Markdown file, and load the modes/*.js for that particular language.

richmahn:
@Unknwon So to be clear, SimpleMD is just two files, the css and js, so I added the loadmodes.js and modes directory to it. That may be ok, as long as we just update the JS and CSS file. We probably also should be updating the loadmodes.js and modes from the corresponding CodeMirror version.

Unknwon:
Thanks.

>We probably also should be updating the loadmodes.js and modes from the corresponding CodeMirror version.

I think we should create separate directory with proper CodeMirror version (CodeMirror-x.x.x) to put its files.

richmahn:
@Unknwon Ok, added the /mode and /addon/mode directories from the 5.17.0 version of CodeMirror to public/plugins/codemirror-5.17.0. Now will update #3287 to use this new location.

Unknwon:
Thanks!

PR #3300, opened 2016-07-21
Add support for .diff and .patch

barracks510:
Add the ability to get text-diff and format-patch by adding .diff or
.patch in the end of a commit url. Issue #2641

Obsoletes PR #2846

### DIFF
--- a/models/git_diff.go
+++ b/models/git_diff.go
@@ -418,13 +418,13 @@ func GetDiffRange(repoPath, beforeCommitID string, afterCommitID string, maxLine
 	return diff, nil
 }
 
-func GetRawDiff(repoPath, commitId, diffType string) (string, error) {
+func GetRawDiff(repoPath, commitID, diffType string) (string, error) {
 	repo, err := git.OpenRepository(repoPath)
 	if err != nil {
 		return "", err
 	}
 
-	commit, err := repo.GetCommit(commitId)
+	commit, err := repo.GetCommit(commitID)
 	if err != nil {
 		return "", err
 	}
@@ -433,17 +433,17 @@ func GetRawDiff(repoPath, commitId, diffType string) (string, error) {
 	switch diffType {
 	case "diff":
 		if commit.ParentCount() == 0 {
-			cmd = exec.Command("git", "show", commitId)
+			cmd = exec.Command("git", "show", commitID)
 		} else {
 			c, _ := commit.Parent(0)
-			cmd = exec.Command("git", "diff", "-M", c.ID.String(), commitId)
+			cmd = exec.Command("git", "diff", "-M", c.ID.String(), commitID)
 		}
 	case "patch":
 		if commit.ParentCount() == 0 {
-			cmd = exec.Command("git", "format-patch", "--no-signature", "--stdout", "--root", commitId)
+			cmd = exec.Command("git", "format-patch", "--no-signature", "--stdout", "--root", commitID)
 		} else {
 			c, _ := commit.Parent(0)
-			query := fmt.Sprintf("%s...%s", commitId, c.ID.String())
+			query := fmt.Sprintf("%s...%s", commitID, c.ID.String())
 			cmd = exec.Command("git", "format-patch", "--no-signature", "--stdout", query)
 		}
 	default:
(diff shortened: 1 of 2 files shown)


models/git_diff.go
```diff
@@ -418,6 +418,44 @@ func GetDiffRange(repoPath, beforeCommitID string, afterCommitID string, maxLine
 	return diff, nil
 }
 
+func GetRawDiff(repoPath, commitId string, diffType string) (string, error) {
```
Unknwon: Can be simplified to `func GetRawDiff(repoPath, commitID, diffType string) (string, error) {`.

models/git_diff.go
```diff
@@ -418,6 +418,44 @@ func GetDiffRange(repoPath, beforeCommitID string, afterCommitID string, maxLine
...
+		if commit.ParentCount() == 0 {
+			cmd = exec.Command("git", "format-patch", "--no-signature", "--stdout", "--root", commitId)
+		} else {
+			c, _ := commit.Parent(0)
+			query := fmt.Sprintf("%s...%s", commitId, c.ID.String())
+			cmd = exec.Command("git", "format-patch", "--no-signature", "--stdout", query)
+		}
+	}
+	cmd.Dir = repoPath
+	cmd.Stderr = os.Stderr
```
Unknwon: Use `bytes.Buffer` to collect `os.Stderr`.

models/git_diff.go
```diff
@@ -418,6 +418,44 @@ func GetDiffRange(repoPath, beforeCommitID string, afterCommitID string, maxLine
...
+		return "", err
+	}
+
+	commit, err := repo.GetCommit(commitId)
+	if err != nil {
+		return "", err
+	}
+
+	var cmd *exec.Cmd
+	if diffType == "diff" {
```
Unknwon: We should use `select diffType {}` and return error is we do not recognize that type.

routers/repo/commit.go
```diff
@@ -195,7 +195,18 @@ func Diff(ctx *context.Context) {
 }
 
 func RawDiff(ctx *context.Context) {
-	panic("not implemented")
+	userName := ctx.Repo.Owner.Name
```
Unknwon: Those variables are only been used once, I think we don't need to create them at all, just pass values to the `GetRawDiff` directly.

models/git_diff.go
barracks510: I believe the precedent is set in Line#395 of `git_diff.go`. Do you have a line of source I could reference to fix this?

Unknwon: https://github.com/gogits/git-module/blob/master/command.go#L92

barracks510:
Ah. Sorry. My bad. It must be getting late..

models/git_diff.go
barracks510: Ok. Awesome. Thanks.

models/git_diff.go
```diff
@@ -418,6 +418,50 @@ func GetDiffRange(repoPath, beforeCommitID string, afterCommitID string, maxLine
...
+	}
+
+	stderr := new(bytes.Buffer)
+
+	cmd.Dir = repoPath
+	cmd.Stderr = stderr
+
+	stdout, err := cmd.Output()
+	if err != nil {
+		return "", fmt.Errorf("Stdout: %v; Stderr: %s", err, stderr.String())
```
Unknwon: `err` is not `Stdout`, you can use `fmt.Errorf("%v - %s", err, stderr)`, `%s` calls `String` method automatically.

Unknwon:
I think we need @odinuge 's permission to hijack his PR.

odinuge:
Hi,
Removing my name from my commit is kinda strange.. I can update my PR with a new commit, squashed from all changes in this PR. (But i dont have time for the next 15 hours).

Unknwon:
I found a better way.

1. Save a copy of changed files.
2. Reset to `develop` and fetch commits made by @odinuge 
3. Copy back changed files and commit as @barracks510 

So no one loses credits. Sounds good?

odinuge:
Sounds good @Unknwon!

Unknwon:
@barracks510 if you also agree, please update this PR in proposed way^

barracks510:
Yes. That sounds good.

barracks510:
Sorry, I think I misunderstood. I still need @odinuge to update his patch to the latest develop branch to proceed with the proposed actions, otherwise, it still shows my current commit message.

Unknwon:
That's sad, ping @odinuge

odinuge:
Rebased my commit. See the topic-branch.

barracks510:
Ok. New PR based on Changes. See #3345.

PR #3301, opened 2016-07-21
avatar: make custom and generated avatars equal

barracks510:
Sets all avatars to use PNG image format.
Keeps avatars consistent at 290x290px resolution.


Unknwon:
Thanks!

PR #3330, opened 2016-07-26
routers/repo/release: Use correct branch reference

### LINKED ISSUE #3326: Incorrect calculation of commit count for releases created not on default branch
## Setup

- Gogs version: `v0.9.48`
- Git version: `2.7.4`
- Operating system: `Ubuntu 16.04 LTS`
- Database:
  - [ ] PostgreSQL
  - [ ] MySQL
  - [X] SQLite
  - [X] Yes (https://try.gogs.io/nalla/test-repo/releases)
  - [ ] No
  - [ ] Not relevant

## Description

When the `release.NumCommitsBehind` (`routers/repo/release.go`) are being calculated, the `countCache` is always using the `master` (default) branch as reference. In [`routers/repo/release.go#L32`](https://github.com/gogits/gogs/blob/master/routers/repo/release.go#L32) the `release.Target` is being looked up but the cache is being filled with a calculation that uses the default (`repoCtx.BranchName`) branch. See also [`routers/repo/release.go#L33`](https://github.com/gogits/gogs/blob/master/routers/repo/release.go#L33), [`routers/repo/release.go#L37`](https://github.com/gogits/gogs/blob/master/routers/repo/release.go#L37) and [`routers/repo/release.go#L42`](https://github.com/gogits/gogs/blob/master/routers/repo/release.go#L42).

If you look at the example repo release page https://try.gogs.io/nalla/test-repo/releases you will notice the following figues:

Tag | Commit | Branch | Commit Count (is) | Commit Count (should)
---- | ---------| -------- | -------------------- | --------------------------
2 | b5151cc0a4 | another | 2 | 1
1 | c4b66f39ec | master | 3 | 3

You can verify the should counts by issuing the following git commands
```
$> git clone https://try.gogs.io/nalla/test-repo.git
$> cd test-repo
$> git rev-list --count origin/master
4
$> git rev-list --count 1
1
$> git rev-list --count origin/another
3
$> git rev-list --count 2
2
```

nalla:
When calculating the current behind commit count the calculation should
use the current release target to get the total commit count.

This fixes #3326

### DIFF
--- a/routers/repo/release.go
+++ b/routers/repo/release.go
@@ -20,7 +20,7 @@ const (
 	RELEASE_NEW base.TplName = "repo/release/new"
 )
 
-// calReleaseNumCommitsBehind calculates given release has how many commits behind default branch.
+// calReleaseNumCommitsBehind calculates given release has how many commits behind release target.
 func calReleaseNumCommitsBehind(repoCtx *context.Repository, release *models.Release, countCache map[string]int64) error {
 	// Fast return if release target is same as default branch.
 	if repoCtx.BranchName == release.Target {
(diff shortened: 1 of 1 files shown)


Unknwon:
Hmm, what if the target branch is deleted, what happens?

nalla:
I think there are two valid options here:

1. use default branch as fallback 
2. use zero as fallback

I'd prefere option two. Since logic dictates that no further commits can happen to a deleted target. The curlpit is that if you have e.g. two releases on a deleted target logic dictates that the latest tag should be the `HEAD`. So option two might complicates things.

```
if target is default 
  use default 
if target exists
  use target
if target is deleted
  get newest release with target
  use commit count of that release 
```

Unknwon:
>use zero as fallback

Prefer this too, please update the PR add a branch exist check.

nalla:
@Unknwon I added the exists check and clarified the side effects of the implementation in the commit message.

Unknwon:
Thanks!

PR #3332, opened 2016-07-26
census of contributors sorted by organization February - July 2016

dachary:
This is a census of the code contributors affiliations over the past six
months. Each commit author was researched to figure out if her/his
contribution was sponsored by an organization (for profit or not).

The result of this research complements the information found in the git
log and is stored in the .organizationmap file. It uses the same format
as the standard .mailmap file which helps resolve duplicates due to
typos.  It can be used with git-check-mailmap(1) as documented in the
file itself.

The fictious "FLOSS Contributor" organization groups all contributors
that are not affiliated to any known organization.

This census covers a given time period and will not require any
maintenance because the git history is not meant to be
rewritten. Another census can be organized at a later time, to reflect
the evolution of contributors population, for another time period.

Signed-off-by: Loic Dachary <loic@dachary.org>

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,6 @@
+Antoine Girard <sapk@sapk.fr>
+Bart Louwers <bart_louwers@hotmail.com>
+Chris Watt <info@chriswatt.org>
+Dan Hearnah <dan.hearnah@gmail.com>
+Luke Stahlman <luke.stahlman@gmail.com>
+Sjon Hortensius <SjonHortensius@users.noreply.github.com>
--- a/.organizationmap
+++ b/.organizationmap
@@ -1,0 +1,87 @@
+#
+# This is a census of the contributors affiliations from February 2016 to July 2016.
+#
+# Number of contributors active in the past six months, per organization
+#
+# git log --no-merges --pretty='%aN <%aE>' --since 'February 2016' --until 'July 2016' | sort | uniq | git -c mailmap.file=.organizationmap check-mailmap --stdin | sort | uniq -c | sort -rn | nl
+#     1	     49 FLOSS Contributor <contact@floss.cc>
+#     2	      1 Unknwon <u@gogs.io>
+#
+# Number of commits in the past six months, per organization
+#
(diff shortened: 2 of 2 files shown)


dachary:
Unless I'm mistaken gogs contributors were not sponsored by any organization over the past six months (which sets a great example by the way ;-).

Unknwon:
Thanks your PR!

Why do we need this?

dachary:
It's a useful fact about gogs and once merged it will be the primary source of census information for https://www.wikidata.org/wiki/Q21091728 as documented at https://www.wikidata.org/wiki/Wikidata:WikiProject_Informatics/FLOSS#number_of_participants

Unknwon:
Thanks, if those info are already in the git history, why do we need another file to store it?

dachary:
The fact that all contributions were made on a voluntary basis is not in the git repository. If an organization was sponsoring some contributors, it would look more like https://github.com/loomio/loomio/blob/master/.organizationmap where you can see about half the developers are employed by the Loomio company.

Does that answer your question ?

Unknwon:
Thanks, but Gogs is not sponsored by any company. I think this project does not need this.

dachary:
You think this census is only for projects sponsored by companies ? I must have expressed myself very badly and I apologize for that. This census is for wikidata, which is the database used by wikipedia to store facts.

Unknwon:
By conclusion... I don't think this is useful for this project..

dachary:
I respect your decision. I spent time studying gogs and creating this census. I'm sure you would not want that work be thrown away. Where do you suggest I store these results ?

Unknwon:
I respect your time that is the reason I'm still commenting, but not every PR I have to accept because people sent them.

dachary:
I'm not disputing your decision :-) Just asking for advice about where I should stash this work.

Unknwon:
Honestly, I don't know, which is the exact reason we have [Gitter room](https://gitter.im/gogits/gogs) for you to jump in and discuss.

PR #3333, opened 2016-07-26
Added Full Name to CreateUser api call

richmahn:
Requires PR 36 from go-gogs-client, https://github.com/gogits/go-gogs-client/pull/36


Unknwon:
Thanks!

PR #3335, opened 2016-07-27
add runtime reload for TLS Certificate update

cs8425:
Add runtime reload support for TLS Certificate update.
Should be useful with letsencrypt.
Send `SIGUSR1` to trigger.

**HAVE NOT test on windows!!**

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -11,8 +11,10 @@ import (
 	"net/http"
 	"net/http/fcgi"
 	"os"
+	"os/signal"
 	"path"
 	"strings"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR! But why do we need this?

cs8425:
When using "Let's Encrypt" with automatically renew certificates,
this could let gogs reload certificate without restart the service.

Unknwon:
Thanks, but I don't think this is a particular useful feature for Gogs because although Gogs supports basic HTTPS if you expose Gogs directly in the public domain, but have a front proxy (NGINX, Apache2) is more recommended.

PR #3336, opened 2016-07-27
Added Units for repo, access and team. First step of three on #3027.

lunny:
Added Units for repo, access and team. First step of three on #3027.
This is a rePR for #3049. And #3049 will be closed.

### DIFF
--- a/models/access.go
+++ b/models/access.go
@@ -55,6 +55,7 @@ type Access struct {
 	UserID int64 `xorm:"UNIQUE(s)"`
 	RepoID int64 `xorm:"UNIQUE(s)"`
 	Mode   AccessMode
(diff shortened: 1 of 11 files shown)


lunny:
For the first step of #3027, I added units for repository, access and team. And it's completely compatible with old repos. And there are no any changes on User's UI.

The second, we can change repo's settings page to let user to choose which units admin to want. Code, Issues, Pulls, Commits, Releases, Wiki, Settings. So the repos' home page will dynamically show the units according the repo's settings.

The last step, we will change repo's access and team's access page to dynamically show the units according the user's access.

I have to change the migration for supporting set the default values to added columns on Sync2.

PR #3351, opened 2016-07-31
Lighter icon colors for repo files

silverwind:
Small visual change to file icon colors:

### Before
before

### After
after

### DIFF
--- a/public/css/gogs.css
+++ b/public/css/gogs.css
@@ -1248,9 +1248,13 @@ footer .ui.language .menu {
 .repository.file.list #repo-files-table tbody .octicon {
   margin-left: 3px;
   margin-right: 5px;
+  color: #777;
 }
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks your PR!

Can you give a more complete screenshot of full page?

silverwind:
Not much changed except the icon colors. The avatar rendering is a Firefox bug with the `screenshot` console commmand in case you're wondering :)

## before
before
## after
after

Unknwon:
>... if it is not related to any particular issues, explain why we should not reject your pull request.

silverwind:
I think you should accept it because I feel the current near-black icons feel a bit harsh to me, and because it's the same color scheme GitHub uses for repo icons.

silverwind:
Screenshot from GitHub for comparision:



I could change the shade of blue if you like. Currently I'm using the same shade that is used for links.

Unknwon:
OK, make sense.

Issue #2384, opened 2016-01-12
Feature request fail2ban


km4efp:
Enable user auth fail log for use with fail2ban in preventing brute force attacks

Unknwon:
Thanks your feedback!

> Enable user auth fail log for use with fail2ban

How does this work?

bkcsoft:
@Unknwon Basically the authentication route has to log failed login-attempts to file with the offending IP
http://www.fail2ban.org/wiki/index.php/Main_Page

actually it would be simplest if Macaron just printed the ip-address to the log as-well :smile:

Unknwon:
Thanks @bkcsoft !

Sounds like a simple integration (log-level) to fail2ban?

bkcsoft:
@Unknwon well, it's a simple integration, _if_ the logs from gogs printed out ip-addresses ;)

Unknwon:
Only IP addresses? To any file?

bkcsoft:
well, fail2ban needs a status-code and an IP... preferably turn this:
```
gogs | [Macaron] Completed /img/favicon.png 304 Not Modified in 91.047µs
```
into:
```
gogs | [Macaron] Completed /img/favicon.png 304 Not Modified in 91.047µs from 172.16.0.14
```

Unknwon:
Thanks. Does it have any so-called standard form of log?

wqweto:
@Unknwon `fail2ban` uses flexible set of filtering rules based on filter/ignore regexes e.g. something like this
```
failregex = (?i)[Macaron] Failed /login .* from <HOST>
```
. . . can be used to filter failed logon attempts *and* gather IP through \<HOST\> capture.

Unknwon:
Thanks @wqweto !

Unknwon:
I think it is now as easy as adding a [NewLogger](https://github.com/gogits/gogs/blob/master/modules%2Flog%2Flog.go#L22) with `file` type and filename `fail2ban.log`, and record failures in https://github.com/gogits/gogs/blob/master/routers%2Fuser%2Fauth.go#L63

AndreaGiardini:
I would love to see this in Gogs! I subscribe :+1:

AndreaGiardini:
I think #2291 is a duplicate

Unknwon:
@AndreaGiardini I think this one is the subset of #2291...

masupilamie:
I would also like to be able to use fail2ban with gogs

[...]

Issue #2291, opened 2015-12-26
login try limit agains bruteforce attack


usern283:

yurivict:
This is generally hard to do because the same brute-force attack can come from different IP addresses. If attackers use a bot, each hit will come from the different IP. If you will simply delay the login, such defense can cause anybody to lock anybody else out of the system by continuous brute-forcing.

Better use 2-factor authentication, like google two-factor authenticator through PAM.

usern283:
Yes, but I think that a brute-force attack is more difficult , if only 5 login attempt per IP are possible.

initbar:
I agree with @yurivict's insight: rather than placing an overhead of n-logins by the unique IP set (and flushing the log once the user successfully logs in), I think the delay with exponential growth time is more elegant solution.

stonerl:
Does Gogs currently log failed login attempts with the corresponding IP-Address? If so fail2ban could be used to take care of this.

yurivict:
Ban who, user-id? IP? Net?

stonerl:
@yurivict fail2ban bans IPs

stonerl:
In general I don't think that it is Gogs responsibility to to take care of this matter, there are tools like fai2ban that are designed to do exactly this single job. Gogs just needs to provide a sane log file that allows such tools to work properly.

yurivict:
agree. should just integrate gogs with fail2ban.

stonerl:
I just checked the logfiles. Currently gogs does not log failed login attempts.

ghevge:
Any updates on this topic? Thanks

PR #3357, opened 2016-08-01
Fail2ban with HTTP status and logger (#2291 #2384)

yweber:
I'm not sure if this patch is acceptable. I never wrote golang before but I was needing this features.
- introduce a new context function RenderWithErrStatus that allows to
  specify an HTTP status when rendering errors on form datas
  - status 403 set on repository permission denied
  - status 401 set on authentication failure
  - status 401 set when no associated email found on forgotten password
    procedure
- add warning log message, containing client IP address, for
  authentication failure
- add warning log message in routers/user/auth.go AutoSignIn function
- add http_status int in returned values for routers/user/auth.go
  AutoSignIn function
- change default log level to Info

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -137,13 +137,13 @@ func (c *Context) SubURLRedirect(location string, status ...int) {
 }
 
 // RenderWithErr used for page has form validation but need to prompt error to users.
-func (c *Context) RenderWithErr(msg, tpl string, f interface{}) {
+func (c *Context) RenderWithErr(msg, tpl string, f interface{}, http_status int) {
 	if f != nil {
 		form.Assign(f, c.Data)
 	}
 	c.Flash.ErrorMsg = msg
 	c.Data["Flash"] = c.Flash
-	c.HTML(http.StatusOK, tpl)
+	c.HTML(http_status, tpl)
 }
 
 // Handle handles and logs error by given status.
--- a/routes/admin/auths.go
+++ b/routes/admin/auths.go
@@ -6,6 +6,7 @@ package admin
 
 import (
 	"fmt"
+	"net/http"
 
 	"github.com/Unknwon/com"
 	"github.com/go-xorm/core"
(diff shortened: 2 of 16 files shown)


conf/app.ini
```diff
@@ -257,7 +257,7 @@ MODE = console
 ; Buffer length of channel, keep it as it is if you don't know what it is.
 BUFFER_LEN = 10000
 ; Either "Trace", "Debug", "Info", "Warn", "Error", "Critical", default is "Trace"
-LEVEL = Trace
+LEVEL = Info
```
Unknwon: Why change this line?

yweber: I was thinking that was a better default log level. But at this time I wasn't knowing I will push on develop branch, where Trace is indeed a relevant log level.  Do you want me to delete this change ?

Unknwon: Obviously you would have to delete this change..

rugk:
:+1: for this.
@yweber Do you also have a Fail2Ban RegExp for this?
If so you might also want to add it to the doc somewhere.

yweber:
@rugk Thx ! And on my installation I don't use fail2ban directly on gogs logfiles but on the nginx reverse-proxy access log ( I filter on HTTP status ). I have written and quickly tested this RegExp for gogs logfiles : `^[^\[]*\[W\] <HOST> Authentication failure for user .*$`

Writing this RegExp I figured out that on password retrieval failure (when password retrieval is asked for a non registered email address) no log message is produced and that a different HTTP status is returned (401 instead of 200). This behavior can allow an attacker to find if an email is registered or not.
What do you think ? Is it a better idea to invert this behavior, returning HTTP 200 and producing a log message containing client IP when password retrieval fails ?

rugk:
> Is it a better idea to invert this behavior, returning HTTP 200 and producing a log message containing client IP when password retrieval fails ?

I think in both cases either 401 or 403 should be returned and you can additionally write the log file.

Unknwon:
Please sign the CLA.

AndreaGiardini:
Any update on this?

dtrv:
+1

rugk:
Maybe @yweber could try to introduce this PR in the fork @gitea.

Fastidious:
I find it of poor taste, @rugk, to come to a project, and try to send people to another. You could have opened an issue on Gitea about this, and mention the OP in it.

rugk:
It does not really make a difference. Also it's @yweber' choice…

yweber:
The merge commit https://github.com/gogits/gogs/pull/3357/commits/47fc72f3d1422426866d430d806ac44d50081cdd is just about solving merge conflicts and making the compilation (and apparently the software) work :
I didn't take the time to check if some other 4xx, 5xx HTTP status or other log lines should be emitted with code changes. I also did not check if a function was implemented achieving the same goal than
[RenderWithErrStatus]( https://github.com/gogits/gogs/pull/3357/commits/47fc72f3d1422426866d430d806ac44d50081cdd#diff-94ab5308b984ba8eb1cb1aa88d0487d7R104 ).

fizzyade:
Why hasn't this made it into the source?  This is a useful feature.

PR #3225, opened 2016-07-01
Provide button to delete merged pull request branch

tanapoln:
I think it is tricky to clean-up branch of pull-request, for someone who implements new feature on new branch. Currently, you have to go back to your local machine and run a command to remove that branch which is very inconvenience.

This pull request add a button to every merged pull request that head branch still exist that allow user to push that button and delete head branch (see screenshot).

screen shot 2559-07-02 at 02 19 44

PS. Please merge this pull-request first, this feature depends on a change of git-module. https://github.com/gogits/git-module/pull/15

PR #3369, opened 2016-08-03
Implement branch deletion on PR merge (as an option). Closes #3201.

### LINKED ISSUE #3201: Add option to delete source branch after merge
It would be nice if the source branch could get deleted when a pull request is approved (as an option). Bitbucket & gitlab both offer this feature.

tomsid:
- adds a checkbox for branch deletion near "merge" button
selection_007

- the checkbox will be disabled if the head branch is the default branch of the repository.


Unknwon:
Somewhat duplicated to https://github.com/gogits/gogs/pull/3225?

kmadel:
I prefer the UX of #3225 - the checkbox is not very intuitive.

tomsid:
Indeed, didn't notice that PR.

In my opinion the ability to delete source branch should be given only at merging.

Lets consider an example:
1. User creates a branch, make changes and creates a PR.
2. PR merged without branch deletion
3. Later, the branch has been deleted manually.
4. Someone creates a branch with the same name (e.g. fix_issue) and make changes there.
5. An another contributor looks at merged PRs and see that the branch is still there. He thinks that the branch is stale and push's the button in order to delete this branch.
6. The branch with valuable changes is deleted.

So there is a possibility to delete a branch with new commits.

Unknwon:
@tomsid good point!

Gogs stores the commitID that is being merged: https://github.com/gogits/gogs/blob/master/models/pull.go#L60

So we can check before deletion, and returns error if new commits have pushed.

tomsid:
@Unknwon In that case the button is a good solution too.
So what solution do you prefer?

Unknwon:
I prefer the button solution which delete branch after merged.

tomsid:
Okay, then let's close that PR and move our discussion to #3225

PR #3374, opened 2016-08-04
Fix #3315: Release dont use tag creation date

### LINKED ISSUE #3315: Release dont use tag creation date
We DO NOT take questions or config/deploy problems on GitHub, please use our forum: https://discuss.gogs.io

Please take a moment to search that an issue doesn't already exist.

**For bug reports, please give the relevant info:**

- *Gogs version (or commit ref):* 0.9.48
- *Git version:* 1.8.3.1
- *Operating system:* Linux CentOS 7 64bit
- *Database:* SQLite
- *Can you reproduce the bug at http://try.gogs.io:* Yes --> https://try.gogs.io/zero-x-baadf00d/psychic-soul/releases (few hour ago)  and https://try.gogs.io/zero-x-baadf00d/psychic-soul/src/release/1.3.1  (6 months ago)

## Description

Release must take the linked tag creation date (like Github)

0xbaadf00d:
Like Github, release note now using the tag creation date rather than the web release note creation date (Issue #3315). If tag does not exist yet, last commit on selected branch will be used.

### DIFF
--- a/models/release.go
+++ b/models/release.go
@@ -40,10 +40,9 @@ type Release struct {
 
 func (r *Release) BeforeInsert() {
 	if r.CreatedUnix == 0 {
-                r.CreatedUnix = time.Now().Unix()
+		r.CreatedUnix = time.Now().Unix()
 	}
 }
-
 
 func (r *Release) AfterSet(colName string, _ xorm.Cell) {
 	switch colName {
--- a/models/release.go
+++ b/models/release.go
@@ -39,8 +39,11 @@ type Release struct {
 }
 
 func (r *Release) BeforeInsert() {
-	r.CreatedUnix = time.Now().Unix()
+	if r.CreatedUnix == 0 {
+                r.CreatedUnix = time.Now().Unix()
+	}
 }
+
 
 func (r *Release) AfterSet(colName string, _ xorm.Cell) {
 	switch colName {
--- a/routers/repo/release.go
+++ b/routers/repo/release.go
@@ -6,6 +6,7 @@ package repo
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/gogits/gogs/models"
 	"github.com/gogits/gogs/modules/auth"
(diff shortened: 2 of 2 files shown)


models/release.go
```diff
@@ -39,9 +39,12 @@ type Release struct {
 }
 
 func (r *Release) BeforeInsert() {
-	r.CreatedUnix = time.Now().Unix()
+	if r.CreatedUnix == 0 {
+                r.CreatedUnix = time.Now().Unix()
+	}
 }
 
+
```
Unknwon: >All .go files must be formatted with gofmt (most of Go editors should be already possible to format on save).

routers/repo/release.go
```diff
@@ -167,6 +168,17 @@ func NewReleasePost(ctx *context.Context, form auth.NewReleaseForm) {
 		return
 	}
 
+	tagCreationDate := time.Now()
```
Unknwon: You can store a `int64` type variable here by `var tagCreatedUnix int64`, so it defaults to `0`.

routers/repo/release.go
```diff
@@ -190,6 +202,8 @@ func NewReleasePost(ctx *context.Context, form auth.NewReleaseForm) {
 		Note:         form.Content,
 		IsDraft:      len(form.Draft) > 0,
 		IsPrerelease: form.Prerelease,
+		Created:      tagCreationDate,
+		CreatedUnix:  tagCreationDate.Unix(),
```
Unknwon: And here you only need to take care of `CreatedUnix`.

0xbaadf00d:
Thanks for your replies and tips. I will do modifications.

models/release.go
Unknwon: Didn't fix this.

routers/repo/release.go
```diff
@@ -260,6 +270,14 @@ func EditReleasePost(ctx *context.Context, form auth.EditReleaseForm) {
 		return
 	}
 
+	tag, err := ctx.Repo.GitRepo.GetTag(rel.TagName)
```
Unknwon: You didn't answer the reason why we need this.

0xbaadf00d: After doing more tests, you are right, this block is not necessary on `EditReleasePost` function. I will remove this today.

Unknwon:
One fix is leaking.

0xbaadf00d:
gofmt is now applied to all modified files (_models/release.go_ and _routers/repo/release.go_)

Unknwon:
Thanks!

PR #3386, opened 2016-08-05
Replace gogs.io http links with https version

rugk:
Nice you use HTTPS on your website, so now only the links to your website need to be updated.


Unknwon:
Thanks!

PR #3391, opened 2016-08-06
Add API endpoint for getting user by token

ethantkoenig:
This pull request adds a new API endpoint for getting a user from an access token. The endpoint is `GET user/`, and is analogous to the Github endpoint `GET user/` (https://developer.github.com/v3/users/#get-the-authenticated-user).

To the best of my knowledge, @Unknwon contacted @richmahn yesterday (08/06/2016) expressing interest in this feature.

### DIFF
--- a/routers/api/v1/user/user.go
+++ b/routers/api/v1/user/user.go
@@ -70,3 +70,8 @@ func GetInfo(ctx *context.APIContext) {
 	}
 	ctx.JSON(200, &api.User{u.ID, u.Name, u.FullName, u.Email, u.AvatarLink()})
 }
+
+func GetSignedInUser(ctx *context.APIContext) {
+	u := ctx.User;
+	ctx.JSON(200, &api.User{u.ID, u.Name, u.FullName, u.Email, u.AvatarLink()})
+}
--- a/routers/api/v1/api.go
+++ b/routers/api/v1/api.go
@@ -195,6 +195,7 @@ func RegisterRoutes(m *macaron.Macaron) {
 		}, reqToken())
 
 		m.Group("/user", func() {
+			m.Get("", user.GetSignedInUser)
 			m.Combo("/emails").Get(user.ListEmails).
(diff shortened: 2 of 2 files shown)


routers/api/v1/user/user.go
```diff
@@ -70,3 +70,13 @@ func GetInfo(ctx *context.APIContext) {
 	}
 	ctx.JSON(200, &api.User{u.ID, u.Name, u.FullName, u.Email, u.AvatarLink()})
 }
+
+func GetUser(ctx *context.APIContext) {
+	u := ctx.User
+
+	// Hide user e-mail when API caller isn't signed in.
+	if !ctx.IsSigned {
```
richmahn: Since this is returning the info of the signed in user, it probably doesn't need to check IsSigned.

richmahn:
@Unknwon Yes, as @ethantkoenig said in the description, he is one of our developers that make this endpoint. Do we want it to be more like the specs at https://developer.github.com/v3/users/#get-the-authenticated-user ?

routers/api/v1/user/user.go
```diff
@@ -70,3 +70,13 @@ func GetInfo(ctx *context.APIContext) {
 	}
 	ctx.JSON(200, &api.User{u.ID, u.Name, u.FullName, u.Email, u.AvatarLink()})
 }
+
+func GetUser(ctx *context.APIContext) {
```
richmahn: I think, to be clearer, this should be called GetAuthenticatedUser()

ethantkoenig:
Took both of @richmahn's suggestions above. Only alteration was renaming `GetUser(..)` as `GetSignedInUser(..)` (as opposed to `GetAuthenticatedUser(..)`), as the "SignedIn" terminology seems to fit better with the rest of the codebase (in particular with `modules/auth/auth.go`)

Unknwon:
Please sign the CLA.

Unknwon:
https://github.com/gogits/go-gogs-client/wiki/Users#get-the-authenticated-user

Issue #3401, opened 2016-08-08
Links of pull request comment email should use pulls URL


Unknwon:
Currently what we have are still linking to `/issues/:index`, instead we should have `/pulls/:index` to be consistent.

lstahlman:
Any preference on how it is implemented?

Some options:

- Handle everything in Issue.FullLink().  Pretty straightforward to do, but puts PR logic in Issue.
- Implement PullRequest.FullLink(), call instead of Issue.FullLink() if issue is a PR. More logical in some ways but also involves more code.

I have the second option working in my test environment.

Unknwon:
Hi, I think it is OK to just make a `if issue.IsPull` inside `Issue.FullLink()`, since many logics on issue and pulls are basically the same. Also this way has the minimal change.

PR #3403, opened 2016-08-09
Fix for #3401 - Links of pull request comment email should use pulls URL

lstahlman:
See discussion on #3401

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -160,7 +160,13 @@ func (i *Issue) State() string {
 }
 
 func (issue *Issue) FullLink() string {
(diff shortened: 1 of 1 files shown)


lstahlman:
I signed the CLA before submitting the PR, but the CLA check says otherwise. When I click details, it confirms I have indeed signed it.

Unknwon:
Never mind, this CLA service looks completely not working.

Unknwon:
Thanks!

Unknwon:
PS: I think CLA is not working because somehow you commited with Gogs(?), and that's not a valid user... just be careful with future commits.

lstahlman:
Yeah, the author on the commit was correct but committer was not. I updated my config, so future commits should show the correct info.

Issue #3398, opened 2016-08-08
Redefine global mirror interval


colundrum:
Hi,

Can you create a variable in the custom/conf/app.ini to redefine the value of Interval (we can see this value at https://github.com/gogits/gogs/blob/master/models/repo.go#L722) without recompiling Gogs.

I've already asked to the discuss : [Is it possible to redefine the global mirror interval?](https://discuss.gogs.io/t/is-it-possible-to-redefine-the-global-mirror-interval/540)

Thanks

- Gogs version (or commit ref): 0.9.48.0722
- Git version: 2.1.4
- Operating system: Debian Jessie
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log gist:

0xbaadf00d:
I can implement this issue if no one already took it.

@Unknwon Do you have any advice for the key name and on what category put it ?

Unknwon:
@0xbaadf00d thanks, maybe `DEFAULT_INTERVAL` is more obvious?

0xbaadf00d:
Of course, I'll make the changes in the evening.

Unknwon:
Patch has pushed to fix this issue, please test on `develop` branch or https://try.gogs.io.

PR #3409, opened 2016-08-09
Feature #3398: Redefine global mirror interval

0xbaadf00d:
This pull request add new key on `app.ini` to configure the mirror global interval (#3398).

By example, to set mirror global interval to 6 hours rather than 24 :
```
[mirror]
GLOBAL_INTERVAL = 6
```

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -358,6 +358,10 @@ MIRROR = 300
 CLONE = 300
 PULL = 300
 
+[mirror]
+; Default interval in hours between each check
+DEFAULT_INTERVAL = 24
+
 [api]
 ; Max number of items will response in a page
 MAX_RESPONSE_ITEMS = 50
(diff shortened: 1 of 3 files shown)


Unknwon:
Commented on #3398.

0xbaadf00d:
The variableis now renamed to `DEFAULT_INTERVAL`

Unknwon:
I think the `Mirror` should put after [this line](https://github.com/0xbaadf00d/gogs/blob/bb3732e082a897bc5ef049de3b1c2bb563cd7306/modules/setting/setting.go#L208), and load with another `else if` before [api](https://github.com/0xbaadf00d/gogs/blob/bb3732e082a897bc5ef049de3b1c2bb563cd7306/modules/setting/setting.go#L499) section. This also requires you add a `[mirror]` section after `[git]` in `conf/app.ini` file.

Unknwon:
Thanks!

Issue #3410, opened 2016-08-09
overall issues for organisations is limited to num_repos from the user


rikvdh:
- Gogs version (or commit ref):  0.9.60.0803
- Git version: 2.1.4
- Operating system: Linux Debian
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (provide example URL) `not possible due to user requirement`
  - [ ] No
  - [ ] Not relevant
- Log gist:

## Description

The list of issues for an organization is incomplete when the user itself has less repositories than the organisation. In the query the `LIMIT X` is used where X is filled as the `num_repos` from the `user` table. 

To reproduce:
- Create organisation
- Add 1 repository
- Add 1 issue
- Make sure the currently logged-in user has no repositories
- Go to issues at the top
- View the organisation
- no issues are displayed *(!!)*
- Create a repo for the user
- Go to issues at the top
- View the organisation
- The issue is displayed *(!!)*

For now I workaround this issue by setting `num_repos` in the user table to 999 for each user. But this isn't a good way to solve this.

lstahlman:
I am able to reproduce this.

Unknwon:
Patch has pushed to fix this issue, please test on `develop` branch or https://try.gogs.io.

rikvdh:
Thanks @Unknwon it is fixed on https://try.gogs.io. My test user now has all the issues I expect to be there! :D

For our own installation we wait for the tagged releases before upgrade. 

Thanks!

xor-gate:
Nice work all for catching and fixing!

PR #3412, opened 2016-08-09
Fix for #3410 overall issues for organisations is limited to num_repo…

lstahlman:
Changed a reference to RepoNum so it is based on the org's repo count instead of the user's.

See #3410


Unknwon:
Thanks!

PR #3418, opened 2016-08-10
Fix #3321: commit tag shortener

### LINKED ISSUE #3321: Short commit references not turning into links.
See https://git.osgeo.org/gogs/geos/php-geos/issues/6#issuecomment-182 and subsequent comment.
The first comment used ca42d272d8 as a commit reference, the second used the full hash and the full url.
Only the second comment turned both references into links, while the first did not.

0xbaadf00d:
Short commit references now working with a 7 to 40 SHA length'  (Issue #3321) and base.ShortSha() function has been modified to uniformize short SHA's representation (length: 7).

### DIFF
--- a/modules/base/tool.go
+++ b/modules/base/tool.go
@@ -45,8 +45,8 @@ func EncodeSha1(str string) string {
 }
 
 func ShortSha(sha1 string) string {
-	if len(sha1) == 40 {
-		return sha1[:10]
+	if len(sha1) > 7 {
+		return sha1[:7]
 	}
 	return sha1
 }
--- a/modules/base/tool.go
+++ b/modules/base/tool.go
@@ -45,8 +45,8 @@ func EncodeSha1(str string) string {
 }
 
 func ShortSha(sha1 string) string {
-	if len(sha1) > 7 {
-		return sha1[:7]
+	if len(sha1) > 10 {
+		return sha1[:10]
 	}
 	return sha1
 }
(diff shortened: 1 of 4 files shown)


modules/base/tool.go
```diff
@@ -45,8 +45,8 @@ func EncodeSha1(str string) string {
 }
 
 func ShortSha(sha1 string) string {
-	if len(sha1) == 40 {
```
Unknwon: We can allow 7-char SHA, does not mean we want to show with 7-char in default, please remove this change.

modules/markdown/markdown_test.go
```diff
@@ -290,8 +290,8 @@ func TestMarkdown(t *testing.T) {
 		Convey("To the internal issue tracker", func() {
 			Convey("It should correctly convert URLs", func() {
 				testCases := []string{
-					"http://localhost:3000/user/project/commit/d8a994ef243349f321568f9e36d5c3f444b99cae", " <code><a href=\"http://localhost:3000/user/project/commit/d8a994ef243349f321568f9e36d5c3f444b99cae\">d8a994ef24</a></code>",
```
Unknwon: We can allow 7-char SHA, does not mean we want to show with 7-char in default, please adjust this change.

modules/base/tool.go
0xbaadf00d: I have done this to always have uniform short tag on all pages regardless of what users use (7-char, 8-char, ..., 40-char commit  SHA).

Ok if you want remove this; I will do needed modifications tonight.

Unknwon: Yes, please do the modification, we still want to show 10-char on the UI. Thought people could just use 7-char in other cases.

Unknwon:
Thanks!

PR #3423, opened 2016-08-10
Fix #3379: syntax error at or near

### LINKED ISSUE #3379: 500 when user leaves organization without relations to any repositories
- Gogs version (or commit ref): 0.9.58.0726
- Git version: 2.1.4
- Operating system: debian
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
  - [ ] No
  - [x] Not relevant
- Log gist:

## Description

I try leave an organization and the result is:
`{"err":"pq: syntax error at or near \")\"","ok":false}`

Maybe help this thread:
http://stackoverflow.com/questions/21073650/golang-pq-syntax-error-when-executing-sql

0xbaadf00d:
Fix issue when user want leave an organization (issue #3379) and don't have repositories.

Fixed by just adding a check before using the SQL statement IN who can't be used with empty list.

### DIFF
--- a/models/org.go
+++ b/models/org.go
@@ -414,8 +414,11 @@ func RemoveOrgUser(orgID, userID int64) error {
 			return err
 		}
 	}
-	if _, err = sess.Where("user_id = ?", user.ID).In("repo_id", repoIDs).Delete(new(Access)); err != nil {
-		return err
+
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

Issue #3396, opened 2016-08-08
Pull request conflict status not updating properly


duwial:
- Gogs version (or commit ref): 0.9.64.0805
- Git version: 1.7.1
- Operating system: Linux (Scientific Linux release 6.8)
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (provide example URL)
  - [ ] No
  - [ ] Not relevant
- Log gist:

## Description

I've got two pull requests in my project, both referencing the same file. When I've merged the first pull request I'm expecting the second not to be able to be merged. However, when going to the pull request I can still see the text "This pull request can be merged automatically." while it clearly can't be merged. When pushing the button "Merge pull request" I'm getting a 500 error with the following appearing in our log file:

`[...routers/repo/pull.go:415 MergePullRequest()] [E] Merge: git merge --no-ff --no-commit [data/tmp/repos/435232857.git]: exit status 1 -`

I've created two pull requests showing this behavior on try.gogs.io:
https://try.gogs.io/xStoryTeller/ErrorTest/pulls/2 (successfully merged)
https://try.gogs.io/xStoryTeller/ErrorTest/pulls/1 (the one generating the error)

duwial:
A little update: it looks like the trigger task is not being executed when a pull request is being merged from the web interface. Is this intended? If I do a manual commit and push to the project all open pull requests are being re-evaluated and updated with the proper status.

PR #3425, opened 2016-08-10
Fix for #3396 - Pull request conflict status not updating properly

lstahlman:
Fix for #3396.  AddTestPullRequestTask needed to be called from PullRequest.Merge.


Unknwon:
Thanks!

PR #3434, opened 2016-08-11
Fix #3314: Cannot edit release with tag name contains slash

### LINKED ISSUE #3314: Cannot edit release with tag name contains slash
We DO NOT take questions or config/deploy problems on GitHub, please use our forum: https://discuss.gogs.io

Please take a moment to search that an issue doesn't already exist.

**For bug reports, please give the relevant info:**

- *Gogs version (or commit ref):* 0.9.48
- *Git version:* 1.8.3.1
- *Operating system:* Linux CentOS 7 64bit
- *Database:* SQLite
- *Can you reproduce the bug at http://try.gogs.io:* Yes --> https://try.gogs.io/zero-x-baadf00d/psychic-soul/releases

## How to reproduce

1. Create a repository
2. Do some commit and create tag "release/xxxxx"
3. Create a release on gogs
4. Click "publish" or "save draft"
5. Back to releases
6. Click "edit" (server ask for basic auth - see screenshoot)
7. Oops: 404 Error

capture

...

0xbaadf00d:
This pull request aim to fix issue #3314 with non flat tagname (ie: release/x.x.x) leading to a 404 error when user trying to edit it.

I have used this solution rather than using `EscapePound` because replacing _/_ with _%2F_ seems not working. Macaron still treat _%2F_ as _/_ before process routing and finally returning a 404 error (I don't even reach the function EditReleasePost(ctx *context.Context, form auth.EditReleaseForm).

__Demonstration:__
https://try.gogs.io/zero-x-baadf00d/psychic-soul/releases/edit/release%2f1.3.1

I'm not sure is a good idea, for URI readability and SEO, to use url encoding on path components. It legit fo query string arguments, but not sure for path.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -485,8 +485,8 @@ func runWeb(ctx *cli.Context) error {
 		m.Group("/releases", func() {
 			m.Get("/new", repo.NewRelease)
 			m.Post("/new", bindIgnErr(auth.NewReleaseForm{}), repo.NewReleasePost)
-			m.Get("/edit/:tagname", repo.EditRelease)
-			m.Post("/edit/:tagname", bindIgnErr(auth.EditReleaseForm{}), repo.EditReleasePost)
+			m.Get("/edit/*", repo.EditRelease)
+			m.Post("/edit/*", bindIgnErr(auth.EditReleaseForm{}), repo.EditReleasePost)
 			m.Post("/delete", repo.DeleteRelease)
 		}, reqRepoWriter, context.RepoRef())
 
--- a/routers/repo/release.go
+++ b/routers/repo/release.go
@@ -214,7 +214,7 @@ func EditRelease(ctx *context.Context) {
 	ctx.Data["PageIsReleaseList"] = true
 	ctx.Data["PageIsEditRelease"] = true
 
-	tagName := ctx.Params(":tagname")
+	tagName := ctx.Params("*")
 	rel, err := models.GetRelease(ctx.Repo.Repository.ID, tagName)
 	if err != nil {
 		if models.IsErrReleaseNotExist(err) {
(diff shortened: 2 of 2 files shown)


Unknwon:
I think this PR is not completed.. you only changed router and handler is not going to work.

0xbaadf00d:
This PR add a workaround to the macaron routing system who treat slash as separator if you use a simple placeholder (https://go-macaron.com/docs/middlewares/routing#globs).

I use this version on version installed on my own server since few days, I can create (tag if not exists, are correctly created), edit and delete release containing slash or other chars accepted in tag name (https://git-scm.com/docs/git-check-ref-format). If you want, I can create you an account for testing ?

Of course, if you have another solution to handle "/" in macaron routing system, I will do needed modifications with pleasure.

Unknwon:
Handler is clearly looking for `:tagname` not `*`.. https://github.com/gogits/gogs/blob/ee28fd92551961070cff4fe1b37ab94268c2f111/routers/repo/release.go#L242

0xbaadf00d:
I'm agree with you, and this PR change `:tagname` with `*` on these files:
- cmd/web.go
- routers/repo/release.go  (now `tagName := ctx.Params("*")`)

I tried to keep `:tagname` (for a better code readability), but, according to macaron documentation, is not possible to use globing and placeholder together. By example:
- `/edit/:tagname`  Only handle alphanumeric char and some other char(ie: - _)
- `/edit/*` Handle all tag name possibilities
- `/edit/:tagname(*)` Don't compile
- `/edit/:tagname(.*)` Like first item, Only handle alphanumeric char and some other char(ie: - _)

Unknwon:
Oops, I must be drunk, mislooked.

PR #3435, opened 2016-08-11
Always set redirect_to on header-provided sign-in

### LINKED ISSUE #3089: landing page lost after login
Signing-in after a 404 response (due to attempts to access a private repo w/out being signed in) takes you to the dashboard rather than back to where you started the navigation (ex 404 page, which would be visible after sign-in).

strk:
WIP

This is an attempt to fix #3089 following the strategy introduced
in d625e41c6c4d40fadbeaf99dd97e77aa0c9085ae, although that strategy
by itself does not seem to be sufficient.

What needs be done is honouring the redirect_url query parameter
from the auth router.

\cc @Unknwon

### DIFF
--- a/templates/base/head.tmpl
+++ b/templates/base/head.tmpl
@@ -139,7 +139,7 @@
 												<i class="octicon octicon-person"></i> {{.i18n.Tr "register"}}
 											</a>
 										{{end}}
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks, I think we also need auto-sign handle `redirect_to` in this form.

Issue #2960, opened 2016-04-11
Qualify cross-repository ticket references


strk:
As of 0.9.13.0318 referencing a ticket of a different repository results in rendering just the target ticket number w/out the target ticket repository name. It is confusing, IMHO.
On github those kind of references are rendered as `org/repo#ticketnum` instead.

Unknwon:
Can you given examples on https://try.gogs.io?

strk:
https://try.gogs.io/strk/test2/issues/1 -- see how the reference to `#3` does not include the repository name (`kalxas/OSGeoLive`)

strk:
Another test (so it doesn't point to a 404): https://try.gogs.io/strk/test1/issues/1

it would be also nice to get a back-reference, but I guess that's for another ticket ?

lstahlman:
I can pick this up.

Unknwon:
Fixed by #3436.

PR #3436, opened 2016-08-11
Improve cross-repository issue references in markdown

lstahlman:
The following changes are made to the way cross-repository issue references are handled:
-Short references can be used now e.g., `gogits/gogs#24`
-Rendering also includes the repo reference (addresses #2960). This applies to both short references and full URL references

### DIFF
--- a/modules/markdown/markdown.go
+++ b/modules/markdown/markdown.go
@@ -88,6 +88,9 @@ var (
 	IssueNumericPattern = regexp.MustCompile(`( |^|\()#[0-9]+\b`)
 	// IssueAlphanumericPattern matches string that references to an alphanumeric issue, e.g. ABC-1234
 	IssueAlphanumericPattern = regexp.MustCompile(`( |^|\()[A-Z]{1,10}-[1-9][0-9]*\b`)
+	// CrossReferenceIssueNumericPattern matches string that references a numeric issue in a difference repository
+	// e.g. gogits/gogs#12345
+	CrossReferenceIssueNumericPattern = regexp.MustCompile(`( |^)[0-9a-zA-Z]+/[0-9a-zA-Z]+#[0-9]+\b`)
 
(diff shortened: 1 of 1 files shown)


modules/markdown/markdown.go
```diff
@@ -258,6 +273,24 @@ func RenderIssueIndexPattern(rawBytes []byte, urlPrefix string, metas map[string
 	return rawBytes
 }
 
+// RenderCrossReferenceIssueIndexPattern renders issue indexes from other repositories to corresponding links.
+func RenderCrossReferenceIssueIndexPattern(rawBytes []byte, urlPrefix string, metas map[string]string) []byte {
+	pattern := CrossReferenceIssueNumericPattern
```
Unknwon: Why need a new variable?

modules/markdown/markdown.go
```diff
@@ -151,7 +154,19 @@ func (r *Renderer) AutoLink(out *bytes.Buffer, link []byte, kind int) {
 			if j == -1 {
 				j = len(m)
 			}
-			out.WriteString(fmt.Sprintf(`<a href="%s">#%s</a>`, m, base.ShortSha(string(m[i+7:j]))))
+
+			issue := string(m[i+7 : j])
+			fullRepoUrl := string(setting.AppUrl + strings.TrimPrefix(r.urlPrefix, "/"))
```
Unknwon: `string()` is redundant.

lstahlman:
Amended to address feedback.

bkcsoft:
Is it just me, or have you forgotten about alphanumeric issues? :smiley:

Unknwon:
Cross ref basically is for Gogs, and Gogs does not use alphanumeric format for issues, I think.

lstahlman:
@bkcsoft I did look at the alphanumeric pattern, but noticed it's only used when the meta style is set by Repository.ComposeMetas, which only sets it external issue trackers.

Unknwon:
I think repo and user name you should also allow `_` and `.` in regexp.

Unknwon:
Thanks!

PR #3439, opened 2016-08-11
Add Content-Disposition header for attachment downloads

andreynering:
This HTTP header tells the browser the filename it should use for downloads.

Today, if you open an image attached in an issue and hit `CTRL+S` to save, the default name would be something like `3adb8e12-4df8-44d4-bba8-d989ea16e58f.jpg`.

With this change, it should be the original filename, something like `my-image-file.jpg`.


Unknwon:
Thanks!

PR #3442, opened 2016-08-12
Fix #3437: Cannot connect to PostgreSQL via IPv6 address

### LINKED ISSUE #3437: Cannot connect to PostgreSQL via IPv6 address
Directive `HOST` in section `[database]` of `app.ini` doesn't support IPv6 addresses which makes Gogs fail, at least when PostgreSQL is used.

When IPv6 addresses get stated without brackets `gogs.log` reads
> [...s/routers/install.go:71 GlobalInit()] [E] Fail to initialize ORM engine: migrate: sync: dial tcp 0.0.7.211:74: connect: invalid argument

When brackets are used there are messages like this
> [...s/routers/install.go:71 GlobalInit()] [E] Fail to initialize ORM engine: Fail to connect to database: parse postgres://\<user\>:\<password\>@[fd99:aaaa/gogs1?sslmode=disable: missing ']' in host

connecting to a ULA `fd99:aaaa:bbbb:...` which all in all suggest there's something wrong with the parsing of the addresses.
Both FQDNs and IPv4 addresses are accepted. The connection to the database server works flawlessly via IPv6 as can be seen by setting the hostname in `HOST` of app.ini and configuring the database server to accept corresponding IPv6 connections only.
On a side note directive `HOST_ADDR` in `[server]` accepts IPv6 addresses, too.

Seen running b756806 on Arch Linux, hence PostgreSQL 9.5.3.

0xbaadf00d:
This PR add ability to handle IPv6 address with PostgreSQL (issue #3437). Now `HOST` can be filled with:

- `HOST = 127.0.0.1:5432`
- `HOST = 127.0.0.1`  // Use default port
- `HOST = [::1]`
- `HOST = [::1]:5432`
- `HOST = /tmp/pg.sock`
- `HOST = /tmp/pg.sock:5342`

_Tested on CentOS 7 with PostgreSQL 9.5_

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -114,15 +114,14 @@ func getEngine() (*xorm.Engine, error) {
 		}
 	case "postgres":
 		host, port := "127.0.0.1", "5432"
-		fields := strings.Split(DbCfg.Host, ":")
-		if len(fields) > 0 && len(strings.TrimSpace(fields[0])) > 0 {
-			host = fields[0]
-		}
-		if len(fields) > 1 && len(strings.TrimSpace(fields[1])) > 0 {
-			port = fields[1]
+		if strings.Contains(DbCfg.Host, ":") && !strings.HasSuffix(DbCfg.Host, "]") {
+			host = DbCfg.Host[0:strings.LastIndex(DbCfg.Host, ":")]
+			port = DbCfg.Host[strings.LastIndex(DbCfg.Host, ":")+1:]
+		} else if len(DbCfg.Host) > 0 {
+			host = DbCfg.Host
 		}
 
-		if host[0] == '/' { // looks like a unix socket
+		if DbCfg.Host[0] == '/' { // looks like a unix socket
 			connStr = fmt.Sprintf("postgres://%s:%s@:%s/%s%ssslmode=%s&host=%s",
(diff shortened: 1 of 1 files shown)


models/models.go
```diff
@@ -114,12 +114,11 @@ func getEngine() (*xorm.Engine, error) {
...
 	case "postgres":
 		host, port := "127.0.0.1", "5432"
-		fields := strings.Split(DbCfg.Host, ":")
-		if len(fields) > 0 && len(strings.TrimSpace(fields[0])) > 0 {
-			host = fields[0]
-		}
-		if len(fields) > 1 && len(strings.TrimSpace(fields[1])) > 0 {
-			port = fields[1]
+		if strings.Contains(DbCfg.Host, ":") && !strings.HasSuffix(DbCfg.Host, "]") {
+			host = DbCfg.Host[0:strings.LastIndex(DbCfg.Host, ":")]
```
Unknwon: 1. `0` is redundant.
2. We should save result of `strings.LastIndex(DbCfg.Host, ":")` to a variable and reuse.

0xbaadf00d: ok, thanks for tip.

PR #3448, opened 2016-08-12
Honour redirect_to query parameter

### LINKED ISSUE #3089: landing page lost after login

strk:
Closes #3089
Complements 3380c946e12e53368f28a2b7285c7bad0c1746dc

### DIFF
--- a/routers/user/auth.go
+++ b/routers/user/auth.go
@@ -76,18 +76,21 @@ func SignIn(ctx *context.Context) {
 		return
 	}
 
+	redirectTo := ctx.Query("redirect_to")
+	if len(redirectTo) > 0 {
+		ctx.SetCookie("redirect_to", redirectTo, 0, setting.AppSubUrl);
+	} else {
+		redirectTo, _ = url.QueryUnescape(ctx.GetCookie("redirect_to"))
+	}
+
 	if isSucceed {
-		if redirectTo, _ := url.QueryUnescape(ctx.GetCookie("redirect_to")); len(redirectTo) > 0 {
+		if len(redirectTo) > 0 {
 			ctx.SetCookie("redirect_to", "", -1, setting.AppSubUrl)
 			ctx.Redirect(redirectTo)
 		} else {
 			ctx.Redirect(setting.AppSubUrl + "/")
 		}
 		return
-	}
-
-	if redirectTo := ctx.Query("redirect_to"); len(redirectTo) > 0 {
-		ctx.SetCookie("redirect_to", redirectTo, 0, setting.AppSubUrl);
 	}
 
 	ctx.HTML(200, SIGNIN)
--- a/routers/user/auth.go
+++ b/routers/user/auth.go
@@ -86,6 +86,10 @@ func SignIn(ctx *context.Context) {
 		return
 	}
 
+	if redirectTo := ctx.Query("redirect_to"); len(redirectTo) > 0 {
+		ctx.SetCookie("redirect_to", redirectTo, 0, setting.AppSubUrl);
+	}
+
 	ctx.HTML(200, SIGNIN)
 }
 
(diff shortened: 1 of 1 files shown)


Unknwon:
This check should be added in the same place as https://github.com/strk/gogs/blob/1b2ae57fbca75146069b6252b89ec0d043d3d8ba/routers/user/auth.go#L80 and do check for both of methods to determine if there is a `redirect_to`. URL query parameter has higher priority than cookie.

strk:
On August 12, 2016 6:51:17 PM CEST, "无闻" <notifications@github.com> wrote:
>This check should be added in the same place as
>https://github.com/strk/gogs/blob/1b2ae57fbca75146069b6252b89ec0d043d3d8ba/routers/user/auth.go#L80
>and do check for both of methods to determine if there is a
>`redirect_to`. URL query parameter has higher priority than cookie.

In my tests that block is never reached
--strk;

Unknwon:
This does not change the fact we should put check together.

strk:
I'm not sure aba2d59 does what you ask

strk:
@Unknwon could you please review the current code to see if your "pu check togheter" request is satisfied ? I think it is...

routers/user/auth.go
```diff
@@ -76,8 +76,15 @@ func SignIn(ctx *context.Context) {
 		return
 	}
 
+	redirectTo := ctx.Query("redirect_to")
+	if len(redirectTo) > 0 {
```
Unknwon: Why we need to set cookie if we're going to clear in the same request?

strk: On Thu, Aug 25, 2016 at 05:47:55PM -0700, 无闻 wrote:
> > @@ -76,8 +76,15 @@ func SignIn(ctx *context.Context) {
> >  		return
> >  	}
> >  
> > +	redirectTo := ctx.Query("redirect_to")
> > +	if len(redirectTo) > 0 {
> 
> Why we need to set cookie if we're going to clear in the same request?

As I mentioned, I've seen isSucceed being often false (I could not
figure out how to make it return true).

In that case, it looks like the cookie is used on subsequent
call to perform the redirection.

routers/user/auth.go
```diff
@@ -120,6 +127,11 @@ func SignInPost(ctx *context.Context, form auth.SignInForm) {
 	// Clear whatever CSRF has right now, force to generate a new one
 	ctx.SetCookie(setting.CSRFCookieName, "", -1, setting.AppSubUrl)
 
+	if redirectTo := ctx.Query("redirect_to"); len(redirectTo) > 0 {
```
Unknwon: Post request does not carry `redirect_to` URL query in any rate, so it is not useful here.

strk: Have you actually tried removing it and doing the sign-in from a private
repository with 404 response ?


On Sat, Aug 27, 2016 at 01:57:22PM -0700, 无闻 wrote:
> > @@ -120,6 +127,11 @@ func SignInPost(ctx *context.Context, form auth.SignInForm) {
> >  	// Clear whatever CSRF has right now, force to generate a new one
> >  	ctx.SetCookie(setting.CSRFCookieName, "", -1, setting.AppSubUrl)
> >  
> > +	if redirectTo := ctx.Query("redirect_to"); len(redirectTo) > 0 {
> 
> Post request does not carry `redirect_to` URL query in any rate, so it is not useful here.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/gogits/gogs/pull/3448/files/01cfe12dd5ed2e16cfc6b0155759bc42cd15ca11#r76522277

-- 

  ()   Free GIS & Flash consultant/developer
  /\   https://strk.kbt.io/services.html

Unknwon: You should not be able to do sign in from a private repo because you can't even see the repository before you sign in.

strk: What happens is:

 1) URL to a private repository is advertised on a mailing list
 2) Legit users click on the link, with a browser having no login cookie
 3) Legit user get a 404 response, with a "sign-in" link on the top-right corner
 4) Legit user performs the sign-in

Right now the above results in loosing the URL of step 1.
This PR fixes that.

Unknwon: I don't understand, the `sign in` button already contains `redirect_to` link and set to cookie in `func SignIn(ctx *context.Context)`

strk: I don't understand either. All I know is it takes this code to make
the redirect work. There are too many layers I don't fully understand
to take care of cleanups / refactoring.

Unknwon:
Tested and worked without second part.

PR #3449, opened 2016-08-12
Do not change page on logout

strk:
Complements the fix to #3089 (for the logout action)


Unknwon:
It is not guaranteed you still have access to same page after logout.

strk:
On August 12, 2016 6:48:27 PM CEST, "无闻" <notifications@github.com> wrote:
>It is not guaranteed you still have access to same page after logout.

Yes, and it may be confusing. But so it is when the url to an unaccessible repo is given. I think the confusion should be reduced enhancing the response page, as it currently is a 404 not found wheread it should be so mm ething like "forbidden" instead. It is good to stay on the same psge because then a further login would make such page accessible again
--strk;

Unknwon:
It is designed to show 404.

strk:
For the record, it looks like GitHub behaves exactly as Gogs on signout, that is, it always redirects to the homepage. If Gogs wants to behave exactly the same, this PR can be closed (I don't have a major problem with that)

bkcsoft:
For the record, so does GitLab and BitBucket as well :)

PR #3462, opened 2016-08-15
Fix #3461: Avoid reading file from other directory (security issue)

### LINKED ISSUE #3461: Security issue on repository creation
- Gogs version (or commit ref): 0.9.77.0814
- Operating system: CentOS 7 / Ubuntu 16.04
  - [x] Yes ~~(https://try.gogs.io/zero-x-baadf00d/test)~~ _force deletion by @unknwon_
  - [ ] No
  - [ ] Not relevant

## Description

Is it possible, for a attacker, to edit HTML (ie: with Chrome developper tools) and change value in form for the _Readme_ and _Licence_ to use.

By example:

If we put __../app.ini__ rather than __Default__ for the Readme. The created repository will have a Readme file containing the content of `app.ini`

0xbaadf00d:
This pull request avoid attacker to use file from other directory by removing ability to move to other directory (issu #3461) .

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -870,7 +870,10 @@ type CreateRepoOptions struct {
 }
 
 func getRepoInitFile(tp, name string) ([]byte, error) {
-	relPath := path.Join("conf", tp, strings.Replace(name, ".", "", -1))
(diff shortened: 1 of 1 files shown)


Unknwon:
Hmm, ... have pushed solution already before saw this..

PR #3473, opened 2016-08-17
Fix #3361: Dumps are created world readable

### LINKED ISSUE #3361: Dumps are created world readable
- Gogs version (or commit ref): 0.9.48
- Git version: 2.1.4
- Operating system: Linux
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
  - [ ] No
  - [x] Not relevant
- Log gist:

## Description

The dump files created by gogs have permissions directly derived from the umask, so 0644 aka word readable, by default. The dumps include private information like database credentials and repositories.

0xbaadf00d:
Permission mask of created dumps are now 0600 (-rw-------) rather than 0644 (-rw-rw-r--). As @waldiTM say on the issue #3361,  dump files include private information like database credentials and repositories, access must be limited.

### DIFF
--- a/cmd/dump.go
+++ b/cmd/dump.go
@@ -97,7 +97,7 @@ func runDump(ctx *cli.Context) error {
 	}
 
 	if err := os.Chmod(fileName, 0600); err != nil {
-		log.Printf("Can't change file access permission to 0600: %v", err)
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

Issue #3120, opened 2016-05-23
PID file


captainark:
Hi,

I use [Monit](https://mmonit.com/monit/) to monitor daemons running on my systems. It needs a PID file to be able to check whether a process is running or not.

Could you please provide an option to generate a PID file when gogs starts ?

Thanks.

stevenroose:
@captainark I don't think it is up to Gogs to create PID files. I suggest using [`start-stop-daemon`](http://manpages.ubuntu.com/manpages/trusty/man8/start-stop-daemon.8.html) to maintain and monitor daemons and their PID files.

captainark:
Most apps that run as a daemon do provide a way to write a PID file when they start.

I can't use `start-stop-daemon` since I run systemd.

For the time being, I've added the following in my gogs.service unit file : 
```
ExecStartPre=/bin/bash -c "/bin/rm -f /run/gogs/gogs.pid"
ExecStartPost=/bin/bash -c "/usr/bin/pgrep -fx '/opt/gogs/gogs web' > /run/gogs/gogs.pid"
ExecStopPost=/bin/bash -c "[[ -f /run/gogs/gogs.pid ]] && /bin/rm /run/gogs/gogs.pid"
```

I've also created a `/etc/tmpfiles.d/gogs.conf` file to manage the `/run/gogs` folder creation when the server reboots : 
```
d /run/gogs  0755  git  git  -
```

I still hope PID file management will get implemented in a future version of gogs, though ;)

Unknwon:
How to generate a PID file?

captainark:
Maybe you could use [this package](https://github.com/facebookgo/pidfile) ?

Marco01809:
@captainark Putting
```
ExecStartPost=/bin/bash -c 'echo $MAINPID > /run/gogs.pid'
PIDFile=/run/gogs.pid
```
into the systemd unit file should be sufficient and more reliable.
Using the ">" pipe operator overwrites the old contents, so no need to delete the file beforehand, and by specifying PIDFile, systemd will the delete the file when stopping the service.

stevenroose:
The pidfile package @captainark suggests looks ok. But make sure to also make it possible to specify the location of the pidfile in the config file.

captainark:
@Marco01809 Thanks, good to know, I have few unit files to edit now ! Since I'm not running gogs as root, I modified what you posted a bit, but it's working great, and it's much cleaner than what I was using : 
```
[...]

PR #3475, opened 2016-08-18
Feature #3120: PID file

0xbaadf00d:
This Pull Request add support for PID file (issue #3120) by adding new settings on `app.ini`:

```
[pid]
; Is Gogs have to create a PID file?
ENABLED  = false
; Location of the PID file to create
PATH     = RUNNING_PID
; Is Gogs can override existing PID file?
OVERRIDE = false
```

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -32,6 +32,7 @@ import (
 	"github.com/gogits/git-module"
 	"github.com/gogits/go-gogs-client"
 
+	"github.com/Unknwon/com"
 	"github.com/gogits/gogs/models"
 	"github.com/gogits/gogs/modules/auth"
 	"github.com/gogits/gogs/modules/bindata"
@@ -79,9 +80,7 @@ func checkVersion() {
 
 	// Check dependency version.
 	checkers := []VerChecker{
-		{"github.com/go-xorm/xorm", func() string {
-			return xorm.Version
-		}, "0.5.5"},
+		{"github.com/go-xorm/xorm", func() string { return xorm.Version }, "0.5.5"},
 		{"github.com/go-macaron/binding", binding.Version, "0.3.2"},
 		{"github.com/go-macaron/cache", cache.Version, "0.1.2"},
 		{"github.com/go-macaron/csrf", csrf.Version, "0.1.0"},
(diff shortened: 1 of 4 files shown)


cmd/web.go
```diff
@@ -79,7 +79,9 @@ func checkVersion() {
 
 	// Check dependency version.
 	checkers := []VerChecker{
-		{"github.com/go-xorm/xorm", func() string { return xorm.Version }, "0.5.5"},
+		{"github.com/go-xorm/xorm", func() string {
+			return xorm.Version
+		}, "0.5.5"},
```
Unknwon: Why change this?

cmd/web.go
```diff
@@ -101,6 +103,31 @@ go get -u %[1]s`, c.ImportPath, c.Version(), c.Expected)
 	}
 }
 
+func createPIDFile() error {
+	if setting.PID.Enabled {
```
Unknwon: You can just return if not enabled.

cmd/web.go
```diff
@@ -101,6 +103,31 @@ go get -u %[1]s`, c.ImportPath, c.Version(), c.Expected)
...
+		_, err := os.Stat(setting.PID.Path)
+		if os.IsNotExist(err) || setting.PID.Override {
+			currentPid := os.Getpid()
+			file, err := os.Create(setting.PID.Path)
+			if err != nil {
+				return fmt.Errorf("Can't create PID file: %v", err)
+			}
+			defer func() {
```
Unknwon: Normally, just `defer file.Close()`.

cmd/web.go
```diff
@@ -101,6 +103,31 @@ go get -u %[1]s`, c.ImportPath, c.Version(), c.Expected)
...
+				if err := file.Close(); err != nil {
+					panic(err)
+				}
+			}()
+			_, err = file.WriteString(fmt.Sprintf("%d\n", currentPid))
```
Unknwon: `com.ToStr()`.

cmd/web.go
0xbaadf00d: My bad, thanks to IntelliJ. I will restore this modification

cmd/web.go
0xbaadf00d: OK

cmd/web.go
0xbaadf00d: I will try

cmd/web.go
0xbaadf00d: ok, thanks for tips

Marco01809:
Seems pretty bad overall. PID files are to be placed in /var/run (or /run on modern systems), so the default should be something like `/var/run/gogs.pid`. The PID file should be deleted on exit. The option not to overwrite the pid file could cause problems, when the gogs daemon crashes and won't restart, because the PID file still exists. Also, gogs won't start when it cannot write the pid file for other reasons.

Doing this via the init manager (systemd, upstart, sysvinit) would be better.

Unknwon:
@Marco01809 do you suggest Gogs should not take care of PID files?

Marco01809:
It wouldn't be wrong to have gogs write a PID file, all other daemons do it as well and it certainly helps when dealing with an older init manager like sysvinit.
It's just that this specific implementation is not very well done, and in the times of systemd, writing pid files isn't absolutely necessary anymore. (I have posted how to do it with systemd in https://github.com/gogits/gogs/issues/3120#issuecomment-242244807)

Unknwon:
@Marco01809 thanks a lot!

PR #3479, opened 2016-08-18
Issues can be closed via API (#3170)

### LINKED ISSUE #3170: FR: API to close already existing issue
Probably via edit, though it can have endpoint of its own.

bkcsoft:
Adds support to change an Issues status via the REST-API. (Fixes #3170)
Depends on gogits/go-gogs-client#41

(Edit: builds failing because of dependency)

### DIFF
--- a/routers/api/v1/repo/issue.go
+++ b/routers/api/v1/repo/issue.go
@@ -161,6 +161,9 @@ func EditIssue(ctx *context.APIContext, form api.EditIssueOption) {
 		ctx.Error(500, "UpdateIssue", err)
 		return
 	}
+	if form.Status != nil {
+		issue.ChangeStatus(ctx.User, ctx.Repo.Repository, strings.Compare("closed", *form.Status) == 0)
+	}
 
 	// Refetch from database to assign some automatic values
 	issue, err = models.GetIssueByID(issue.ID)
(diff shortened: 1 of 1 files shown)


routers/api/v1/repo/issue.go
```diff
@@ -161,6 +161,12 @@ func EditIssue(ctx *context.APIContext, form api.EditIssueOption) {
 		ctx.Error(500, "UpdateIssue", err)
 		return
 	}
+	if form.State != nil {
+		if err = issue.ChangeStatus(ctx.User, ctx.Repo.Repository, strings.Compare("closed", *form.State) == 0); err != nil {
```
Unknwon: I think it should compare with https://github.com/gogits/go-gogs-client/blob/master/issue.go#L16

bkcsoft: Doesn't seem like I can do that...  
`invalid type assertion: gogs.STATE_CLOSED.(string) (non-interface type gogs.StateType on left) (build)at line 165 col 1`  
`cannot use gogs.STATE_CLOSED (type gogs.StateType) as type string in argument to strings.Compare (build)at line 165 col 1`

Unknwon: `gogs.StateType(*form.State)`

bkcsoft: so I shouldn't use `strings.Compare` ?

Unknwon: I think `==` is good..

bkcsoft:
@Unknwon done :)

Unknwon:
Thanks!

Issue #2901, opened 2016-03-28
Encourage usage of certain licenses


gtirloni:
- Gogs version (or commit ref): 0.9.18.0327
- Git version: N/A
- Operating system: N/A
- Database: N/A
- Can you reproduce the bug at http://try.gogs.io: Yes
- Log gist: N/A

## Description

It would help users if Gogs would provide a set of 3-4 licenses at the top of the license drop-down when creating a new repository, just like GitHub does with Apache, GNU and MIT licenses. It would help preventing license fragmentation when few users have the legal background to choose among dozens of licenses (and their different versions).

bkcsoft:
Do you want the ability to set the "defaults" in app.ini?

gtirloni:
@bkcsoft the ability to specify, in `app.ini`, the license file names that should be at the top would be nice. another possibility is to define some convention around how the file names are organized in the `conf/license` dir (e.g. add a prefix or suffix to identify the top ones).

Unknwon:
Patch has pushed to fix this issue, please test on `develop` branch or https://try.gogs.io.

PR #3488, opened 2016-08-19
Support for choosing Preferred Licenses (#2901)

bkcsoft:
Implements #2901.

Adding this to `app.ini` will make these Licenses show at the top of the List in the other they are listed.
```ini
[other]
PREFERRED_LICENSES = MIT,BSD,GPLv2
```

Stuff left to fix:
- [x] Should be in defined order instead of alphabetical
- [x] Add Example to default `app.ini`

More?

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -425,3 +425,7 @@ cs-CZ = cs-CZ
 SHOW_FOOTER_BRANDING = false
 ; Show version information about gogs and go in the footer
 SHOW_FOOTER_VERSION = true
+; Preferred Licenses to place at the top of the List
+; example:
+; PREFERRED_LICENSES = MIT,BSD 3,GPL Lesser
+PREFERRED_LICENSES =
--- a/models/repo.go
+++ b/models/repo.go
@@ -60,6 +60,22 @@ var (
 	ItemsPerPage = 40
 )
 
+// FIXME: This should probably be done in-place instead of creating a new string-slice...
+func sortPreferredLicenses(licenses []string) []string {
+	output := make([]string, 0)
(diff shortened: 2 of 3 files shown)


bkcsoft:
Trying to sort it inline resulted in this cluster bomb, which threw panics... so lets not do that :laughing: 
```go
func sortPreferredLicenses(licenses *[]string) {
	preferred := setting.PreferredLicenses
	log.Info("%d preferred licenses, they are %#v", len(preferred), preferred)
	iter := 0
	for _, p := range preferred {
		found := true
		// run loop until we don't find anything...
		for found {
			found = false
			for i, l := range *licenses {
				if strings.Contains(l, p) {
					temp := append((*licenses)[iter:i], (*licenses)[1+i:]...)
					if iter < 1 {
						*licenses = append([]string{l}, temp...)
					} else {
						*licenses = append((*licenses)[:iter], l)
						*licenses = append((*licenses), temp...)
					}
					iter++
					log.Info("%s was preferred...", l)
					//licenses = append(licenses[:i], licenses[1+i:]...)
					found = true
					break
				}
			}
		}
	}
}
```

bkcsoft:
@Unknwon Not really worth putting any time into making it sort it inline. Thoughts?

Unknwon:
>Not really worth putting any time into making it sort it inline. Thoughts?

How does it work without sorting?

bkcsoft:
@Unknwon Well, now that I think about it, it does exactly what I want it to do... I'm just being stupid and overthinking things :laughing:

Unknwon:
More cleaner implementation is pushed...

PR #3510, opened 2016-08-24
Fix PR-webhook issue where Label-data is stale (#3486)

### LINKED ISSUE #3486: Pull request webhook action 'label_update' does not contain label information
- Gogs version (or commit ref):  0.9.83.0816
- Git version: 1.8.3.1
- Operating system: Linux
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [X] MySQL
  - [ ] SQLite
  - [ ] No
  - [X] Not relevant

The current webhook implementation triggers a 'label_update' action but it seems like it currently does not contain information what was set in the update. It seems that the label information (what labels are set) is available with the next triggered webhook.

bkcsoft:
Fixes #3486 

Label-data is stale, so we have to fetch Issue-data again before sending it...

(Please feel free to suggest a better idea :laughing: )

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -233,9 +233,17 @@ func (i *Issue) HasLabel(labelID int64) bool {
 }
 
 func (issue *Issue) sendLabelUpdatedWebhook(doer *User) {
-	var err error
+	var (
+		new_issue *Issue
+		err       error
+	)
 	if issue.IsPull {
-		issue.PullRequest.Issue = issue
+		new_issue, err = GetIssueByID(issue.ID)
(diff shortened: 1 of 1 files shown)


models/issue.go
```diff
@@ -233,9 +233,17 @@ func (i *Issue) HasLabel(labelID int64) bool {
 }
 
 func (issue *Issue) sendLabelUpdatedWebhook(doer *User) {
-	var err error
+	var (
+		new_issue *Issue
```
Unknwon: Bad naming style.

PR #3513, opened 2016-08-25
added support to set pull_request event from api

kmadel:
The pull_request event for webhooks was not supported when creating/updating webhooks via the API, this is a fix for that.


Unknwon:
Thanks!

PR #3534, opened 2016-08-30
Update .editorconfig

andreynering:
The primary motivation is documenting `.less`, `.js` and `.tmpl` extensions, conform existing conventions.


Unknwon:
Thanks!

PR #3537, opened 2016-08-30
Adds support for gogs-client to retrieve repos by username and org

kmadel:
This is related to gogs/gogs-client issue 37

Requires gogs/gogs-client PR 45

Provides the ability to retrieve repos by username or by org via the REST API (gogs-client).

Note: 
- /users/:user/repos will only return public repos - same as https://developer.github.com/v3/repos/#list-user-repositories; api.Permission defaults to false, false, true
- /orgs/:org/repos - returns private if ctx.user matches org owner, and public; api.Permission determined by ctx.user ownership
- still may need some work around repo api.Permission


routers/api/v1/repo/repo.go
```diff
@@ -109,6 +118,63 @@ func ListMyRepos(ctx *context.APIContext) {
...
+			ctx.Status(404)
+		} else {
+			ctx.Error(500, "ListUserRepos", err)
+		}
+		return
+	}
+
+	ownRepos, err := listUserRepos(ctx, u, false)
+	if err != nil {
+		ctx.Error(500, "GetUserRepositories", err)
```
Unknwon: All error desc string do not look right..

PR #3549, opened 2016-08-30
Add all, check and dist Makefile rules

strk:
These are to follow the GNU Coding Standards Makefile Targets:
https://www.gnu.org/prep/standards/html_node/Standard-Targets.html


Unknwon:
Thanks!

Issue #3557, opened 2016-08-31
Init script for (open)SuSE distributions with sysv init


Corben78:
Feature request:
Please add a (open)SuSE init script using Sys V init.

Unknwon:
Closed by https://github.com/gogits/gogs/pull/3558

PR #3558, opened 2016-08-31
Add a (Open)SuSE init script

Corben78:
Referring to #3557 here is a working init script for (open)SuSE distributions with Sys V init.


Unknwon:
Thanks!

Issue #3511, opened 2016-08-24
Pull Request in dashboard


hgaiser:
- Gogs version (or commit ref): 0.9.71
- Git version: 2.9.3
- Operating system: Arch
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [x] No
  - [ ] Not relevant
- Log gist:

Adding Pull Requests or assigning Pull Requests to people does not make them appear in the Dashboard (under Pull Requests). The Pull Requests section under Dashboard simply shows 0 active Pull Requests. Is there anyway I can debug what is going wrong here? I tried it on try.gogs.io and it seems to work there.

Unknwon:
Hi, this is basically same to https://github.com/gogits/gogs/issues/1820, you can comment on that thread.

Issue #3132, opened 2016-05-25
My issues don't show up when repo belongs to organisation


DeKugelschieber:
- Gogs version (or commit ref): 0.9.13.0318
- Operating system: Ubuntu 14.04 LTS (Server)
- Database:
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at http://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not tried yet

## Description

Issues assigned to me won't show up below the "Issues" section when the repo the issue is created for belongs to an organisation. Also on the left: "Assigned to you" shows 0.

I'll may try it on try.gogs.io later.

edacval:
#1820 ?

DeKugelschieber:
Yes, sound familiar :)

Issue #3052, opened 2016-05-03
gogs issues can't show in right place


niluanxy:
gogs issues can't show in right place,for example,i create a tema that's named <strong>temaA</strong>,and my user is aaa,now i should have two user, one is <strong>aaa</strong>, one is <strong>teamA</strong>.Then i use <strong>temaA</strong> to create a project that's named <strong>work1</strong>.After i create issues on <strong>work1</strong>,but it can't tip my user <strong>temaA</strong> or <strong>aaa</strong> to soluve this issues,even i have set to user <strong>aaa</strong>.

<img width="1174" alt="qq20160503-2 2x" src="https://cloud.githubusercontent.com/assets/7076522/14977249/074d1ef4-1145-11e6-97b1-f18168c00da1.png">

<img width="1188" alt="qq20160503-3 2x" src="https://cloud.githubusercontent.com/assets/7076522/14977319/781051e2-1145-11e6-959b-6ca988e577b2.png">

Zeeker:
If I understand you correctly, this is a duplicate of #1820 and will be realised in 0.10.0.

Unknwon:
@Zeeker has posted what needs to be said...

Issue #2865, opened 2016-03-21
No issues showing up on Issues tab


johnnylee:
- Gogs version (or commit ref): 0.9.13.0318
- Git version: 2.1.4
- Operating system: Debian Jessie
- Database:
  - [X] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
- Can you reproduce the bug at http://try.gogs.io:
  - [ ] Yes
  - [X] No
  - [ ] Not relevant
- Log gist:

## Description

I don't see any issues in the `Issues` tab, even though I have an assigned issue, and an issue in one of my repositories, and I have created issues. 

Looking at the log files, the problem appears to be due to the query here:

```sql
[xorm] [info]  2016/03/21 08:58:48.617567 [sql] SELECT COUNT(*) FROM "issue"  WHERE issue.is_closed=$1 AND issue.repo_id=0 AND issue.is_pull=0 AND assignee_id=$2 [args] [false 1]
```

Specifically `issue.repo_id=0` doesn't make sense.

TilmannBach:
I think its another duplicate to #1820 ?

johnnylee:
It looks like it.

PR #3560, opened 2016-08-31
Fix assigned/created issues in dashboard.

hgaiser:
This pull requests adds functionality to the Issues and Pull Requests view in the Dashboard. Previously these pages would only show Issues / Pull Requests placed in repositories **owned** by the user. This pull request changes the functionality of the "Assigned to you" and "Created by you" filters. These filters now show Issues / Pull Requests you are assigned to or you created from **any** repository. Issues related to this pull request are:

~~https://github.com/gogits/gogs/issues/3511~~
https://github.com/gogits/gogs/issues/1820
~~https://github.com/gogits/gogs/issues/2202~~
https://github.com/gogits/gogs/issues/1580
~~https://github.com/gogits/gogs/issues/2518~~
~~https://github.com/gogits/gogs/issues/2865~~
~~https://github.com/gogits/gogs/issues/3052~~
~~https://github.com/gogits/gogs/issues/3132~~

Feedback on this Pull Request is appreciated, as I understand it there are strict guidelines ;)

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -1062,7 +1062,7 @@ func UpdateIssueMentions(issueID int64, mentions []string) error {
 // IssueStats represents issue statistic information.
 type IssueStats struct {
 	OpenCount, ClosedCount int64
-	AllCount               int64
+	YourRepositoriesCount  int64
 	AssignCount            int64
 	CreateCount            int64
 	MentionCount           int64
@@ -1070,7 +1070,7 @@ type IssueStats struct {
 
 // Filter modes.
 const (
-	FM_ALL = iota
+	FM_YOUR_REPOSITORIES = iota
 	FM_ASSIGN
 	FM_CREATE
 	FM_MENTION
@@ -1122,7 +1122,7 @@ func GetIssueStats(opts *IssueStatsOptions) *IssueStats {
 	}
 
 	switch opts.FilterMode {
-	case FM_ALL, FM_ASSIGN:
+	case FM_YOUR_REPOSITORIES, FM_ASSIGN:
 		stats.OpenCount, _ = countSession(opts).
 			And("is_closed = ?", false).
 			Count(&Issue{})
@@ -1165,9 +1165,9 @@ func GetUserIssueStats(repoID, uid int64, repoIDs []int64, filterMode int, isPul
 	countSession := func(isClosed, isPull bool, repoID int64, repoIDs []int64) *xorm.Session {
 		sess := x.Where("issue.is_closed = ?", isClosed).And("issue.is_pull = ?", isPull)
 
-		if repoID > 0 || len(repoIDs) == 0 {
+		if repoID > 0 {
 			sess.And("repo_id = ?", repoID)
-		} else {
+		} else if repoIDs != nil {
 			sess.In("repo_id", repoIDs)
 		}
 
--- a/models/issue.go
+++ b/models/issue.go
@@ -1174,28 +1174,38 @@ func GetUserIssueStats(repoID, uid int64, repoIDs []int64, filterMode int, isPul
 		return sess
 	}
 
-	stats.AssignCount, _ = countSession(false, isPull, repoID, repoIDs).
+	stats.AssignCount, _ = countSession(false, isPull, repoID, nil).
 		And("assignee_id = ?", uid).
 		Count(&Issue{})
 
-	stats.CreateCount, _ = countSession(false, isPull, repoID, repoIDs).
+	stats.CreateCount, _ = countSession(false, isPull, repoID, nil).
 		And("poster_id = ?", uid).
 		Count(&Issue{})
 
-	openCountSession := countSession(false, isPull, repoID, repoIDs)
-	closedCountSession := countSession(true, isPull, repoID, repoIDs)
+	stats.YourRepositoriesCount, _ = countSession(false, isPull, repoID, repoIDs).
+		Count(&Issue{})
 
 	switch filterMode {
+	case FM_YOUR_REPOSITORIES:
+		stats.OpenCount, _   = countSession(false, isPull, repoID, repoIDs).
+		Count(&Issue{})
+		stats.ClosedCount, _ = countSession(true, isPull, repoID, repoIDs).
+		Count(&Issue{})
 	case FM_ASSIGN:
-		openCountSession.And("assignee_id = ?", uid)
-		closedCountSession.And("assignee_id = ?", uid)
+		stats.OpenCount, _   = countSession(false, isPull, repoID, nil).
+		And("assignee_id = ?", uid).
+		Count(&Issue{})
+		stats.ClosedCount, _ = countSession(true, isPull, repoID, nil).
+		And("assignee_id = ?", uid).
+		Count(&Issue{})
 	case FM_CREATE:
-		openCountSession.And("poster_id = ?", uid)
-		closedCountSession.And("poster_id = ?", uid)
+		stats.OpenCount, _   = countSession(false, isPull, repoID, nil).
+		And("poster_id = ?", uid).
+		Count(&Issue{})
+		stats.ClosedCount, _ = countSession(true, isPull, repoID, nil).
+		And("poster_id = ?", uid).
+		Count(&Issue{})
 	}
-
-	stats.OpenCount, _ = openCountSession.Count(&Issue{})
-	stats.ClosedCount, _ = closedCountSession.Count(&Issue{})
 
 	return stats
 }
--- a/models/issue.go
+++ b/models/issue.go
@@ -1125,35 +1125,35 @@ func GetIssueStats(opts *IssueStatsOptions) *IssueStats {
 	case FM_YOUR_REPOSITORIES, FM_ASSIGN:
 		stats.OpenCount, _ = countSession(opts).
 			And("is_closed = ?", false).
-			Count(&Issue{})
+			Count(new(Issue))
 
 		stats.ClosedCount, _ = countSession(opts).
 			And("is_closed = ?", true).
-			Count(&Issue{})
+			Count(new(Issue))
 	case FM_CREATE:
 		stats.OpenCount, _ = countSession(opts).
 			And("poster_id = ?", opts.UserID).
 			And("is_closed = ?", false).
-			Count(&Issue{})
+			Count(new(Issue))
 
 		stats.ClosedCount, _ = countSession(opts).
 			And("poster_id = ?", opts.UserID).
 			And("is_closed = ?", true).
-			Count(&Issue{})
+			Count(new(Issue))
 	case FM_MENTION:
 		stats.OpenCount, _ = countSession(opts).
 			Join("INNER", "issue_user", "issue.id = issue_user.issue_id").
 			And("issue_user.uid = ?", opts.UserID).
 			And("issue_user.is_mentioned = ?", true).
 			And("issue.is_closed = ?", false).
-			Count(&Issue{})
+			Count(new(Issue))
 
 		stats.ClosedCount, _ = countSession(opts).
 			Join("INNER", "issue_user", "issue.id = issue_user.issue_id").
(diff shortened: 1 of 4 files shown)


routers/user/home.go
```diff
@@ -215,63 +219,48 @@ func Issues(ctx *context.Context) {
...
 
-	allCount := 0
-	repoIDs := make([]int64, 0, len(repos))
-	showRepos := make([]*models.Repository, 0, len(repos))
 	for _, repo := range repos {
 		if (isPullList && repo.NumPulls == 0) ||
-			(!isPullList &&
-				(!repo.EnableIssues || repo.EnableExternalTracker || repo.NumIssues == 0)) {
+		(!isPullList &&
+		(!repo.EnableIssues || repo.EnableExternalTracker || repo.NumIssues == 0)) {
```
Unknwon: Does not seem like it's the result of `go fmt`.

routers/user/home.go
```diff
@@ -282,26 +271,87 @@ func Issues(ctx *context.Context) {
...
-	ctx.Data["ViewType"] = viewType
-	ctx.Data["SortType"] = sortType
-	ctx.Data["RepoID"] = repoID
+	var issueStats *models.IssueStats
+	if filterMode == models.FM_YOUR_REPOSITORIES {
+		issueStats = models.GetUserIssueStats(repoID, ctxUser.ID, userRepoIDs, filterMode, isPullList)
+	} else {
+		issueStats = models.GetUserIssueStats(repoID, ctxUser.ID, nil, filterMode, isPullList)
+	}
+	issueStats.YourRepositoriesCount = int64(len(yourIssues))
```
Unknwon: This does not make sense to me, the count is supposes to be the number of all issues in given type.

hgaiser: Meaning both open and closed issues?

I indeed agree that what is there now is not logical. I'm wondering if it should always be the number of **open** issues / pull requests or the **total** number of issues / pull requests. What do you think?

Unknwon: It should be **total number of open issues**.

routers/user/home.go
hgaiser: I'm sorry, I don't quite follow. What do you mean by `go fmt`? The `go` formatting package?

routers/user/home.go
hgaiser: Right, but then filtered on for example assignee in the case of "Assigned to you", so total number of open issues assigned to you at that moment.

routers/user/home.go
Unknwon: Ref: https://blog.golang.org/go-fmt-your-code

routers/user/home.go
Unknwon: `int64(len(yourIssues))` this is not total number, it just number of returned results.

routers/user/home.go
hgaiser: Ah, thank you. I will format the code using that tool tomorrow.

hgaiser:
I used `gofmt` as requested to format the file. I also modified the code to base the Dashboard stats purely on `GetUserIssueStats`, which makes the code in `Issues` more clean.

routers/user/home.go
```diff
@@ -215,93 +219,121 @@ func Issues(ctx *context.Context) {
...
 		return
 	}
 
-	// Get posters and repository.
-	for i := range issues {
-		issues[i].Repo, err = models.GetRepositoryByID(issues[i].RepoID)
+	showRepos := make([]*models.Repository, 0, len(issues))
+	showReposSet := make(map[int64]bool)
+
+	if repoID > 0 {
```
hgaiser: Does this part need some kind of check to see if the user has access to this repository? Can the user otherwise fill in random repoID's in the URL and see the names of private repositories? Please provide feedback on this part.

models/issue.go
```diff
@@ -1062,15 +1062,15 @@ func UpdateIssueMentions(issueID int64, mentions []string) error {
 // IssueStats represents issue statistic information.
 type IssueStats struct {
 	OpenCount, ClosedCount int64
-	AllCount               int64
+	YourRepositoriesCount  int64
```
Unknwon: I think you did not answer my question, what is the purpose of changing this?

hgaiser: It wasn't clear to me that you were asking that. I renamed AllCount to YourRepositoriesCount because AllCount sounds like it is the count of **all** issues open _in which you are participating_, **not** the count of open issues _in your repositories_ (which is what is actually displayed).

Before this Pull Request, the name "All" makes sense, since the "Assigned" and "Created" views show filtered versions of "All". After this Pull Request "Assigned" and "Created" views show different information compared to the "Your repositories" view (namely that it can also show issues from repositories not owned by you).

Unknwon: Ok Sounds good.

hgaiser: As a side note, personally I'd think the most interesting views are "Created by you" "Assigned to you" and "Mentions you" (in other words, no "Your repositories" view and an added "Mentions you" view). This is perhaps a discussion for another time, for this PR I left this functionality unchanged.

models/issue.go
```diff
@@ -1122,7 +1122,7 @@ func GetIssueStats(opts *IssueStatsOptions) *IssueStats {
 	}
 
 	switch opts.FilterMode {
-	case FM_ALL, FM_ASSIGN:
+	case FM_YOUR_REPOSITORIES, FM_ASSIGN:
 		stats.OpenCount, _ = countSession(opts).
 			And("is_closed = ?", false).
 			Count(&Issue{})
```
Unknwon: It would be better if you could include following change in this PR:

`&Issue{}` => `new(Issue)`.

hgaiser: I read up a bit on this distinction, do not fully understand it yet but I will do as you request ;).

From what I understand, &T{} only works for certain types (structs, slices, etc.) whereas new(T) works for any type. Is that the reason you prefer the latter?

routers/user/home.go
Unknwon: Yes we should, see below.

routers/user/home.go
```diff
@@ -215,93 +219,121 @@ func Issues(ctx *context.Context) {
...
+		repo, err := models.GetRepositoryByID(repoID)
 		if err != nil {
-			ctx.Handle(500, "GetRepositoryByID", fmt.Errorf("[#%d]%v", issues[i].ID, err))
+			ctx.Handle(500, "GetRepositoryByID", fmt.Errorf("[#%d]%v", repoID, err))
 			return
 		}
```
Unknwon: Here we can check access by:

```go
if !repo.IsOwnedBy() && !repo.HasAccess() {
    // 404
    return
}

...
```

routers/user/home.go
```diff
@@ -215,93 +219,121 @@ func Issues(ctx *context.Context) {
...
+			ctx.Handle(500, "GetOwner", fmt.Errorf("[#%d]%v", repoID, err))
 			return
 		}
+
+		showReposSet[repoID] = true
+		showRepos = append(showRepos, repo)
 	}
-	ctx.Data["Issues"] = issues
 
+	showIssues := make([]*models.Issue, 0, len(issues))
```
Unknwon: This block of code is not right.

We must control returned issues when we call`models.Issues`, not filter after that. Otherwise pagination is broken.

models/issue.go
Unknwon: No.. just personal taste, if does not initialize any field, use `new()`.

routers/user/home.go
hgaiser: Shouldn't this be 500 error?

Unknwon: it is not server error.

hgaiser: 403 ? Forbidden access.

This way people can still figure out "valid" repoID's based on the error returned, not sure how much of an issue that is.

routers/user/home.go
hgaiser: Actually I believe this block can be removed now. Previously I had filtered here indeed, but now this filtering is done in the retrieving of the issues.

hgaiser:
Updated with latest remarks.

routers/user/home.go
```diff
@@ -285,18 +285,18 @@ func Issues(ctx *context.Context) {
 			return
 		}
 
+		// Check if user has access to given repository.
+		if !repo.IsOwnedBy(ctxUser.ID) && !repo.HasAccess(ctxUser) {
+			ctx.Handle(403, "Issues", fmt.Errorf("#%d", repoID))
```
Unknwon: Gogs does not expose repository as 403 if user does not have access, so we need to use `404`.

routers/user/home.go
Unknwon: Good, then we don't even need to create `showIssues`.

routers/user/home.go
hgaiser: `404` it is.

routers/user/home.go
hgaiser: Yes exactly, it has been removed as well already.

routers/user/home.go
```diff
@@ -215,93 +219,119 @@ func Issues(ctx *context.Context) {
...
+			}
+		}
 	}
-	ctx.Data["Issues"] = issues
 
+	issueStats := models.GetUserIssueStats(repoID, ctxUser.ID, userRepoIDs, filterMode, isPullList)
+
+	ctx.Data["Issues"] = issues
+	ctx.Data["Repos"] = showRepos
+	ctx.Data["Page"] = paginater.New(len(issues), setting.UI.IssuePagingNum, page, 5)
```
Unknwon: `len(issues)` is not total.

hgaiser: Right, because we only select x issues in the query. I will adjust this.

hgaiser:
Any idea why the Travis CI build failed? It appears it failed on downloading the code?

Unknwon:
Looks like randomly fail on fetching code...

hgaiser:
Is there anything left to change? Do you want me to squash the commits into one commit?

Unknwon:
Nope, I'll squash while merging.

hgaiser:
May I ask why this is not merged? Is there anything I can do or have you not had the time?

acornejo:
Is this abandoned? When will it be merged.

hgaiser:
@Unknwon will you merge this ?

gsantner:
Maybe contact maintainers (eg @bkcsoft ) as Unknwon seems quite inactive recently?

0xbaadf00d:
unfortunately @gsantner, according to another thread, it seems that only @Unknwon have permission to merge PR

gsantner:
@0xbaadf00d  Okay, hmm. Thats somewhat really bad, such a big projet controlled by just one person (No time, no interest, dead, .. anything can happen..).  But okay, this is offtopic for this PR.

jeppech:
It might be worth mentioning, that due to Unknwons absence, the [Gitea](https://github.com/go-gitea/gitea) fork is coming alive again, with several maintainers, instead of one.

gsantner:
@jeppech Yeah, but who wants to use it, if it won't get updated for a long time anymore if gogs get commits on master again? Nobody^^

de-vri-es:
> @jeppech Yeah, but who wants to use it, if it won't get updated for a long time anymore after gogs get commits on master again? Nobody^^

Unless this is a recurring problem. And the fork seems to keep up-to-date with this repository. So it seems interesting enough.

jeppech:
@gsantner @de-vri-es, I know this thread, is not meant for this discussion. But he's been AWOL for quite some time now, and this has happend before. Furthermore, he have already stated that, he will be the only maintainer of this project. Ie. it could happen again..

hgaiser:
Reopend PR to trigger Travic CI build. Too bad there isn't enough maintenance on this project, it has potential.. Perhaps that fork is something interesting if it is properly maintained.

gsantner:
@hgaiser You could send a PR to <https://github.com/go-gitea/gitea> also

Unknwon:
Thanks!

Issue #3480, opened 2016-08-18
Adding SSH key with newline yields invalid key format


Daxda:
- Gogs version (or commit ref): 0.9.83.0816
- Git version: 2.9.3
- Operating system: CentOS
- Database:
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [X] Not relevant
## Description

Upon adding a new public SSH key by issuing

``` bash
cat ~/.ssh/id_rsa.pub | xclip -selection c
```

and then inserting it on the Gogs' settings tab it produces the error `Sorry, we're not able to verify your SSH key: extractTypeFromBase64Key: invalid key format:`

I guess that's because the terminating null byte is also copied by my command above.
Would it be possible to fix this _issue_ by a simple replace of the null byte?

gramakri:
Ah, thanks for opening this. I hit exactly this in the cloudron.io app and was talking about this on gitter.

gramakri:
This worked in 0.9.48 and stopped working from 0.9.60

Daxda:
Okay I was wrong about the null byte, it seems to be a plain `\n`

bkcsoft:
Remove `ssh-<crypto>` and `<user>@<host>` and it works. I'll see if I can make a patch for this

Daxda:
It also just works if I remove the trailing newline :)

bkcsoft:
ooh... that's even easier :laughing:

W-quan:
Use gedit to open it and copy. It work.

strk:
I was just struck by this issue, sounds like an easy fix indeed

Unknwon:
Should be fixed by https://github.com/gogits/gogs/pull/3565.

PR #3565, opened 2016-09-01
#3480 Fix new ssh key adding issues

mohamme89d:
https://github.com/gogits/gogs/issues/3480
Added regular exp match (108) and remove training new line (111) that cause system to think its an ssh2 key.

tested with Windows OpenSSH and OpenSSH2 keys

### DIFF
--- a/models/ssh_key.go
+++ b/models/ssh_key.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -104,8 +105,10 @@ func extractTypeFromBase64Key(key string) (string, error) {
 
 // parseKeyString parses any key string in OpenSSH or SSH2 format to clean OpenSSH string (RFC4253).
 func parseKeyString(content string) (string, error) {
-	// Transform all legal line endings to a single "\n".
-	content = strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(content)
+	// Transform all legal line endings to a single "\n"
+	newlineExpression := regexp.MustCompile(`\r?\n`) // windows style optional with (\r?)
+	content = newlineExpression.ReplaceAllString(content, "\n")
+	content = strings.TrimRight(content, "\n")
 	lines := strings.Split(content, "\n")
 
 	var keyType, keyContent, keyComment string
--- a/models/ssh_key.go
+++ b/models/ssh_key.go
@@ -14,7 +14,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


Unknwon:
I think what you did with _regular exp match_ is overkill, all we need is `content = strings.TrimRight(content, "\n")`..

mohamme89d:
I did try few tools and sometimes the comments part come in a new line so this will cater of few edge cases.

models/ssh_key.go
```diff
@@ -104,8 +105,10 @@ func extractTypeFromBase64Key(key string) (string, error) {
 
 // parseKeyString parses any key string in OpenSSH or SSH2 format to clean OpenSSH string (RFC4253).
 func parseKeyString(content string) (string, error) {
-	// Transform all legal line endings to a single "\n".
-	content = strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(content)
+	// Transform all legal line endings to a single "\n"
+	newlineExpression := regexp.MustCompile(`\r?\n`) // windows style optional with (\r?)
+	content = newlineExpression.ReplaceAllString(content, "\n")
```
bkcsoft: Why not just `content = strings.Replace(content, "\r\n", "\n")` ?

Unknwon:
>I did try few tools and sometimes the comments part come in a new line so this will cater of few edge cases.

Any examples?

models/ssh_key.go
mohamme89d: > content = strings.Replace(content, "\r\n", "\n")  
will only fix windows new lines

bkcsoft: well, `\r?\n` will only match `\r\n` and `\n`, not `\r` like the previous implementation... I think you're looking for `\r\n?` in that case

mohamme89d:
Did try tools below to copy the ssh-key public content 

1. Git kraken 
2. Windows Notepad
3. Notepad++ 

all of then append a new line at the end (which i think this is how ssh-keygen getenrate ssh-key [ssh-keygen from git 2.9.x])
capture


Case 1 : 
On Windows we when the old expression ran it replaces '\r\n' to '\n\n' which is two new line and it throws the whole thing out of desired behavior 
`old replacer uses ("\r\n", "\n", "\r", "\n"), but not sure why go is actually making "\r" more prioretized than "\r\n"
by right it should find "\r\n" and first but it seems like it goes for "\r" first so we end up with ("\r\n" --> "\n\n")
`
fix for that was 
1- define a regular expression that matches both environemts 

> regexp.MustCompile(`\r?\n`) // windows style optional with (\r?) 

2- match against the regular expression and replace matches 

>   content = newlineExpression.ReplaceAllString(content, "\n") 

Case 2 :
On single line key will have an extra (Windows '\r\n' or Linux '\n') at the end to fix that we have 
> strings.TrimRight(content, "\n")
Which just remove any additional new line at the end which is useless in all cases

Unknwon:
>Case 1 : 
On Windows we when the old expression ran it replaces '\r\n' to '\n\n' which is two new line and it throws the whole thing out of desired behavior 
old replacer uses ("\r\n", "\n", "\r", "\n"), but n

We can use two `strings.Replace()` to fix this.

mohamme89d:
The key is not small (looking at scale) using two replaces will scan the key content twice and generate 2 new immutable strings, but if regular expression is used we scan it once then matches are replaced generation only 1 new immutable string.

models/ssh_key.go
mohamme89d: That is ture, but we should not get \r only, if that happens it means the input end is not a user with an interface ... it is a script.

Unknwon:
Key is not big enough to have such concern, I prefer more readable code.

mohamme89d:
Cool, will apply the changes and push.

KokaKiwi:
Hi, why not merging the PR?

Unknwon:
Thanks!

Issue #3578, opened 2016-09-01
LDAP first login fails since 0.9.97


strk:
Upgrading from 0.9.68 to 0.9.97 left us with a blank page resulting from login attempts by new LDAP users.
Reverting to 0.9.68 allowed people to perform their first login.

The database was just not being updated, with 0.9.97, so the new user was not being created locally.
No traces of errors in the logs, and just a blank page.
This cannot be tested on try.gogs.io for obvious reasons (no LDAP there).

It's not easy to test because existing users could login just fine, so you need to either create a new LDAP user or locally delete an existing LDAP user, and try to login again.

Unknwon:
Fixed already in `master` https://github.com/gogits/gogs/commit/7c1fbed05764ff27c0684df92961a030e2e3344b

strk:
Will test ASAP. Meanwhile a question: should I see any error in the logs ? Because I didn't see any...

Unknwon:
I think it is crashed, panic does not go into log file, only to terminal.

strk:
Uhm, so maybe scripts/init/debian/gogs should not redirect stdout to
/dev/null as it does, but rather direct it to another logfile

strk:
Fails again as of commit 491407ddf89fb4182c5401c47d2af38f0a30adca (0.9.99.0903) -- please reopen

strk:
Tested again, works as 491407ddf89fb4182c5401c47d2af38f0a30adca -- don't bother opening

herbat73:
works fine with OpenLDAP with the latest 0.9.99.0915 sources 
binary version is a bit outdated and it does not work properly with LDAP auth

PR #3579, opened 2016-09-01
Redirect stdout and stderr to logfile

strk:
Makes it harder to loose panic messages (see #3578)

### DIFF
--- a/modules/log/file.go
+++ b/modules/log/file.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


modules/log/file.go
```diff
@@ -113,6 +114,10 @@ func (w *FileLogWriter) StartLogger() error {
 	if err = w.initFd(); err != nil {
 		return err
 	}
+	// redirect stdout and stderr to file
+	// TODO: allow specifying in configuration ?
+	syscall.Dup2(int(fd.Fd()), 1) // stdout
+	syscall.Dup2(int(fd.Fd()), 2) // stderr
```
Unknwon: This won't compile on Windows.

Unknwon:
Close and merge to https://github.com/gogits/gogs/issues/3893.

Issue #3532, opened 2016-08-29
Convert an user into an organization


adaur:
Hello,

Is there any (easy) way to convert an user into an organization? I've created the first user with the name of my company and created a first repo. I realized my mistake later, once the repo and its hooks were set.

I did not find any documentation on this, hence my ticket.

Thanks!

Unknwon:
Hi, it is not possible currently..

DblK:
I propose to begin with the implementation of the menu "Organization" under settings to display the current user organization and to add new organization as the screenshot :

![image](https://cloud.githubusercontent.com/assets/832617/18146860/3af8fdb8-6fd2-11e6-82d5-04581d679e73.png)

In a second PR I will add the button to turn user into organization.
@Unknwon on Are you okay with that?

Unknwon:
@DblK yes, sounds good!

kryptus36:
It looks as though #3587 fixed one part of this but not the conversion part. I've done the same as @adaur 

@DblK did you ever get to finishing the feature?

PR #3587, opened 2016-09-02
Setting orgs

DblK:
This is the first step for resolving #3532 by implements a new menu "Organizations" under User.Settings.
It display all the organizations that the current user is member and allow also to create a new organization.

image

image

image

image

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -246,8 +246,7 @@ func runWeb(ctx *cli.Context) error {
 		m.Combo("/applications").Get(user.SettingsApplications).
 			Post(bindIgnErr(auth.NewAccessTokenForm{}), user.SettingsApplicationsPost)
 		m.Post("/applications/delete", user.SettingsDeleteApplication)
-		m.Combo("/organizations").Get(user.SettingsOrganizations).
-			Post(bindIgnErr(auth.CreateOrgForm{}), user.SettingsOrganizationsPost)
+		m.Get("/organizations", user.SettingsOrganizations)
 		m.Route("/delete", "GET,POST", user.SettingsDelete)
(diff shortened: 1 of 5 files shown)


Unknwon:
Thanks, but

1. We don't need create organization UI inside this page, instead redirect to the proper page.
2. I prefer the way GitHub displays organizations..

image

DblK:
Hi. Thanks for your reply.
I'm not part on any organization on Github so I didn't know how to display it.
I will modify the template according to the screenshot.

I understand the fact to redirect to the page of the new organization button.
How should it act for the leave button?

Unknwon:
>How should it act for the leave button?

You can leave this part to me, it's a bit hard to explain..

DblK:
Okay, I will make the change and commit them when finished.
My question about the leave button was about:
- Should it be a link to redirect to the correct page
or
- an implementation in this page like I did for "New Organization" aka handling leaving organization within this page?

After your implementation and the merge of the PR, I will continue on the "Transform %s into an organization" feature. Before developing, I will post some idea of implementation to be sure to go in the wanted way.

Unknwon:
@DblK 

>an implementation in this page like I did for "New Organization" aka handling leaving organization within this page?

similar, but UI work is hard to explain..

DblK:
Here is the new UI:
image

and when there are at least 2 orgs:
image

Now, when you click on "New Organization", it will redirect to the create organization page.

Unknwon:
Thanks, will review ASAP.

appleboy:
@DblK Could you mind port this feature to `Gitea`?

DblK:
@appleboy [done ;)](https://github.com/go-gitea/gitea/pull/1739)

Issue #3590, opened 2016-09-03
404 on release draft edition


0xbaadf00d:
- Gogs version (or commit ref): 0.9.98.0901
- Git version: 2.9.3
- Operating system: CentOS
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (https://try.gogs.io/zero-x-baadf00d/test/releases => https://try.gogs.io/zero-x-baadf00d/test/releases/edit/test)
  - [ ] No
  - [ ] Not relevant
- Log gist:

```
2016/09/03 07:36:39 [...ules/context/repo.go:432 func1()] [E] RepoRef invalid repo: branch or tag not exist: r/1
```

## Description

We can't edit / publish a release draft.

## How to reproduce

1. Go to releases tab
2. Click on 'New Release'
3. Choose a 'Tag name' and fill other fields
4. Click on 'Save draft'
5. On the releases tab, click 'Edit' for the newly created draft

0xbaadf00d:
I do a try by removing `context.RepoRef()` from `web.go`.

```
m.Group("/releases", func() {
			m.Get("/new", repo.NewRelease)
			m.Post("/new", bindIgnErr(auth.NewReleaseForm{}), repo.NewReleasePost)
			m.Get("/edit/*", repo.EditRelease)
			m.Post("/edit/*", bindIgnErr(auth.EditReleaseForm{}), repo.EditReleasePost)
			m.Post("/delete", repo.DeleteRelease)
		}, reqRepoWriter)
```

and the issue seams fixed. The _/releases/xxxxxx_  don't use repository reference and don't handle it on the different URLs  _/releases_  _/releases/<tagname>/edit_ ...  

Or maybe add a new `else if` on `context.RepoRef()` to add an exception for releases.

Unknwon:
Let me guess... the commits count is now missing from the repository header bar?

0xbaadf00d:
Hmm you right. The commit count is missing :-(

I try with `context.RepoRef()` but by commenting the last line 

```
ctx.Handle(404, "RepoRef Invalid repo", fmt.Errorf("branch or tag not exist: %s", refName))
```

Page is OK, but commit count always missing.

Maybe, in fact, the issue is Gogs try to request Git repository for something (Release)  that it must be only database requested ?

lstahlman:
[...]

PR #3591, opened 2016-09-03
Fix for 3590: 404 on editing a draft release

lstahlman:
This PR addresses #3590 by loading default repository references when a editing or deleting a release that is in draft state.


lstahlman:
DefaultRepoRefOnDraftRelease might be a bit too verbose, so suggestions for a better name are welcome.

Unknwon:
Thanks, but pushed more minimal fix.

lstahlman:
No problem.  The only drawback I see to 491407d is that it shows the wrong commit count on the page when editing a published release.

PR #3595, opened 2016-09-03
[WIP] Openid support

strk:
this is really just a stub, adding here in case others want to help

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -99,7 +99,7 @@ func checkVersion() {
 	for _, c := range checkers {
 		if !version.Compare(c.Version(), c.Expected, ">=") {
 			log.Fatal(4, `Dependency outdated!
-Package '%s' current version (%s) is below requirement (%s), 
+Package '%s' current version (%s) is below requirement (%s),
 please use following command to update this package and recompile Gogs:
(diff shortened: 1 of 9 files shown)


rawlingsj:
This would be a great feature.  I'm not sure I'm in a position to help but I can try, perhaps with a bit of testing when the time comes.  We use gogs on the [fabric8](https://fabric8.io/) project which runs on Kubernetes.  We're looking at authentication using OpenID with Keycloak so if any of that is interesting let me know and I can try to help.

strk:
At this stage the help needed is with macaron and interaction design.
The code in the branch should let you add an "OpenID" authentication
method and then would take a Username containing "://" as an OpenID
url, and try to verify you against that (but won't create a local user
yet, just print - in the logs - whether or not you're verified).

strk:
Oh, you also need to enter a password but is not used, just type
in a random character (that's where I need help with Macaron)

Unknwon:
Is this still in development?

strk:
Yes but the new development is based on Gitea. Feel free to close
this one unless you find something useful.

Unknwon:
Thanks for clarifying!

Issue #1122, opened 2015-03-23
log.smtp receiver configure error


lostsnow:
when I configure log.smtp

``` ini
[log.smtp]
#...
RECEIVERS = ["a@mail.com"]  #or ["a@mail.com","b@mail.com"]
```

got error

```
2015/03/23 17:26:05 [I] Gogs 0.6.0.0319 Beta
2015/03/23 17:26:05 [I] Log Mode: File(Info)
2015/03/23 17:26:05 [log.go:294 Error()] [E] Fail to set logger(smtp): invalid character 'a' after object key:value pair
r 'a' after object key:value pair
```

Unknwon:
Thanks for your report!

raphael10241024:
Is there any special meaning of adding bracekt here?
It seems that removing bracket can fix it .

PR #3601, opened 2016-09-05
fix bug #1122 log.smtp config

raphael10241024:
fix bug #1122 log.smtp receiver configure error
INI file doesn't support [] in fact ,  we usually use ',' to join array .


raphael10241024:
I found I made a mistake , just use string.replace() is more effective and clear.

PR #3602, opened 2016-09-05
fix bug #1122 log.smtp receiver configure error

raphael10241024:
INI file doesn't support array format with bracket "[]",
It may be better to use ',' to split it.


Unknwon:
Thanks!

PR #3616, opened 2016-09-07
Fix sender of issue notifications

### LINKED ISSUE #3615: Notifications emails are not sent due to invalid From
Noticing I was not receiving email notifications I checked the logs and found:

```
gomail: could not send email 1: gomail: invalid address "\"bjornharrtell\" <>": mail: invalid string
```

As the user `bjornharrtell` has a valid email registered (checked via web UI and on the database), I went looking at code and found this in modules/log/smtp.go:

```
    mailmsg := []byte("To: " + strings.Join(s.RecipientAddresses, ";") + "\r\nFrom: " + s.Username + "<" + s.Username +
        ">\r\nSubject: " + s.Subject + "\r\n" + content_type + "\r\n\r\n" + fmt.Sprintf(".%s", time.Now().Format("2006-01-02 15:04:05")) + msg)

    return smtp.SendMail(                                                       
        s.Host,                                                                 
        auth,                                                                   
        s.Username,                                                             
        s.RecipientAddresses,                                                   
        mailmsg,                                                                
    )             
```

Note how `s.Username` is used for the `From` field both in the mail header and in the `smtp.SendMail` 3rd parameter.

I suspect `smtp.SendMail` just refuses such `From` address (lacking an `@` character) and fails to send anything.

I believe things did work as of 0.9.68, while are failing as of 0.9.99

strk:
It is the FROM field in mailer configuration that needs be used,
not the USER field, which is for authentication.

Closes #3615


models/mail.go
```diff
@@ -160,7 +160,7 @@ func composeIssueMessage(issue *Issue, doer *User, tplName base.TplName, tos []s
 	if err != nil {
 		log.Error(3, "HTMLString (%s): %v", tplName, err)
 	}
-	msg := mailer.NewMessageFrom(tos, fmt.Sprintf(`"%s" <%s>`, doer.DisplayName(), setting.MailService.User), subject, content)
+	msg := mailer.NewMessageFrom(tos, fmt.Sprintf(`"%s" <%s>`, doer.DisplayName(), setting.MailService.From), subject, content)
```
Unknwon: This is not going to work, because this setting can contain `name <email>`, not just email address.

strk: With 6e0a1a93d99ecc9549d4609d116dbc0f64023aab it would not be
a problem for the `FROM` to be `name <email>`.

BTW, according to the `Config Cheat Sheet`, `USER` could also not
be an email (it says "*usually* your e-mail address").

models/mail.go
```diff
@@ -160,7 +161,12 @@ func composeIssueMessage(issue *Issue, doer *User, tplName base.TplName, tos []s
 	if err != nil {
 		log.Error(3, "HTMLString (%s): %v", tplName, err)
 	}
-	msg := mailer.NewMessageFrom(tos, fmt.Sprintf(`"%s" <%s>`, doer.DisplayName(), setting.MailService.User), subject, content)
+	sender, err := mail.ParseAddress(setting.MailService.From)
```
Unknwon: I think we should put this process in the `modules/setting` while loading email configs? Add a `FromEmail` field in the struct https://github.com/gogits/gogs/blob/master/modules/setting/setting.go#L716 . So we can print error and halt at start time.

strk: Good idea, done in commit 7216f446f5623fe1c149c7a2d7bc0b44aeb4b34d

strk:
Sorry but I'm afraid the branch attached to this PR cannot be merged into Gogs anymore as it was rebased to Gitea develop branch. Closing.

Issue #1263, opened 2015-05-10
Will there be a packager.io package for Ubuntu 15.04?


gutenye:
The big changes for Ubuntu 15.04 is use systemd as default init .

Unknwon:
Well... you have to ask @crohr about it. :satisfied:

crohr:
@gutenye only LTS versions of ubuntu will be supported by Packager.io, so you'll have to wait for ubuntu 16.04. Note that systemd is already supported for other distributions (centos 7 and debian 8), so that's not a blocking point.

tboerger:
@crohr it's sad to have only that limited amount of distros for packaging, especially for that easy kind of packages with golang applications :(

crohr:
Well, true for Go apps (but Packager is not limited to Go only), but at the same time I'm not sure why you would want to use anything other than LTS versions for your servers.

Unknwon:
Close for now, but open for continuing discussion.

dudheit:
Hi, same question for Ubuntu 16.04 :-)

Buzut:
@dudheit Meanwhile you can use the debian package. As both Debian 8 and Ubuntu 16.04 use systemctl its works perfectly!

tmortagne:
packager.io now claim they support Ubuntu 16.04 but I don't see it listed on https://packager.io/gh/pkgr/gogs/install?bid=481#ubuntu-14-04-gogs so I guess it require something on gogs side too.

tboerger:
It have to be enabled at https://github.com/gogits/gogs/blob/master/.pkgr.yml

Unknwon:
Free to send a PR :)

angelog:
Is it as simple as this?
```  ubuntu-16.04:
<<: *debian```
Currently looking for solutions to my broken instance.. :)

PR #3617, opened 2016-09-07
Enable Ubuntu 16.04 build at packager.io

gianperrone:
This enables building a package for Ubuntu 16.04 LTS (Xenial) as suggested in #1263. I seriously have no idea why bzr and mercurial are needed as build dependencies, but the build only proceeds with them; I got the idea from the packager.io go example project.

The build succeeds: https://packager.io/gh/gianperrone/gogs/build_runs/8#71773
I have a test installation running and found no problems.

### DIFF
--- a/.pkgr.yml
+++ b/.pkgr.yml
@@ -7,10 +7,15 @@ targets:
       - git
   debian-8:
     <<: *debian
-  ubuntu-14.04:
-    <<: *debian
   ubuntu-12.04:
     <<: *debian
+  ubuntu-14.04:
+    <<: *debian
+  ubuntu-16.04:
+    <<: *debian
+    build_dependencies:
+      - bzr
+      - mercurial
   centos-6: &el
     build_dependencies:
       - pam-devel


.pkgr.yml
```diff
@@ -7,10 +7,15 @@ targets:
...
   debian-8:
     <<: *debian
+  ubuntu-12.04:
+    <<: *debian
   ubuntu-14.04:
     <<: *debian
-  ubuntu-12.04:
+  ubuntu-16.04:
     <<: *debian
+    build_dependencies:
```
bkcsoft: Why were these added? and why not add them to `debian-7` instead as I'm guessing all builds would need them?

gianperrone: This is an example of a build without these dependencies: https://packager.io/gh/gianperrone/gogs/build_runs/1#71723

As you can clearly see, not all builds need it. The idea to add these came from looking at the log of the failing build (see link) and from looking at this history: https://github.com/pkgr/example-go-app/commits/master
Apparently bzr and mercurial are needed in all builds, but apart from Ubuntu 16.04 they are installed by default on the build images.

MrFloppy:
Please merge this PR, as I really like to use the package manager for gogs on my ubuntu 16.04.

jhasse:
systemd configuration is still missing though, correct?

PR #3162, opened 2016-06-02
[WIP] better memory-handling for git over HTTP #636

bkcsoft:
Should fix #636 when it works...

Not working as intended :unamused: 

Basically what I've done is instead of reading the entire http body into memory, I split the body-reader with `io.TeeReader` and `io.Pipe` and read both in parallel... And now I have concurrency-bugs instead :laughing:

```go
if hr.Config.OnSucceed != nil {
	log.Info("Running callback...")
	hr.Config.OnSucceed(rpc, cr, cmdRet)
	log.Info("Callback ran...")
} else {
	log.Info("Not running callback")
}
log.Info("CB Done waiting for RPC")
if err := <-gitDone; err != nil {
	log.Info("Done with err: %s", err.Error())
} else {
	log.Info("Done without err")
}
```

It prints `Running callback...` but not `Callback ran...`... anyone have any ideas? :)

NOTE: merge-conflicts because @Unknwon pushed fixes to develop, I'll rebase it soon :smiley:

PR #3622, opened 2016-09-08
Avoid reading the whole request body in memory on push

### LINKED ISSUE #636: Large repo push http  : consumption of ram
I tried to push a large repository via http. Everything goes well until the transfer is finished eating ram becomes important (Gogs). It's a shame Gogs is so light yet in other cases. To see if this is not a technical limitation caused by the protocol, but I think it is possible to optimize.

strk:
Fixes #636
Deprecates #3162

@bkcsoft, @Unknwon review welcome

### DIFF
--- a/routers/repo/http.go
+++ b/routers/repo/http.go
@@ -9,7 +9,6 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


strk:
Ok this doesn't work either:
```
[Macaron] PANIC: runtime error: slice bounds out of range
/usr/lib/go-1.6/src/runtime/panic.go:426 (0x42dc09)
        gopanic: reflectcall(nil, unsafe.Pointer(d.fn), deferArgs(d), uint32(d.siz), uint32(d.siz))
/usr/lib/go-1.6/src/runtime/panic.go:21 (0x42c139)
        panicslice: panic(sliceError)
```

Maybe a ReadClose cannot be used here

strk:
NOTE: despite the panic, pushes succeed (could be a panic triggered by the "Close" part of ReadCloser)

bkcsoft:
I've looked into this one previously and it's a bit more difficult than that. We have to move some code into post-recieve-hook which means a large refactor...

strk:
will close, not knowing what I'm doing anyway :)

Issue #1294, opened 2015-05-25
Feature request: set max repository size


csu:

Unknwon:
Thanks your feedback!

By `max repository size`, do you mean the Git repository size, or the number of repository?

csu:
Size as in Git repository size (e.g. a limit in KB or MB for non-admin users).

Unknwon:
@csu  thanks for clarifying!

vroomanj:
@Unknwon : I'm also interested in seeing this feature.

vroomanj:
@Unknwon Will this by chance make it into any milestone anytime soon? It is the only thing holding me back from deploying Gogs in my College.

Unknwon:
@vroomanj what is the correct git command to check repository objects file size?

bkcsoft:
@Unknwon `du`, don't thinkngit itself has any clue about how big a repo is
in total

Unknwon:
@bkcsoft I've tried `git count-objects` which does give a size of object files, but it is different result from `du`... not sure which one should I believe.

bkcsoft:
`du` is size on disk, which is what most sysops care about

Unknwon:
@bkcsoft but Windows does not have that, right?...

vroomanj:
@Unknwon: @bkcsoft is right `du` is what a sysop is going to be concerned about. Unfortunately, Windows doesn't have an equivalent that I am aware of (that is built into the OS). They do have https://technet.microsoft.com/en-gb/sysinternals/bb896651 but that requires a download and probable couldn't be included in Gogs because of licensing implications.

bkcsoft:
You could do a `filepath.WalkTree` and add the sizes... it'll be about as
slow as `du` (which does exactly that...) and be platform-agnostic :wink:

Unknwon:
@bkcsoft seems the best idea right now.

The ideal solution would be able to prevent/reject git push that exceeds size limitation right now. Both the `du` and `filepath.WalkTree` can't prevent a super large push that exceeds size limitation a lot.

drsect0r:
@Unknwon @bkcsoft Did you try `git count-objects -v` and `git count-objects -vH`? These come pretty close to their actual value. You may need to run `git gc` before doing this.

Unknwon:
@drsect0r thanks!

But after `git gc`:

```
$ git count-objects -vH
count: 0
size: 0 bytes
in-pack: 33
packs: 1
[...]

Issue #1158, opened 2015-03-26
Ability to view repository size


boarpig:
I think it would be useful to be able to see the repository size when viewing either the repository main window or in the settings windows.

It would also be useful to be able to see repository size for each repo in Admin panel's repository list.

jacobhands:
I'm going to take a shot at this.

drsect0r:
+1 @jacksonpan how is it going? If you're currently not working on this; I'm going to take a shot this weekend.

jacobhands:
Hey yeah, life happened etc and I haven't had time. Go for it.

DblK:
Should be linked to #1294 for the way to compute size.

@drsect0r are you still working on this one?

drsect0r:
@DblK It is still on my todo list. Feel free if you want to have a go at it.

DblK:
I modified in the admin section to see the size:
![image](https://cloud.githubusercontent.com/assets/832617/18352934/a18ed5a2-75e0-11e6-993e-6c1fcfd62317.png)

Need to finish implementation of triggering repo size before PR.
For now it's only a fake data in the database but will be soon operational once the triggering will be done.
I will add in another PR after this one will be merged, the user display size and repo.

DblK:
Now it's working !
![image](https://cloud.githubusercontent.com/assets/832617/18387049/8f7a4fa8-769a-11e6-9069-f4065c786295.png)

To keep the PR small, I did not implemented the user menu to display repository including the size.
Once the first PR will be merge, I will make the GUI change for user settings.

DblK:
Once the PR #3631, I already have the next PR for closing this issue.
Here is a screenshot:

![image](https://cloud.githubusercontent.com/assets/832617/18469769/ae4d23a8-79a9-11e6-8382-0123eb28e37c.png)

I will continue to tweak the display. Maybe a table for another display same as GitHub.
Here is in GitHub:
![image](https://cloud.githubusercontent.com/assets/832617/18469818/f2cba86a-79a9-11e6-90be-8b924fe896be.png)

Any thought?

Unknwon:
Patch has pushed to fix this issue, please test on `develop` branch. Currently only admin can view size in admin panel. For user to view own repo sizes, please follow https://github.com/gogits/gogs/issues/4277.

Unknwon:
![image](https://cloud.githubusercontent.com/assets/2946214/23829814/ca81e8dc-06d1-11e7-9921-62aeda396d07.png)

[...]

PR #3631, opened 2016-09-09
Implementation of the size of repo

DblK:
This PR is for the first step to answer #1158 and #1294 by implementing the size of the repo in database.
It's working when fork, migrate, push and create.

Once this PR will be merge, the implement in user setting (#1158) could be easily done and #1294 could use this implementation to make a restriction.

The only GUI change made, is in the admin panel to display the size inside the repo list as following:
image

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -982,6 +982,7 @@ repos.private = Private
 repos.watches = Watches
 repos.stars = Stars
 repos.issues = Issues
+repos.size = Size
 
 auths.auth_manage_panel = Authentication Manage Panel
 auths.new = Add New Source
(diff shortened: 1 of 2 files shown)


DblK:
Here is a better screenshot:

image

LefsFlarey:
Looks really good. One question though, are the inconsistencies with KB having decimal places and the MB not having decimal places intended?

DblK:
I'm using an existing function "humanateBytes" in "modules/base/tool.go".
So the inconsistencies are due to this one.

I don't know if I should fix this or not.

bkcsoft:
> I'm using an existing function "humanateBytes" in "modules/base/tool.go".
> So the inconsistencies are due to this one.
> I don't know if I should fix this or not.

Should be fixed in a separate PR since it's not related to this PR (even
though it's being used)

Unknwon:
Closed by https://github.com/gogits/gogs/commit/bb19bb601e25201982b6b8d205d1a2d237ff22ad.

PR #3637, opened 2016-09-10
Fix string format verbs

alunegov:
This PR fix string format verbs in log messages.

### DIFF
--- a/models/pull.go
+++ b/models/pull.go
@@ -782,7 +782,7 @@ func TestPullRequests() {
 
 		pr, err := GetPullRequestByID(com.StrTo(prID).MustInt64())
 		if err != nil {
-			log.Error(4, "GetPullRequestByID[%d]: %v", prID, err)
+			log.Error(4, "GetPullRequestByID[%s]: %v", prID, err)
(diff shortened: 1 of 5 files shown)


alunegov:
To prevent this kind of errors we could rename log functions to have an "f" at the end and let **go vet** do the checks.

LefsFlarey:
This is not a good idea. Traditional C++ variables like `%s` and `%d` denote `string` and `decimal/digit` types respectively and there is a reason why it is like that. We don't want strings in a decimal variable and vice versa and as such, another solution has to be implemented.

*On a site note though, what is this supposed to fix in the first place?*

alunegov:
Originally this PR fixes wrong format verbs. For example, in models/update.go were used **%d** for **opts.PusherName**, which is actually a string. In http.log this produce [I] Reference '...' has been deleted from '...' by **%!d(string=...)**.

LefsFlarey:
I'm not too sure if this is right or wrong, maybe @Unknwon could give his insight on this...

Unknwon:
Thanks!

PR #3642, opened 2016-09-12
Store password hashes prefixed with the used scheme/algorithm + BCRYPT support

gronke:
- Stores the password scheme and hash options among with the hash itself.
- Users can still sign in with their deprecated passwords (which allows password hash migration paths on existing installations)
- `PASSWORD_HASH_ALGORITHM` can be configured in app.ini
- New users are always created with the configured hash algorithm
- Stored hashes are updated on next successful user login
### Supported Hash Functions
- BCRYPT _(default)_
- PBKDF2-HMAC-SHA256
- PBKDF2-HMAC-SHA512

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -167,7 +167,7 @@ COOKIE_USERNAME = gogs_awesome
 COOKIE_REMEMBER_NAME = gogs_incredible
 ; Reverse proxy authentication header name of user name
 REVERSE_PROXY_AUTHENTICATION_USER = X-WEBAUTH-USER
-; Password Hashing Algorithm
+; Password Hashing Algorithm (available options: sha256, sha512)
 PASSWORD_HASH_ALGORITHM = "sha256"
 
 [service]
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -167,6 +167,8 @@ COOKIE_USERNAME = gogs_awesome
 COOKIE_REMEMBER_NAME = gogs_incredible
 ; Reverse proxy authentication header name of user name
 REVERSE_PROXY_AUTHENTICATION_USER = X-WEBAUTH-USER
+; Password Hashing Algorithm
+PASSWORD_HASH_ALGORITHM = "sha256"
 
 [service]
 ACTIVE_CODE_LIVE_MINUTES = 180
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -167,8 +167,6 @@ COOKIE_USERNAME = gogs_awesome
 COOKIE_REMEMBER_NAME = gogs_incredible
 ; Reverse proxy authentication header name of user name
 REVERSE_PROXY_AUTHENTICATION_USER = X-WEBAUTH-USER
-; Password Hashing Algorithm (available options: sha256, sha512)
-PASSWORD_HASH_ALGORITHM = "sha256"
 
 [service]
 ACTIVE_CODE_LIVE_MINUTES = 180
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -167,6 +167,8 @@ COOKIE_USERNAME = gogs_awesome
 COOKIE_REMEMBER_NAME = gogs_incredible
 ; Reverse proxy authentication header name of user name
 REVERSE_PROXY_AUTHENTICATION_USER = X-WEBAUTH-USER
+; Default password hash algorithm, either "BCRYPT", "PBKDF2-HMAC-SHA512" or "PBKDF2-HMAC-SHA256"
+PASSWORD_HASH_ALGORITHM = PBKDF2-HMAC-SHA512
 
 [service]
 ACTIVE_CODE_LIVE_MINUTES = 180
--- a/models/user.go
+++ b/models/user.go
@@ -325,15 +325,32 @@ func (u *User) EncodePasswd() {
 // ValidatePassword checks if given password matches the one belongs to the user.
 func (u *User) ValidatePassword(clear string) bool {
 
+	needsPasswordUpgrade := false
+
 	verified, err := password.Verify(clear, u.Passwd)
 	if(err != nil) {
 
-		// try with backwards compatibility
-		return password.VerifyBackwardsCompatible(clear, u.Passwd, []byte(u.Salt))
+		if(verified == false) {
+			// backwards compatibility
+			verified = password.VerifyBackwardsCompatible(clear, u.Passwd, []byte(u.Salt))
+		}
+
+		needsPasswordUpgrade = true
 
 	}
 
-	return verified == true
+	if (verified == true) && (needsPasswordUpgrade == true) {
+
+		updatedPasswordHash, err := password.Hash(clear)
+		if(err == nil) {
+			u.Salt = ""
+			u.Passwd = updatedPasswordHash
+			UpdateUser(u);
+		}
+
+	}
+
+	return (verified == true)
 
 }
 
--- a/models/user.go
+++ b/models/user.go
@@ -315,7 +315,7 @@ func (u *User) NewGitSig() *git.Signature {
 
 // EncodePasswd encodes password to safe format.
 func (u *User) EncodePasswd() {
-	newPasswd, err := password.Hash(u.Passwd, u.Salt, 10000, 50, "PBKDF2-HMAC-SHA256")
+	newPasswd, err := password.Hash(u.Passwd)
 	if(err != nil) {
 		panic("password encoding failed")
 	}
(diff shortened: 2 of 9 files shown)


models/user.go
```diff
@@ -315,7 +317,14 @@ func (u *User) NewGitSig() *git.Signature {
 
 // EncodePasswd encodes password to safe format.
 func (u *User) EncodePasswd() {
-	newPasswd := base.PBKDF2([]byte(u.Passwd), []byte(u.Salt), 10000, 50, sha256.New)
+	var hashAlgorithm func() hash.Hash
+	switch setting.Cfg.Section("security").Key("PASSWORD_HASH_ALGORITHM").String() {
+	case "sha512":
+		hashAlgorithm = sha512.New
+	default:
```
bkcsoft: Should probably be specific?
Like this:
```go
switch hash := setting.Cfg.Section("security").Key("PASSWORD_HASH_ALGORITHM").String() {
  case "sha512":
    hashAlgorithm = sha512.New
  case "sha256":
    hashAlgorithm = sha256.New
  default:
    log.Error(log.ERROR, `Unknown HashAlgorithm "%s"`, hash);
}
```
Otherwise people might set it to "hmac" and think that it works :unamused:

bkcsoft: Also this should be done once on setup, not each time a user sets their password...

conf/app.ini
```diff
@@ -167,6 +167,8 @@ COOKIE_USERNAME = gogs_awesome
 COOKIE_REMEMBER_NAME = gogs_incredible
 ; Reverse proxy authentication header name of user name
 REVERSE_PROXY_AUTHENTICATION_USER = X-WEBAUTH-USER
+; Password Hashing Algorithm
```
bkcsoft: Add a line with the currently available algorithms, like this:
```ini
; - Available algorithms are: sha256, sha512
```

models/user.go
gronke: @bkcsoft totally agree, the hash algorithm should always be explicit

gronke:
@bkcsoft thank you for your feedback. I'm not sure if the `updateHashAlgorithm` approach is appropriate.

Before that I tried accessing the `settings` object within the the user model's `main()` function, but did not have access to it yet (exited with runtime error). IMO bombing out while starting the application with invalid settings is better than to wait for the first user attempting to login.

models/user.go
```diff
@@ -317,13 +334,7 @@ func (u *User) NewGitSig() *git.Signature {
...
 // EncodePasswd encodes password to safe format.
 func (u *User) EncodePasswd() {
-	var hashAlgorithm func() hash.Hash
-	switch setting.Cfg.Section("security").Key("PASSWORD_HASH_ALGORITHM").String() {
-	case "sha512":
-		hashAlgorithm = sha512.New
-	default:
-		hashAlgorithm = sha256.New
-	}
+	updateHashAlgorithm()
```
bkcsoft: This still runs once on each password-change... see this for reference: https://github.com/gogits/gogs/pull/2406/files

bkcsoft:
@gronke see this for reference: https://github.com/gogits/gogs/pull/2406/files

Especially this: https://github.com/bkcsoft/gogs/blob/1ab8a60d737b278c176d0b6204843a79dab0e878/routers/install.go#L96
and this: https://github.com/bkcsoft/gogs/blob/1ab8a60d737b278c176d0b6204843a79dab0e878/modules/base/tool.go#L36

gronke:
@bkcsoft Nice, that was exactly what I was looking for. You had a good eye seeing my lack of best practice patterns during my first steps with Go. Many thanks for that!

LefsFlarey:
Would be really nice if say, you could add a feature to allow auto-upgrade from sha256 to sha512...

gronke:
I think auto-upgrading password hashes should go into another pull-request and would also require us to store the password scheme (like [described in the Dovecot docs](http://wiki.dovecot.org/Authentication/PasswordSchemes)). The intention of this PR was to introduce an option that changes the global default. The list of supported hashing functions can be later extended based on this.

LefsFlarey:
Yet if this pull-request gets accepted, lots of discrepancies regarding different users having different password hash will complicate things. A system to store the hash used should come first. That way, we don't have to force a password reset on everyone.

gronke:
@LefsFlarey let's move the discussion around that to #3662 and focus this PR on optionally replacing sha256 with sha512 entirely.

LefsFlarey:
On-topic, the difference between `SHA-256` and `SHA-512` is merely the fact that the former is generated through 32-bit words versus the latter, which is generated through 64-bit words. Strength-wise, they are more or less, equivalent.

While I think this feature implementation may be of some use, I **will have to advise that until implementation of #3662** has been done, implementing this pull-request will ***break*** all current user accounts that rely on the old `SHA-256` hashing algorithm. This means that this pull-request, will only effectively be useful for fresh installations from this commit onwards and not so much on older installations.

**TL;DR: ** Bad idea to merge this PR till implementation of some sort of backward compatibility is done.

gronke:
@LefsFlarey my initial intention was not to increase security, but to open the door for changing the algorithm dynamically. You convinced me that this is only of value when providing an upgrade path for existing installations.

FSMaxB:
You are talking about the hash function to be used in PBKDF2, not directly using SHA256 or SHA512 right?

Because both SHA256 and SHA512 would be horribly bad for password hashing if used directly.

And maybe you should consider making it switchable between PBKDF2, bcrypt and scrypt instead of switching the input to PKBDF2.

gronke:
> You are talking about the hash function to be used in PBKDF2, not directly using SHA256 or SHA512 right?
@FSMaxB yes, only the hash function used in PBKDF2 is replaced here.

Currently I'm looking into parsing the password scheme stored in the database and select the the hashing library according to that. In the first step I'd like to target `{SHA256}` (default when no scheme present) and `{SHA512}` and stick with PBKDF2. Later we can easily extend the list of supported schemes and add `{SHA512-CRYPT}`, `{SSHA}` or whatever schemes may come to our mind.

LefsFlarey:
@FSMaxB, I would like to add using the hashes directly isn't actually horribly bad. If you do your homework, learn to salt and learn to iterate, it's exactly like PBKDF2.

gronke:
The latest changes add support for dynamic password scheme detection like [described in the Python Passlib documentation](https://pythonhosted.org/passlib/lib/passlib.hash.pbkdf2_digest.html#passlib.hash.pbkdf2_sha1). Both the Salt and the Checksum are stored Base64 encoded.

## Supported Schemes:
- PBKDF2-HMAC-SHA256
- PBKDF2-HMAC-SHA512

## Example 
#### Input
Iterations: 10000
keyLength: 50
Salt: `cxewZCrp2T`
Password: `demo`

#### Sample Output
```
{PBKDF2-HMAC-SHA256}$10000$aFNFcldwcmludQ==$GcvY3b3CE5PWx5SSLB3kvhtYrA9My7pQ9Ln8F5psWl+GZznnIYy3oGQTxOaWYHkmrZU=
```

There is no need to store the keyLen, because we can easily calculate it from the given checksum.

## References used
- https://pythonhosted.org/passlib/lib/passlib.hash.pbkdf2_digest.html#passlib.hash.pbkdf2_sha1
- https://en.wikipedia.org/wiki/Crypt_(C)
- http://wiki.dovecot.org/Authentication/PasswordSchemes

## To Do
The next feature should be automatically upgrading deprecated password schemes on login. We can add a configuration option that defines the target one that should be propagated with future logins.

gronke:
@LefsFlarey @bkcsoft may I ask you for another review?

LefsFlarey:
@gronke, I'm personally not too certain if I, as a sysadmin, would like to give control over password hash storage to the endusers. I'm going to have to sit out of this discussion from here on and wait for someone like @Unknwon to decide upon this.

gronke:
The control is **not** in the hands of the end-users, but in control of the system administrator. The admin can define the preferred hash algorithm and on the next login of a user the stored password is updated. The system supports mixed password hashes in the database, so that users can still login after a change.

@LefsFlarey Is that solution a better fit for you?

LefsFlarey:
@gronke, not for me. There are plenty of people who fork this project without knowing a single grain of salt about hashes or password security. Those are who I am referring to.

FSMaxB:
All of those are solid choices and it's good to have infrastructure in place for changing the hash if something happens (like new found vulnerabilities).

gronke:
> not for me. There are plenty of people who fork this project without knowing a single grain of salt about hashes or password security. Those are who I am referring to.

@LefsFlarey All of the choices available provide the same or higher level of security than the current implementation. To support other hashing algorithms more manual work is involved. If somebody decides to fork Gogs and implement bad security, that is absolutely not our fault.

PR #3654, opened 2016-09-15
Forces a confirmation page on 'Forgot Password' page even if user does not exist

### LINKED ISSUE #3639: user/forget_password confirmation/error messages should be changed
You should change the error output the page "user/forget_password" gives when the email address is not known. Currently it is "This email address is not associated with any account." while it should just print a positive confirmation like mailman does[0] something like "if this email account exists (etc etc)", likewise the positive result "A new confirmation email has been sent to name@host.tld, please check your inbox within the next 3 hours to complete the registration process." should be something similar.

At the moment I don't know enough about gogs to fix it myself, I'm just packaging it.

[0]: "If you are a list member, your password has been emailed to you."

LefsFlarey:
# TLDR; Closes #3639 
This small hot-fix is basically a very cheat way of showing a successful 'confirmation email' sent to the email even if it does not exists.

# Why?
Well, by allowing users to see if an email exists breaks a very fundamental law of security through obscurity. This small patch just forces Gogs to show a confirmation page even if the user does not exist.

### DIFF
--- a/routers/user/auth.go
+++ b/routers/user/auth.go
@@ -341,8 +341,10 @@ func ForgotPasswdPost(ctx *context.Context) {
 	u, err := models.GetUserByEmail(email)
 	if err != nil {
 		if models.IsErrUserNotExist(err) {
-			ctx.Data["Err_Email"] = true
-			ctx.RenderWithErr(ctx.Tr("auth.email_not_associate"), FORGOT_PASSWORD, nil)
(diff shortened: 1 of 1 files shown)


LefsFlarey:
Ah, I'm not sure if @Unknwon actually wants this feature. In my opinion, it's important that validity of email address should not be disclosed and this patch ensures confidentiality of your user's emails.

Unknwon:
Thanks!

Issue #3649, opened 2016-09-14
@mention link is not parsed correctly


xor-gate:
- Gogs version (or commit ref):  0.9.99.0903
- Can you reproduce the bug at https://try.gogs.io:
  - [X] Yes: https://try.gogs.io/admin1/mention_rendering/issues/1

## Description

Mentioning with `@` is not parsed correctly. On github you can have a line ending dot or place it between parentheses. Which is not correctly parsed in gogs. The one with the dot at the end will link to a non-existing profile. This is how it works on github:

* Works: @xor-gate 
* Broken (dot after mention is not stripped). @xor-gate.
* Broken (between parentheses). (@xor-gate).

DblK:
Seems to be a regexp problem.
Will look at it.

DblK:
General remarks: The mention in markdown are done by regexep only without checking if a user exists or not; that is the implementation @Unknwon wanted (at least for now).

Test 1: Okay, working
Test 2: The dot character is an accepted character for the name so it's working as expected. ==> Nothing to fix
Test 3: The parenthesis has been an issue; I have just made a PR for this issue (#3655).

Unknwon:
Fixed by #3655.

PR #3655, opened 2016-09-15
Fix mention in comments like (@dblk)

DblK:
It a small fix for issue #3649.

### DIFF
--- a/modules/markdown/markdown.go
+++ b/modules/markdown/markdown.go
@@ -76,7 +76,7 @@ func IsReadmeFile(name string) bool {
 
 var (
 	// MentionPattern matches string that mentions someone, e.g. @Unknwon
-	MentionPattern = regexp.MustCompile(`(\s|^)?@[0-9a-zA-Z-_\.]+`)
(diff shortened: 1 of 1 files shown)


LefsFlarey:
Looks good, though I don't think the expected behaviour is somethinglikethis@dblk

DblK:
Thank for your reply. I will try to fix the later example.
I didn't know the negative lookbehind. Will have a look for Golang and make another commit.

DblK:
The RegExp you suggested does not seems to work in Golang.
I need to find another trick.

LefsFlarey:
@DblK, as you have reported, apparently GoLang does not support lookaround. As such, here's an alternative that ***should*** work.

[`(?:[^\w])@([0-9a-zA-Z-_\.]+)`](https://regex101.com/r/zM8dC4/2)

If extraction of the user is important, extract from group 1 through `\1` or `$1`.

DblK:
The above RegExp does not work either. I fixed in another way.

LefsFlarey:
Alright, that seems to work along the same principles too, but CI kinda failed there right there...

xor-gate:
👍 looking good

Unknwon:
Thanks!

PR #3665, opened 2016-09-17
Pluralize "Statistic" for English locale to be consistent with others

ipavl:
Other locales appear to use the plural version, and the text is being associated with more than one statistic.


Unknwon:
Thanks!

Issue #3666, opened 2016-09-17
"?" handling in folder names


tuudik:
- Gogs Version: 0.9.99.0903 
- Operating system: Ubuntu 16.04
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
## Description

Having "?" in folder names, will cause error 404.
For example:
- "Are we there?" will direct to: https://mygit.eu/user/repo/src/master/Are%20we%20there?
  - This will not work, it will cause error 404
- When changing "?" in URL to "%3F": https://mygit.eu/user/repo/src/master/Are%20we%20there%3F
  - This works perfectly

LefsFlarey:
That is unfortunate but `?` symbols are almost universal for stating GET variables. In this case, perhaps a small fix can be made to url_encode it first?

LefsFlarey:
I can try to work on this but I need to know where this error occurs.

0xbaadf00d:
Using "?" for a file or directory name is not common and some file systems can't handle special char (like NTFS on Windows by instance)

tuudik:
I think it doesnt matter if it's common to use somekind of symbols or not imho. Currently the issue is that URL encoding is not done properly.

0xbaadf00d:
Just tested on Github.  Github replace ? with %3F on URL. I can push/clone on Linux, but not on windows.

@LefsFlarey You can probably just replace special char (eg: ? = , ; &) by the URL encode value.  Slash must stay as "/" to avoid weird URL.

LefsFlarey:
Okay, I am now doing some testing for fixing this issue.

LefsFlarey:
I'm done :)
Let me know if I missed anything but it should work in every sense.

Unknwon:
Should have been fixed by https://github.com/gogits/gogs/pull/3668.

PR #3668, opened 2016-09-18
Escaping '?' symbols in filenames

LefsFlarey:
Very simply, this pull request modifies certain files to escape symbols in filenames that may cause 404s unescaped.

This solves issues #3666 and #3674.

### DIFF
--- a/modules/template/template.go
+++ b/modules/template/template.go
@@ -97,7 +97,7 @@ func NewFuncMap() []template.FuncMap {
 		"MD5":               base.EncodeMD5,
 		"ActionContent2Commits": ActionContent2Commits,
 		"EscapePound": func(str string) string {
-			return strings.NewReplacer("%", "%25", "#", "%23", " ", "%20").Replace(str)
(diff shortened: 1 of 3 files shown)


LefsFlarey:
Additionally, I would like to suggest changing `EscapePound` to something more appropriate like `EscapeHTML`.

LefsFlarey:
Like mentioned, this quick-fix is for #3666. I'm not sure if any other symbols are affected, particularly symbols that are usually URL-encoded.

LefsFlarey:
I've found an additional bug that occurs during creation of files with weird names. Pushed commit to fix.

silverwind:
Wouldn't it be better to use https://golang.org/pkg/net/url/#QueryEscape?

LefsFlarey:
That would be the case, if not for the fact that some of the links given are like, `src/test/go`. Using normal escape on it would screw up the paths.

tuudik:
@Unknwon Why it has not been merged still? :(

Unknwon:
Thanks!

PR #3676, opened 2016-09-20
Fix 500 error caused by deleted users on issues (#3675)

### LINKED ISSUE #3675: 500 error when userid doesn't exist
On latest version when browsing to a page of issues that has one created by a user who no longer exists:

```
2016/09/20 00:43:17 [D] CSRF Token: ndGTQptT4O0uhq8vu8PXMh5w_Jk6MTQ3NDMzMTkyOTM4OTMxMDUyMQ==
2016/09/20 00:43:17 [...outers/repo/issue.go:183 Issues()] [E] Issues: LoadAttributes [66]: user does not exist [uid: 1, name: ]
```

LefsFlarey:
This PR will resolve #3675.

### DIFF
--- a/models/issue.go
+++ b/models/issue.go
@@ -94,13 +94,8 @@ func (issue *Issue) loadAttributes(e Engine) (err error) {
 	if issue.Poster == nil {
 		issue.Poster, err = getUserByID(e, issue.PosterID)
 		if err != nil {
-			if IsErrUserNotExist(err) {
-				issue.PosterID = -1
-				issue.Poster = NewGhostUser()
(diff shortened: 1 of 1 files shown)


0xbaadf00d:
I think it could be interresting to link issues to another user (like YouTrack).

LefsFlarey:
@0xbaadf00d It's not actually saved to SQL with `-1`. It's just a placeholder for a user that no longer exists. While it could be done to link to another user, it simply is too much work to do via source code, thus the current 'Ghost' user exists.

PR #3685, opened 2016-09-22
[HIGH PRIORITY] Fixed Xorm V6 compile errors

LefsFlarey:
# DO NOT CLOSE THIS. THIS FIXES XORM COMPILE ERRORS.
---
# Description
The newest version of xorm changes the xorm.Engine interface where the Where, is no longer `Where(string, ...interface{}) *xorm.Session` but instead `Where(interface{}, ...interface{}) *xorm.Session`. However, xorm _**[still supports string queries](https://github.com/go-xorm/xorm/blob/master/statement.go#L139-L162)**_.

I've read somewhere that the string query may be deprecated and removed soon, so this fix may or may not be a permanent solution, thus it is best to slowly convert existing database queries to use go-xorm/builder.

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -34,7 +34,7 @@ type Engine interface {
 	InsertOne(interface{}) (int64, error)
 	Iterate(interface{}, xorm.IterFunc) error
 	Sql(string, ...interface{}) *xorm.Session
-	Where(string, ...interface{}) *xorm.Session
+	Where(interface{}, ...interface{}) *xorm.Session
 }
 
 func sessionRelease(sess *xorm.Session) {


gronke:
@Unknwon please 🚢

bkcsoft:
Could you update the glide & gopm-files to reflect the changed version as well

lunny:
I think maybe you have made a mistake. `string query` will be supported always. `github.com/go-xorm/builder` is a new choice. @LefsFlarey

Unknwon:
Thanks!

LefsFlarey:
@lunny I read somewhere that it might not be supported always. Still, might be useful to upgrade it to go-xorm/builder.

PR #3688, opened 2016-09-22
Implement the remaining issue-endpoints

bkcsoft:
This should implement the remaining Issue-endpoints. Had to refactor slightly so as to not duplicate code.


Unknwon:
Thanks!

PR #3696, opened 2016-09-24
Actually use RUN_MODE in config.ini for what I think it is used for.

LefsFlarey:
Pardon me if I'm wrong but apparently RUN_MODE in config is not even being parsed. Which results in the issue #3611. This PR hopefully fixes it.

### DIFF
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -480,6 +480,16 @@ func NewContext() {
 	}[Cfg.Section("time").Key("FORMAT").MustString("RFC1123")]
 
 	RunUser = Cfg.Section("").Key("RUN_USER").String()
+
+	switch mode := Cfg.Section("").Key("RUN_USER").String(); mode {
+	case "prod":
+		ProdMode = true
+	case "dev":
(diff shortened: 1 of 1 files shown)


LefsFlarey:
Completely missed the typo. Anyone can try this patch out and see if #3611 is fixed?

modules/setting/setting.go
```diff
@@ -480,6 +480,16 @@ func NewContext() {
 	}[Cfg.Section("time").Key("FORMAT").MustString("RFC1123")]
 
 	RunUser = Cfg.Section("").Key("RUN_USER").String()
+
+	switch mode := Cfg.Section("").Key("RUN_MODE").String(); mode {
+	case "prod":
+		ProdMode = true
+	case "dev":
+		ProdMode = false
+	case "test":
```
bkcsoft: What is `test` for?

LefsFlarey: Absolutely, nothing. I just follow the standards, which consists of `prod`, `dev` and `test`.

PR #3748, opened 2016-10-10
Use temporary file to avoid out-of-memory when receiving big chunks.

### LINKED ISSUE #636: Large repo push http  : consumption of ram

typeless:
Not perfect but I think it's a reasonable solution.
For small request bodies, I suppose performance wouldn't be an issue.
For large ones, this seems to be a necessary evil.

Fixes #636

### DIFF
--- a/routers/repo/http.go
+++ b/routers/repo/http.go
@@ -364,14 +364,22 @@ func serviceRPC(h serviceHandler, service string) {
 	}
 
 	if h.cfg.OnSucceed != nil {
-		input, err = ioutil.ReadAll(reqBody)
+		tmpfile, err := ioutil.TempFile("", "gogs")
 		if err != nil {
-			log.GitLogger.Error(2, "fail to read request body: %v", err)
+			log.GitLogger.Error(2, "fail to create temporary file: %v", err)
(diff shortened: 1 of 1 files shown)


strk:
@typeless would you consider moving this to http://github.com/go-gitea/gitea ?
@bkcsoft this was the PR I was looking for...

typeless:
@strk I am not sure what you mean by 'moving' the PR but I can try to send the PR to http://github.com/go-gitea/gitea.

strk:
@typeless yes, send the PR to gitea was the request, rebasing it against gitea "master" branch.

Unknwon:
Thanks!

Timetopretend:
Thanks 😉

barracks510:
Sorry, but this set of patches only worsens the issue on modern Linux systems. `ioutil.TempFile` creates a file on `/tmp` which for the majority of Linux systems is mounted as tmpfs, and is shared with system RAM.

PR #3761, opened 2016-10-13
init.d script for Gentoo Linux

afilippov1985:
init script for gentoo

place it into `/etc/init.d` directory
then run `service gogs start` or `/etc/init.d/gogs start`


Unknwon:
Thanks!

PR #3763, opened 2016-10-13
New settings option for a custom SSH host

imilo:
Mostly security reasons.

There should be an option to make a builtin SSH server listen on the specified network interface. Currently it uses **0.0.0.0** only. If somebody has several network interfaces: to the Internet, to the intranet, to the localhost etc. - Gogs shouldn't open a service to the Internet without any options.

You could say "There is no shell", but it is still a service and every service potentially is vulnerable.

I've added some code changes to make it configurable and **0.0.0.0** by default like HTTP server.


Unknwon:
Thanks!

PR #3765, opened 2016-10-13
Fix rendering of non-markdown readme files

### LINKED ISSUE #3749: README.txt file rendered with injected HTML and shown as empty inline
A README.txt file in the root of repositories seems to trigger some code to automatically show the content inline but the content is shown as empty. Clicking on the file to show it in the browser shows it as including HTML while there is no HTML in the real file.

You can see this on try.gogs.io: https://try.gogs.io/strk/readme.txt

If you clone the repository you can see there is no HTML in the file, but the file browser injects the html:
https://try.gogs.io/strk/readme.txt/src/master/README.txt

odinuge:
Rendering of non-markdown readme files is kinda strange.

Fixes #3749

### DIFF
--- a/routers/repo/view.go
+++ b/routers/repo/view.go
@@ -90,6 +90,9 @@ func renderDirectory(ctx *context.Context, treeLink string) {
 				ctx.Data["IsMarkdown"] = true
 				buf = markdown.Render(buf, treeLink, ctx.Repo.Repository.ComposeMetas())
 			default:
+				// FIXME This is the only way to show non-markdown files
+				// instead of a broken "View Raw" link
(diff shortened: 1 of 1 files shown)


odinuge:
Isn't "readmeExist" more or less the same as "isMarkdown" in the fileRender function?

`readmeExist := isMarkdown || markdown.IsReadmeFile(blob.Name())`
That means "readmeExist" is true when you are rendering a normal non-readme markdown file.

strk:
@odinuge would you consider migrating this PR to https://github.com/go-gitea/gitea ? Higher chances for it to be merged :)

Issue #3764, opened 2016-10-13
Error with issue template in Spanish localization


einacio:
- Gogs version (or commit ref): 0.9.97 (present in 0.9.99 in try.gogs)
- Git version: Not Relevant
- Operating system: Not Relevant
- Database (use `[x]`): Not Relevant
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (provide example URL) https://try.gogs.io/danil/dotfiles/pulls/5
  - [ ] No
  - [ ] Not relevant
## Description

When opening an issue or a pull request with Spanish language set, the template that renders the comment header is broken, the HTML looks like this:

```
<span class="text grey"><a href="/user">user</a> 'comentado <a href="%!(EXTRA string=issue-77, template.HTML=<span class=" time-since"="" title="Thu, 06 Oct 2016 17:41:04 ART">hace 6 días</a></span><a href="%!(EXTRA string=issue-77, template.HTML=<span class=" time-since"="" title="Thu, 06 Oct 2016 17:41:04 ART">)
                        </a><div class="ui right actions"><a href="%!(EXTRA string=issue-77, template.HTML=<span class=" time-since"="" title="Thu, 06 Oct 2016 17:41:04 ART">

                                </a><div class="item action"><a href="%!(EXTRA string=issue-77, template.HTML=<span class=" time-since"="" title="Thu, 06 Oct 2016 17:41:04 ART">
                                    </a><a class="edit-content" href="#"><i class="octicon octicon-pencil"></i></a>
                                </div>

                        </div>
```

Tested all the languages in try, it's the only one that has this problem (altough čeština has another similar problem)

stroucki:
what is wrong?

einacio:
that's the rendered html, not the template. the attribute values are invalid and it breaks.

it looks like this  (notice the `'` and the `)`, also the tooltip doesn't have style)

![spanish](https://cloud.githubusercontent.com/assets/1534114/19363535/42846c24-9161-11e6-8982-47f46dc36ecc.png)

in čeština the problem is on the text

![cestina](https://cloud.githubusercontent.com/assets/1534114/19363613/7dbfa6c8-9161-11e6-85ca-53b834f1d63b.png)

it seems like a syntax error in the template or something like that, that echoes the wrong thing

stroucki:
The Czech issue is reported in #3791. Looks like a problem in how to handle complex grammar :).

einacio:
[...]

PR #3769, opened 2016-10-13
suggested locale fixes addresses issue #3764

stroucki:
This addresses the Spanish part of #3764 and two instances of non-ascii apostrophes that caught my eye. Also, some untranslated German messages are translated.


stroucki:
Get it from gitea if you want.

Issue #499, opened 2014-09-25
Additional emails for commit resolving


Cysioland:
Now, when you committed using another email than one added in gogs, then clicking on user name returns 404.
I would like to be able to specify additional emails from which I commit, and gogs should take it into account.

PR #3772, opened 2016-10-16
Adding mssql support.

dlob:
This pull request adds support for mssql (Microsoft SQL Server) databases. 

It depends on a pull request to xorm (go-xorm/xorm#469), which is mandatory to get mssql support for gogs without fundamental changes.

### DIFF
--- a/models/migrations/migrations.go
+++ b/models/migrations/migrations.go
@@ -87,6 +87,7 @@ func Migrate(x *xorm.Engine) error {
 	} else if !has {
 		// If the version record does not exist we think
 		// it is a fresh installation and we can skip all migrations.
+		currentVersion.ID = 0
 		currentVersion.Version = int64(_MIN_DB_VER + len(migrations))
 
 		if _, err = x.InsertOne(currentVersion); err != nil {
--- a/models/models.go
+++ b/models/models.go
@@ -13,6 +13,7 @@ import (
 	"path"
 	"strings"
 
+	_ "github.com/denisenkom/go-mssqldb"
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/go-xorm/core"
 	"github.com/go-xorm/xorm"
@@ -85,6 +86,8 @@ func LoadConfigs() {
 		setting.UseMySQL = true
 	case "postgres":
 		setting.UsePostgreSQL = true
+	case "mssql":
+		setting.UseMSSQL = true
 	case "tidb":
 		setting.UseTiDB = true
 	}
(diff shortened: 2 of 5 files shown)


btrepp:
Rebased onto master, when editing LDAP authentication configuration it gives a 500 internal server error, logs show

>  [...uters/admin/auths.go:228 EditAuthSourcePost()] [E] UpdateSource: mssql: Operand type clash: varbinary is incompatible with text

btrepp:
For the second part, It appears xorm doesn't transparently shift between LIMIT/TOP, as limit is not supported in SQL server. Not knowing xorm/go very well, should an issue be raised there?, or is it the responsibility of gogs to detect the DB and translate queries?.

dlob:
@btrepp The first issue should be solved by this pull request to xorm: [#499](https://github.com/go-xorm/xorm/pull/499).

btrepp:
@dlob.! awesome.

From my testing the only other issue is the hardcoded limit expression, which is an easy change for someone who knows xorm. 

Everything else I've encountered are already tracked/not related to mssql support.

dlob:
@btrepp Did you create a new database? The fix does not work with already existing dbs out of the box, because the type definition of the 'cfg' column of the 'login_source' table has changed to varchar(max) instead of text. Otherwise you could do this change manually.

lunny:
Hi, I have merged go-xorm/xorm#499 and push some commits for mssql support.
Limit has been supported by XORM. You should use `Limit(limit, start)`.
Could you also send a PR to github.com/go-gitea/gitea ?

johnharris85:
Any update on whether this is going to get merged?

Unknwon:
@johnharris85 I have to install a SQL server to test it, not easy on a Mac, need download and setup a Windows.

johnharris85:
OK cool. I might be able to help in the next day or so (on Linux but have a Windows box hanging around somewhere I'm sure).

Corben78:
@johnharris85: mssql is also available for [Linux](https://www.microsoft.com/en-us/sql-server/sql-server-vnext-including-Linux) if that helps.
For testing on macOS you would have to use Docker.

Unknwon:
I'm using MSSQL for Linux and failed with:

```
2017/02/10 03:02:59 [FATAL] [...om/urfave/cli/app.go:483 HandleAction()] Fail to initialize ORM engine: sync database struct error: Unknown colType SYSNAME for MSreplication_options - optname
```

dlob:
Did you create a database first? I was able to reproduce the error when no database has been created and mssql uses the default 'master' database.

If you create a database (e.g. 'gogs'), it works. I tested this also with MSSQL for Linux running inside Docker.

There is definitely a need of improvement regarding error handling and error messages in xorm/mssql-driver.

Unknwon:
Ah.. stupid me! I'll test again. :sweat:

Unknwon:
Succeed.

BTW, any recommendations of Mac MSSQL client?

dlob:
@Unknwon I only know this one: https://docs.microsoft.com/en-us/sql/connect/odbc/linux/connecting-with-sqlcmd
But I'm not sure if it also runs on mac. But at least it should run inside a Docker container.

Unknwon:
Thanks for the link! But I wasn't clear about my question... I'll try google with some GUI clients...

PR #3779, opened 2016-10-17
Tweak awkward wording in en-US translation file

sbrl:
I spotted some awkward wording when I was updating my repository's settings, so I decided to put together a quick PR to fix them.


LefsFlarey:
Slightly more awkward since there's not really anything wrong with the original localization.

bkcsoft:
`Repository` and `Organization` should be in upper-case and without "the" :)

LefsFlarey:
@bkcsoft too awkward man.

sbrl:
The main issue that this commit fixes is the [`has` / `have`](http://writingexplained.org/has-vs-have-difference) problem.

> Repository options has been saved.

This feels awkward because _has_ is being used with the plural _options_.

> Repository options have been saved.

This is better, but it would be better if the _repository options_ were quantified with an article, since specific options were saved[[1]](http://www.dictionary.com/browse/the)

> The repository options have been saved successfully.

Alternatively:

> The repository options were saved successfully.

Unknwon:
Thanks for the suggestion, but I don't feel very comfortable to merge...

sbrl:
@Unknwon Sure :-)

PR #3781, opened 2016-10-17
Handle user ssh key input better

### LINKED ISSUE #3780: ssh key import does not work unless very careful
- Gogs version (or commit ref): 2bec8a4f1ef6b48220a0010c1482e561abd55810
- Git version: 2.1.4
- Operating system: Linux
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [x] SQLite
  - [ ] No
  - [ ] Not relevant
- Log gist: Sorry, we're not able to verify your SSH key: extractTypeFromBase64Key: invalid key format: (nil)
## Description

When trying to add an ssh key by copy-paste, the trailing newline results in a two-element slice in models/ssh_key.go/parseKeyString, which causes the wrong branch to be taken when the key is OpenSSH format.

Pull request follows.

stroucki:
Fixes #3780.

ssh_key: when user submitted keys had a newline at the end, strings.Split
would have created a slice with an empty last element, and the key type
check would be incorrect. Perhaps a better way is to look for 'ssh-rsa' or
'ssh-dsa' at the beginning of the string, but this is simple.

### DIFF
--- a/models/ssh_key.go
+++ b/models/ssh_key.go
@@ -106,6 +106,8 @@ func extractTypeFromBase64Key(key string) (string, error) {
 func parseKeyString(content string) (string, error) {
 	// Transform all legal line endings to a single "\n".
 	content = strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(content)
+        // remove trailing newline (and beginning spaces too)
+        content = strings.TrimSpace(content)
 	lines := strings.Split(content, "\n")
 
 	var keyType, keyContent, keyComment string
--- a/models/ssh_key.go
+++ b/models/ssh_key.go
@@ -106,8 +106,8 @@ func extractTypeFromBase64Key(key string) (string, error) {
 func parseKeyString(content string) (string, error) {
 	// Transform all legal line endings to a single "\n".
 	content = strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(content)
-        // remove trailing newline (and beginning spaces too)
-        content = strings.TrimSpace(content)
+	// remove trailing newline (and beginning spaces too)
+	content = strings.TrimSpace(content)
 	lines := strings.Split(content, "\n")
 
 	var keyType, keyContent, keyComment string


models/ssh_key.go
```diff
@@ -106,6 +106,8 @@ func extractTypeFromBase64Key(key string) (string, error) {
 func parseKeyString(content string) (string, error) {
 	// Transform all legal line endings to a single "\n".
 	content = strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(content)
+        // remove trailing newline (and beginning spaces too)
```
bkcsoft: Indentations are off (also on the next line)

stroucki: Thanks, I think I noticed already but had not pushed yet.

models/ssh_key.go
```diff
@@ -106,6 +106,8 @@ func extractTypeFromBase64Key(key string) (string, error) {
 func parseKeyString(content string) (string, error) {
 	// Transform all legal line endings to a single "\n".
 	content = strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(content)
```
LefsFlarey: Can be merged with below.

models/ssh_key.go
```diff
@@ -106,6 +106,8 @@ func extractTypeFromBase64Key(key string) (string, error) {
...
+	// remove trailing newline (and beginning spaces too)
+	content = strings.TrimSpace(content)
```
LefsFlarey: Likewise, it can be merged with the above line like so,

    content = strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(content).TrimSpace(content)

stroucki: Sure, it can be, at the cost of readability.

LefsFlarey: @stroucki Granted, I'll close it for now since it's a valid argument.

strk:
@stroucki could you please move this to http://github.com/go-gitea/gitea ?

stroucki:
@strk will do

Unknwon:
Thanks for the PR!

But I think the current code already done trim for trailing newlines.

PR #3786, opened 2016-10-19
Small UI fixes on full width form titles

joubertredrat:
Small fixed with center title, as below
gogs-title-before
gogs-title-after


public/css/gogs.css
```diff
@@ -31,18 +31,23 @@ pre.wrap,
 code.wrap {
   white-space: pre-wrap;
   /* CSS 3 */
+
   word-break: break-word;
 }
 .dont-break-out {
   /* These are technically the same, but use both */
+
```
bkcsoft: new-lines should be before the comment, not after (for readability)

Unknwon:
Seems having strange commits mixed in.

joubertredrat:
Closed, I removed original repo with this PR

Issue #3789, opened 2016-10-20
Team with "New" in name caused go to the create page of team


robinshi007:
Server info:
- Gogs version (or commit ref):  v0.9.97
- Git version:  1.9.1
- Operating system:  ubuntu 14.04
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io: I have no orgination right to set this.
  - [ ] Yes (provide example URL)
  - [x] No ( have no orgination right)
  - [ ] Not relevant
- Log gist:
## Description

step to reproduce:
1. login gogs site with admin account
2. create a new orgaziation.
3. create a new team in that organization with name "New"
4. go to detail page of the newly created team.

Actual result:
step 4 goes to the create page of team, not detail page of that team. this cause by the url "*/team/new"

LefsFlarey:
Working on a fix right about now, thanks for reporting.

Unknwon:
Name "new" is now reserved word for team name, so users won't mistakenly create a team with such name.

PR #3797, opened 2016-10-22
Adds checking of reserved keywords against team names.

LefsFlarey:
This fixes and solves #3789.


models/org_team.go
```diff
@@ -196,7 +196,11 @@ func (t *Team) RemoveRepository(repoID int64) error {
 
 // NewTeam creates a record of new team.
 // It's caller's responsibility to assign organization ID.
-func NewTeam(t *Team) error {
+func NewTeam(t *Team) (err error) {
+	if err = IsUsableUsername(t.Name); err != nil {
```
bkcsoft: Should probably have it's own check? Since some usernames are valid as team-names

LefsFlarey: Hrm...

LefsFlarey:
Pushing a rebased version to cover up my embarrassing errors.

models/org_team.go
```diff
@@ -194,9 +194,25 @@ func (t *Team) RemoveRepository(repoID int64) error {
 	return sess.Commit()
 }
 
+func IsUsableTeamName(name string) (err error) {
```
bkcsoft: no need to name `err`, just go for `error`

models/org_team.go
```diff
@@ -194,9 +194,25 @@ func (t *Team) RemoveRepository(repoID int64) error {
...
+	var names = []string{"new"}
```
LefsFlarey: Would a `var reservedTeamNames` be better? In terms of capitalizations I mean.

bkcsoft: yeah, but outside the function, since it doesn't change during runtime :smile:

PR #3802, opened 2016-10-24
Update gogs.service

tehscorpion:
Hi,
this is based on caddy's service file and should be more sane.

### DIFF
--- a/scripts/systemd/gogs.service
+++ b/scripts/systemd/gogs.service
@@ -9,13 +9,13 @@ After=network.target
 #After=redis.service
 
 [Service]
-Restart=on-failure
+Restart=always
 
 ; User and group the process will run as.
 User=git
 Group=git
  
-; Gogs directoryies.
+; Gogs directories.
 Environment=USER=git HOME=/home/git
 WorkingDirectory=/home/git/gogs
  
@@ -31,15 +31,9 @@ PrivateTmp=true
 ; Use a minimal /dev
 PrivateDevices=true
  
-; Hide /home, /root, and /run/user. Nobody will steal your SSH-keys.
-ProtectHome=true
- 
-; Make /usr, /boot, /etc and possibly some more folders read-only.
+; Mount /usr, /boot, /etc folders read-only.
 ProtectSystem=full
- 
-; Make our directory writable
-ReadWriteDirectories=/home/git
- 
+
 ; Drop all other capabilities. Important if you run gogs as privileged user (which you should not).
 CapabilityBoundingSet=CAP_NET_BIND_SERVICE
  


scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
-# Modify these two values and uncomment them if you have
-# repos with lots of files and get an HTTP error 500 because
-# of that
-###
-#LimitMEMLOCK=infinity
-#LimitNOFILE=65535
-Type=simple
+Restart=on-failure
+
+; User and group the process will run as.
```
bkcsoft: Useless comment

scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
 User=git
 Group=git
+ 
+; Gogs directoryies.
```
bkcsoft: Useless comment

scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
+Environment=USER=git HOME=/home/git
```
bkcsoft: Why did you move this?

scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
 
 [Service]
-# Modify these two values and uncomment them if you have
-# repos with lots of files and get an HTTP error 500 because
-# of that
-###
-#LimitMEMLOCK=infinity
-#LimitNOFILE=65535
-Type=simple
+Restart=on-failure
```
bkcsoft: Why did you change this? and did you move it?

scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
+; Gogs directoryies.
+Environment=USER=git HOME=/home/git
 WorkingDirectory=/home/git/gogs
+ 
 ExecStart=/home/git/gogs/gogs web
-Restart=always
-Environment=USER=git HOME=/home/git
+
+; Repos with lots of files and get an HTTP error 500 will need that
+;LimitMEMLOCK=infinity
```
bkcsoft: Again, why did you change things without changing them? (same for line above and below)

scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
+;LimitNOFILE=65535
+ 
+; Use private /tmp and /var/tmp, which are discarded after gogs stops.
+PrivateTmp=true
+ 
+; Use a minimal /dev
+PrivateDevices=true
+ 
+; Hide /home, /root, and /run/user. Nobody will steal your SSH-keys.
+ProtectHome=true
```
bkcsoft: This will make Gogs crash since all repos by default are in `/home/git/gogs-repositories`.

scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
+ 
+; Make /usr, /boot, /etc and possibly some more folders read-only.
```
bkcsoft: "possible some more folders", which? `man systemd.exec` makes no mentions of this... (also useless comment)

scripts/systemd/gogs.service
```diff
@@ -8,19 +9,43 @@ After=network.target
...
+ProtectSystem=full
+ 
+; Make our directory writable
+ReadWriteDirectories=/home/git
```
bkcsoft: You just undid `ProtectHome=true` with regards to protecting SSH-keys...

bkcsoft:
Have you actually tried this? If so, did you check that it does what it's supposed to? And what's with all the comments? o.O

LefsFlarey:
This is a systemd file and as far as I can see, this is slightly overkill.

scripts/systemd/gogs.service
LefsFlarey: Yep, he moved it.

scripts/systemd/gogs.service
LefsFlarey: Copy pasted probably.

scripts/systemd/gogs.service
LefsFlarey: I guess a work-around could be done with the default but yeah, bad idea I guess.

tehscorpion:
Hey guys, thanks for your comments. Let me explain :+1: 

`Restart=on-failure` is because if set to `Restart=always` it will be restarted regardless of its exit signal, which is not the default behaivour for most of network facing services. (At least on debian/ubuntu systems).

`+; Gogs directoryies.` sorry, that was a typo.

`Type=simple` is the default  if neither Type= nor BusName=, but ExecStart= are specified, so no need for this.

`ProtectHome=true` and `ReadWriteDirectories=/home/git` should work fine together. Otherwise we need to change it to `ProtectHome=read-only` 

`with regards to protecting SSH-keys...` i mean SSH-keys from other users.

At least on my system this service unit is working very well. And those changed were just to harden it a bit further.

LefsFlarey:
> Restart=on-failure is because if set to Restart=always it will be restarted regardless of its exit signal, which is not the default behaivour for most of network facing services. (At least on debian/ubuntu systems).

Then again, having it restart always is a safe assurance that Gogs is online no matter what, unless specified otherwise.

> +; Gogs directoryies. sorry, that was a typo.

Yes.

> ProtectHome=true and ReadWriteDirectories=/home/git should work fine together. Otherwise we need to change it to ProtectHome=read-only

This makes Gogs' git-hook capabilities almost crippling to use.

> with regards to protecting SSH-keys... i mean SSH-keys from other users.

Then again, it's public keys.

> At least on my system this service unit is working very well. And those changed were just to harden it a bit further.

Hardening is fine but hardening to this level ***should*** be made optional as the hardening mechanisms you've made pretty much crippled some functions of Gogs

> Regarding the comments, just to make it more readable. If you dont like them, just remove them.

Actually, for this PR to work, you have to remove them on your end before the PR can be merged, otherwise @Unknwon has to create an additional commit to remove your comments. What a trouble eh?

tehscorpion:
Okay,
I changed it

Unknwon:
Hi, what is the status of this PR right now?

PR #3804, opened 2016-10-24
Fix problem with es-ES locale in years translation

ogarcia:
This fix a small issue with es-ES translation.

Now the string appears broken as `%!s(int64=2) %!d(string=hace) años` but this small PR fix to `hace 2 años`


bkcsoft:
This should be done in https://crowdin.com/project/gogs :slightly_smiling_face:

Issue #3806, opened 2016-10-25
Use fullname when user is resolved


0xbaadf00d:
- Gogs version (or commit ref): 0.9.99.0903
- Git version: 2.10
- Operating system: CentOS 7 64bit
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes
  - [ ] No
  - [ ] Not relevant

## Description

It would be nice to use the user full name set on GOGS profile rather than commit _author_ field when user has been resolved from its email address. 

This could be useful to standardize different GOGS views (commit, files explorer, ...) and deal with developer who don't care about configuring _.gitconfig_ file.

In corporate environment too, we have to get the real user full name (as set by the system administrator) rather than what develop set on its _.gitconfig_ file.

![image](https://cloud.githubusercontent.com/assets/1005086/19676149/c97f3224-9a92-11e6-95c2-5122a90b264a.png)

LefsFlarey:
Perhaps having your developer set it with a simple command like `git config --add user.name {name}` would be easier than this? Though I guess it could be done.

0xbaadf00d:
@LefsFlarey  Query is already done to check if account exists to put a href (see on the enclosed picture)

![image](https://cloud.githubusercontent.com/assets/1005086/19685155/5b994f82-9abb-11e6-9e5e-939170c659b6.png)

LefsFlarey:
Alright, that seems mightily fair. I've got a fix working, hopefully it compiles. As far as I know, this issue happens on the commits table as well as the latest committer on the main page, did I miss anything else out?

LefsFlarey:
@0xbaadf00d Pushed tiny fix at #3808

0xbaadf00d:
Nice, thanks

bkcsoft:
@LefsFlarey 
> Perhaps having your developer set it with a simple command like git config --add user.name {name} would be easier than this? Though I guess it could be done.

`git commit` will barf in your face if you don't set `user.name` and `user.email` so that should be a non-issue.

Though they _can_ set `user.name=foobar` instead if they're lazy :slightly_smiling_face:

LefsFlarey:
@bkcsoft 

>git commit will barf in your face if you don't set user.name and user.email so that should be a non-issue.

I know, I was referring to him asking his developers to set a single name and not change it for the repository.

[...]

PR #3808, opened 2016-10-25
Use Gogs name if user is registered

LefsFlarey:
This addresses and solves #3806, which is an issue about standardizing the view of the committer's name as long as he/she/it is registered with the Gogs server.

### DIFF
--- a/templates/repo/view_list.tmpl
+++ b/templates/repo/view_list.tmpl
@@ -4,7 +4,7 @@
 			<th class="four wide">
 				{{if .LatestCommitUser}}
 					<img class="ui avatar image img-12" src="{{.LatestCommitUser.RelAvatarLink}}" />
-					<a href="{{AppSubUrl}}/{{.LatestCommitUser.Name}}"><strong>{{.LatestCommit.Author.Name}}</strong></a>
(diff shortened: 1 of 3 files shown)


0xbaadf00d:
I just test your branch on localhost, really nice PR. I had just write a review for an little issue with users who don't have set full name on profile page.

Thanks

templates/repo/view_list.tmpl
```diff
@@ -4,7 +4,7 @@
 			<th class="four wide">
 				{{if .LatestCommitUser}}
 					<img class="ui avatar image img-12" src="{{.LatestCommitUser.RelAvatarLink}}" />
-					<a href="{{AppSubUrl}}/{{.LatestCommitUser.Name}}"><strong>{{.LatestCommit.Author.Name}}</strong></a>
+					<a href="{{AppSubUrl}}/{{.LatestCommitUser.Name}}"><strong>{{.LatestCommitUser.FullName}}</strong></a>
```
LefsFlarey: Is it not automatically fallback on username?

0xbaadf00d: No

Issue #3760, opened 2016-10-13
MailParticipants fails


stroucki:
- Gogs version (or commit ref): 2bec8a4f1ef6b48220a0010c1482e561abd55810
- Git version: 2.1.4
- Operating system: Linux
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [x] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log gist:
  [T] mp wtf [5]: UpdateIssueUsersByMentions: database table is locked
## Description

If a comment tags a name (@stroucki), emails do not send out because of the "database table is locked" error above. This is within function MailParticipants in models/issue_comment.go after the line
mentions := markdown.FindAllMentions(cmt.Content)

Note that the error never gets reported; I made it visible by inserting a log.Trace.

This fails for extant users; if a random text string is tagged, a mail is sent.

stroucki:
This is the xorm log for the issue. The error results from the last update statement. It is not clear to me what is locking the table.

```
[xorm] [info]  2016/10/26 23:30:23.088011 [sql] BEGIN TRANSACTION
[xorm] [info]  2016/10/26 23:30:23.088041 [sql] INSERT INTO `comment` (`type`,`poster_id`,`issue_id`,`commit_id`,`line`,`content`,`created_unix`,`updated_unix`,`commit_sha`) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?) [args] [0 4 5 0 0 @stroucki: x 1477539023 1477539023 ]
[xorm] [info]  2016/10/26 23:30:23.088209 [sql] UPDATE `issue` SET num_comments=num_comments+1 WHERE id=? [args] [5]
[xorm] [info]  2016/10/26 23:30:23.088278 [sql] SELECT `id`, `user_id`, `repo_id` FROM `watch` WHERE `repo_id`=? [args] [43]
[xorm] [info]  2016/10/26 23:30:23.088418 [sql] INSERT INTO `action` (`user_id`,`op_type`,`act_user_id`,`act_user_name`,`repo_id`,`repo_user_name`,`repo_name`,`ref_name`,`is_private`,`content`,`created_unix`) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) [args] [4 10 4 test 43 test test  false 2|@stroucki: x 1477539023]
[...]

PR #3822, opened 2016-10-27
Fix database write context interleaving bug

stroucki:
UpdateIssueUsersByMentions was calling database write operations while
a transaction session was in progress. MailParticipants was failing
silently because of the SQLITE_LOCKED error. Make sure failures in
MailParticipants enter the log, and pass on the transaction context.
Addresses #3760.

issue: let caller pass in database context, and use it
issue_comment: obtain database context to pass to UpdateIssueMentions
issue_comment: log any error from call to MailParticipants
issue_mail: pass on database context to UpdateIssueMentions


stroucki:
That output looks like a problem with Travis.

Unknwon:
Thanks!

PR #3827, opened 2016-10-28
Check for zero length passwords in LDAP module.

ruanda:
According to https://tools.ietf.org/search/rfc4513#section-5.1.2 LDAP client should always check before bind whether a password is an empty value. There are at least one LDAP implementation which does not return error if you try to bind with DN set and empty password - AD.


Unknwon:
Thanks!

PR #3852, opened 2016-11-04
Fix typos

blindpirate:
Some typos.


strk:
Would you consider porting this over to https://github.com/go-gitea/gitea ? Development is continuing there

Unknwon:
Thanks!

Issue #3845, opened 2016-11-02
Ordering organizations and users by name


0xbaadf00d:
- Gogs version (or commit ref): 0.9.99.0915
- Git version: 2.10.1
- Operating system: CentOS 64bit
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (all pages)
  - [ ] No
  - [ ] Not relevant

## Description

It could be more user friendly and make search more easy if organizations and user accounts are ordered by name rather than last update datetime.

PR #3853, opened 2016-11-04
Feature #3845: Ordering organizations and user accounts by name

0xbaadf00d:
Relative to the issue #3845, this Pull Request add modifications to ordering by name Organizations and User Accounts.

It's more user friendly and make search more easy when organizations and user accounts are ordered by name rather than last update datetime.

I added new function `GetUsersByIDs` to decrease the number of SQL queries by using IN predicate rather than just running 1 SQL query per users.

### DIFF
--- a/templates/admin/user/list.tmpl
+++ b/templates/admin/user/list.tmpl
@@ -18,7 +18,6 @@
 					<table class="ui very basic striped table">
 						<thead>
 							<tr>
-								<th>ID</th>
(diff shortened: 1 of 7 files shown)


templates/admin/user/list.tmpl
```diff
@@ -18,7 +18,6 @@
 					<table class="ui very basic striped table">
 						<thead>
 							<tr>
-								<th>ID</th>
```
Unknwon: Why remove these IDs?

PR #3854, opened 2016-11-04
Feature #3845: Ordering organizations and users by name

0xbaadf00d:
Relative to the issue #3845, this Pull Request add modifications to ordering by name Organizations and User Accounts.

It's more user friendly and make search more easy when organizations and user accounts are ordered by name rather than last update datetime.

I added new function `GetUsersByIDs` to decrease the number of SQL queries by using IN predicate rather than just running 1 SQL query per users.

### DIFF
--- a/models/org.go
+++ b/models/org.go
@@ -50,7 +50,7 @@ func (org *User) GetOwnerTeam() (*Team, error) {
 }
 
 func (org *User) getTeams(e Engine) error {
-	return e.Where("org_id=?", org.ID).Find(&org.Teams)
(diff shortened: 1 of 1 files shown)


bkcsoft:
This should go in :develop :/

models/org.go
```diff
@@ -50,7 +50,7 @@ func (org *User) GetOwnerTeam() (*Team, error) {
 }
 
 func (org *User) getTeams(e Engine) error {
-	return e.Where("org_id=?", org.ID).Find(&org.Teams)
+	return e.Where("org_id=?", org.ID).OrderBy("CASE WHEN name LIKE '" + OWNER_TEAM + "' THEN '' ELSE name END").Find(&org.Teams)
```
Unknwon: Does this SQL only work on MySQL?

PR #3876, opened 2016-11-12
use correct import for libravatar

smcquay:

This fixes building on my machine.


Unknwon:
Thanks, was fixed sometime ago already.

PR #3885, opened 2016-11-16
Fix Typo Errors

seongahjo:
Fix Typo Errors


strk:
Would you consider moving this to https://github.com/go-gitea, if still needed there ?

Unknwon:
Thanks!

Issue #3814, opened 2016-10-26
Inactive users receive email notifications


jhit:
Gogs Version: 0.9.99.0903
## Description

If I deactivate a former user of our gogs instance he still receives notifications by email. 
I think it would preferable to not send notifications to inactive users.

Unknwon:
This is claimed to be fixed by merging #3903, please help test on `develop` branch.

PR #3903, opened 2016-11-24
Issue #3814 - Inactive users receive email notifications

renatoaquino:
Just a small patch for #3814.
Not sure if it's enough to cover all unwanted mailing situations.


Unknwon:
Thanks!

PR #3906, opened 2016-11-25
#3791 update LDAP username check

### LINKED ISSUE #3791: LDAP login with spaces in CN
Hello, 

I have a question is it possible with ldap to login with CN=  Geoffry Van den Eede
I get the following error when I try that.
Or is there another solution so I can login with the sAMAccountName attribute? I can't get that to work :(

2016/10/20 15:36:26 [T] LDAP will bind directly via UserDN template: CN=%s,OU=DVIT Users,DC=DVIT,DC=LOCAL
2016/10/20 15:36:26 [D] 'Geoffry Van den Eede' contains invalid DN characters. Aborting.
2016/10/20 15:36:26 [W] Failed to login 'Geoffry Van den Eede' via 'login': user does not exist [uid: 0, name: Geoffry Van den Eede]

Thanks already!

solo12zw74:
TGis request try to fix #3791 
As https://msdn.microsoft.com/en-us/library/aa366101(v=vs.85).aspx describe spaces should not be in start or at the end of username but they can be inside the username. So please check my solution for it.

### DIFF
--- a/modules/auth/ldap/ldap.go
+++ b/modules/auth/ldap/ldap.go
@@ -59,8 +59,8 @@ func (ls *Source) sanitizedUserQuery(username string) (string, bool) {
 
 func (ls *Source) sanitizedUserDN(username string) (string, bool) {
 	// See http://tools.ietf.org/search/rfc4514: "special characters"
-	badCharacters := "\x00()*\\,='\"#+;<> "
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

PR #3908, opened 2016-11-26
Add UI config to toggle whether user email shows up in Explore Users page

tanob:
In some scenarios it's undesirable to show the user's email in the Explore page, even though currently Gogs shows only for signed-in users. 

In my case i'm planning to use Gogs only as a nice interface to Git, the users won't benefit from the Explore features (it'd actually be nicer to disable the Explore section for non-admin users).


Unknwon:
Thanks your PR!

But if you're signed in user, you can get email from a lot places in current Gogs version. Why does explore matter?

tanob:
@Unknwon I don't recall seeing the email exposed in other places for signed-in users, maybe they're exposed but the user would need to guess/know the other user's username to build the URL to that page?

Do you see a different way to implement this scenario where we disable the social features of Gogs?

Unknwon:
@tanob User profile page (e.g. `/unknwon`) shows email for signed-in users too. And in relevant APis.

tanob:
@Unknwon got it but users would need to guess the other user's username, right? If usernames are UUIDs then it becomes more difficult.

Unknwon:
They don't need to guess, just click on links on explore page.

tanob:
You're right, I don't think this PR helps achieving what I was looking for.

PR #3952, opened 2016-12-12
Don't use custom PBKDF2 function

leonklingele:
Don't know why we'd want to duplicate something which already exists in `golang.org/x/crypto/pbkdf2`

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -32,6 +32,8 @@ import (
 	"github.com/gogits/gogs/modules/log"
 	"github.com/gogits/gogs/modules/markdown"
 	"github.com/gogits/gogs/modules/setting"
+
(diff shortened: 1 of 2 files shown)


lunny:
Hi, could you send a PR to github.com/go-gitea/gitea too?

bkcsoft:
@leonklingele because at the time this was implemented, that package didn't exist officially :)

leonklingele:
@lunny why not apply the patch yourself? ;)

lunny:
@leonklingele OK. I will do it and reference to this issue.

tboerger:
> @lunny why not apply the patch yourself? ;)

Because it's a community project ;)

Unknwon:
Thanks!

PR #3953, opened 2016-12-12
Fix random string generator

leonklingele:
This PR fixes the following issues:

- The random string generator was [modulo-biased](https://stackoverflow.com/a/10984975).
- The random string generator silently ignored any errors returned by `rand.Read`. In the worst case, `GetRandomString` would simply return an empty string.


Unknwon:
Thanks!

PR #3967, opened 2016-12-19
Support for video files using the HTML5 video tag

nmandery:
This pull requests adds support for video files using the HTML5 video tag. This allows viewing videos files directly in the webinterface - essentially the same way PDF files are supported.

Video files are not ideal to be managed using git, but may be stored in repositories as part of a projects documentation. This is the case in a project I am currently working on, and it helps a lot being able to preview these files directly in gogs instead of having to download and to open them in a video player. With video support in most of the current browsers (http://caniuse.com/#search=video) this seems like a helpful feature to be implemented in gogs.

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -427,6 +427,7 @@ file_history = History
 file_view_raw = View Raw
 file_permalink = Permalink
 file_too_large = This file is too large to be shown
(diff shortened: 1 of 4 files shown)


lunny:
hi, could you send PR to Gitea also? Gitea LFS with video will be useful!

nmandery:
@lunny Sure, the PR to gitea is at go-gitea/gitea#418

Unknwon:
Thanks!

PR #3969, opened 2016-12-19
Add support for using "libravatar" as the GravatarSource

strk:

1. Please make sure you are targeting the `develop` branch.
2. Please read contributing guidelines:
4. ... if it is not related to any particular issues, explain why we should not reject your pull request.

**You MUST delete above content including this line before posting; too lazy to take this action considered invalid pull request.**

Just to make it easier for administrator to configure libre avatar,
as it is done for "duoshuo" and "gravatar"


Unknwon:
Thanks!

Issue #3919, opened 2016-11-30
Wrong anchors for non-latin symbols in wiki pages


tetafro:
- Gogs version (or commit ref): 0.9.100.1119
- Git version: 2.6.6
- Operating system: Ubuntu 16.04 + Docker 1.12.3 + Alpine Linux 3.3 (inside container)
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (https://try.gogs.io/tetafro/test/wiki/Home)
  - [ ] No
  - [ ] Not relevant
- Log gist: -

## Description

Wrong anchors for non-latin symbols in wiki pages: when head (starts with # symbols) contain non-latin symbols, like cyrillic for example, anchor link for this head will not contain any of these symbols.

Unknwon:
This should be fixed by https://github.com/gogits/gogs/pull/3981.

Issue #3843, opened 2016-11-02
Imcomplete anchors for non-latin headings in markdown


nook-ru:
- Gogs version (or commit ref): 0.9.99.0915
- Git version: not relevant
- Operating system: 
- Database:
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (https://try.gogs.io/nook_ru/test)
  - [ ] No
  - [ ] Not relevant
- Log gist: none

## Description

Each heading has an anchor pointing to it, i.e. https://try.gogs.io/nook_ru/test#test-heading

Unfortunately all non-latin characters are stripped from the heading anchor and replaced with minus sign. So, for example, for cyrillic heading *Еще один заголовок* anchor will look like `#--`. **GitHub** correctly deals with such headings, you can compare same markdown:

* [GitHub](https://gist.github.com/nook-ru/900f4c2d85b0b053f5b9618cd8439743) renders anchor as `#Еще-один-заголовок`
* [Gogs](https://try.gogs.io/nook_ru/test) gives `#--` for the same heading

Unknwon:
This should be fixed by https://github.com/gogits/gogs/pull/3981.

PR #3981, opened 2016-12-26
Fix wrong anchors for non-latin headings

Joel-Liang:
Change Javascript regular expression to match non-latin characters
The regex comes from here: http://stackoverflow.com/questions/150033/regular-expression-to-match-non-english-characters#comment22322603_150078

And this patch should fixed these two issues: #3919 #3843


Unknwon:
Thanks!

PR #3996, opened 2016-12-28
Verify repository is not bare before retrieving blob (Fixes #3992).

### LINKED ISSUE #3992: API: nil pointer dereference when retrieving content from unintialized repository
- Gogs version: v0.9.113
- Git version: 114c179
- Operating system: Alpine Linux 3.4
- Database:
  - [ ] PostgreSQL
  - [ ] MySQL
  - [x] SQLite
  - [x] Yes
  - [ ] No
  - [ ] Not relevant
- Log gist:
Stracktrace: https://gist.github.com/cybe/603458382308f3bd27ee98b95641feb9
## Description
Retrieving a blob from an unintialized repository results in a nil pointer dereference, as shown in the Stacktrace. The corresponding API call looks like this:

`curl -v -H "Authorization: token 30af10dae468fa01d658c2b7962a204880fb6751" https://try.gogs.io/api/v1/repos/cybe/test/raw/master/Dockerfile`
→ HTTP/1.1 500 Internal Server Error

The relevant line in the source code: https://github.com/gogits/gogs/blob/114c179e5a50e3313f7a5894100693805e64e440/routers/api/v1/repo/file.go#L22

cybe:
This pull request adds a simple check for bare repositories before retrieving raw files. If the repository is indeed bare (There are no branches or commits), just return a 404.

This fixes #3992.

### DIFF
--- a/routers/api/v1/repo/file.go
+++ b/routers/api/v1/repo/file.go
@@ -19,6 +19,11 @@ func GetRawFile(ctx *context.APIContext) {
 		return
 	}
 
+	if ctx.Repo.Repository.IsBare {
+		ctx.Status(404)
+		return
+	}
+
 	blob, err := ctx.Repo.Commit.GetBlobByPath(ctx.Repo.TreePath)
 	if err != nil {
 		if git.IsErrNotExist(err) {


Unknwon:
Thanks!

PR #3997, opened 2016-12-29
Enable federated avatars by default

strk:
Enabling federated avatars by default makes it easier to be nice with users who have their own avatar distribution setting.


bkcsoft:
rebase on develop and change the branch target ;)

strk:
Done (damn `hub` commandline, keeps failing)

strk:
So done right now, and ready to merge. What do you think @Unknwon  ?

Unknwon:
Thanks!

PR #4010, opened 2017-01-04
Remove call to set GOMAXPROCS

aaronjwood:
Since Go 1.5 GOMAXPROCS is already set to the number of CPUs in the system. This call will also be removed from Go in the near future (part of their plans once they feel the scheduler is good enough) so I thought it'd make sense to remove it here.


Unknwon:
Thanks!

PR #4011, opened 2017-01-04
Produce a position independent executable

aaronjwood:
This will output a position independent executable. A few benefits to this:

- While normally used as a hardening technique to better take advantage of ASLR (thus making exploits harder), this doesn't really have the same effect on Go binaries since everything is bounds checked and (hopefully) has no dangling pointers underneath. Still, this can help if any code calls out to unsafe code (C/C++) and can potentially help guard against situations where there may be a hole in the Go runtime (probably never the case).
- A lot of distros are now requiring packages to be shipped as PIE executables.

Note that this setting won't work on Windows or OS X so I'm assuming that the Makefile will never be used on Windows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ LESS_FILES := $(wildcard public/less/gogs.less public/less/_*.less)
 GENERATED  := modules/bindata/bindata.go public/css/gogs.css
 
 TAGS = ""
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks!

PR #4017, opened 2017-01-05
Enable federated avatars by default

strk:


strk:
Oops, was already in #3997 (done differently)

PR #4018, opened 2017-01-06
Dockerfile for aarch64 (odroid c2 & pine64)

atzoum:
Minimal changes compared to `Dockerfile.rpi`, you may find and test an already built docker image
 [here](https://hub.docker.com/r/atzoum/aarch64-gogs/).


Unknwon:
Thanks!

PR #4064, opened 2017-01-28
Safe compare password (timing attack)

denji:
The time taken is a function of the length of the slices and is independent of the contents.
* https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare

### DIFF
--- a/models/user.go
+++ b/models/user.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"container/list"
 	"crypto/sha256"
+	"crypto/subtle"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for the PR, but please read https://github.com/gogits/gogs/blob/master/.github/PULL_REQUEST_TEMPLATE.md

leonklingele:
There's no timing attack here. All an attacker can learn if you're using non constant-time `==` is that _some_ hash of his password (which he doesn't even get to see) is different from the one stored at byte x.
However I too prefer to use constant time for every equals-check which includes some kind of crypto, just to always be on the right side. :)

Issue #3715, opened 2016-09-30
jupyter notebook support


oroszl:
One of the neatest features of github is its ability to render jupyter notebooks.
This sort of feature has also came up as a request in the gitlab community.
So I thought I might bring it to gogs as well. 
Github uses nbconvert as a server side solution for rendering the notebooks but there have been some client side proposals for the same, see for example [here](https://github.com/kokes/nbviewer.js) and [here](https://github.com/nteract/notebook-preview).
I think having this feature would greatly increase the appeal of gogs for certain type of users.

kellpossible:
I've got a few notebooks in one of my repos, might check out how hard this would be to do

bkcsoft:
https://github.com/fabian-z/gopherlab would probably be a good starting-point if anyone wanna have a crack at this

oroszl:
@bkcsoft I dont think your link is relevant here.. my suggestion was to have the capability in gogs to display jupyter notebooks not to have a go kernel for jupyter. The stuff you see in the link you posted is just a display of github's capability to rendner notebooks for a go kernel.

tgy:
**I would also be greatly interested in seeing this feature in Gogs.**

I wonder how it's implemented on GitHub but I don't think it is too complicated.

If I had to implement it I would use a program that converts notebooks to HTML server-side like [nbconvert](https://nbconvert.readthedocs.io/en/latest/) and include this HTML in the page showing the notebook file. The problem with this approach is that there is a need for `jupyter` to be installed on the server (I'm not entirely sure but if the notebook's version and the version installed server-side differ there might be a problem). What would be better is to convert the notebook directly on the client's browser using JavaScript but I'm not convinced such a tool exists and it doesn't solve the potential version mismatch issue.

prinsherbert:
I created a PR for this feature, regarding **Python** notebooks. Probably there are some easy fixes to also make it work for other languages, but I can not test this as I have only experience using i**py**nb's.

https://github.com/gogits/gogs/pull/4067

oroszl:
As far as I know there is not a lot of difference to other kernels in the ipynb format, if you work with other kernels they will use the same format, with maybe a minimal overhead in metadata specifying the language..

prinsherbert:
@oroszl Indeed, there isn't a lot of difference to the best of my knowledge. However, styling can go wrong and possibly also other stuff that I can't test thoroughly because I'm not using non-python notebooks.

[...]

PR #4067, opened 2017-01-30
Ipython notebook support

prinsherbert:
I added ipython notebook renderering, unfortunately I could not find any Go-libraries that do this, hence I implemented it via Javascript.

This would in part solve this issue https://github.com/gogits/gogs/issues/3715

Not completely, because I only created it for python notebooks, I have no Idea how it would and should work for julia, R and possibly other programming language notebooks


prinsherbert:
Simple example ipython notebook:

ipynb-gogs

Unknwon:
Thanks for the PR! And the preview looks incredibly good.

However, the change set is so huge, it is not easy for merging. Instead of merging, I may take this PR as a source when implement this feature when Gogs has plugin system, that's much easier to maintain.

prinsherbert:
> Thanks for the PR! And the preview looks incredibly good.

Thank you! And thank you for your fast response. 

> However, the change set is so huge, it is not easy for merging. Instead of merging, I may take this PR as a source when implement this feature when Gogs has plugin system, that's much easier to maintain.

@Unknwon Some of the library-files might be superfluous, I simply downloaded the github-releases. Moreover, the changes should only be in the `public/plugins`-folders, which should not prove troublesome when merging.

Another, possibly less relevant note, I like to collaborate with notebooks in git, because they allow me to share results of scientific research with others easily (including plots and such) and help me document experiments. This feature in gogs would help me a great deal.

Unknwon:
Thanks for the response!

To help review, would you mind put actual code and library dependency in separate commits?

prinsherbert:
> To help review, would you mind put actual code and library dependency in separate commits?

Yes, here you are: https://github.com/gogits/gogs/pull/4068

oroszl:
Hey @prinsherbert great effort! Can you discribe a bit more in detail how you do the notebook rendnering ?
Specifically are you using nbconvert ?

prinsherbert:
@CarlosGtrz No, I use notebookjs, a javascript library which transforms the notebook json into HTML, then I created some CSS to style

PR #4068, opened 2017-01-30
Ipython notebook support separate libraries

prinsherbert:
See https://github.com/gogits/gogs/pull/4067


Unknwon:
Thanks. Have you resync with latest `develop` code? Because I see many deletions are newly added in Gogs codebase.

Unknwon:
Hmm, after rebased from `develop`, I think you might need recommit these two commits so critical changes are combined and able to review.

prinsherbert:
It should be up to date with the newest develop commit.

Most of those deletes are in `gogs.css`, which my lessc did, it created a slightly different css at some lines, I have no idea why, I simply used `make less`.

Unknwon:
Many deletions in `.less` are newly added as well...

prinsherbert:
Yes, and they are, right?

Unknwon:
No,

For example:

- Deleted: https://github.com/gogits/gogs/pull/4068/commits/7cf86c5fc52521d1304c83e35dbc2a01a907f471#diff-4c18d74f0d0687b8503915779673f3ebL254
- Added: https://github.com/gogits/gogs/blob/master/public/less/_repository.less#L254

PR #4078, opened 2017-02-01
Check deploy keys when Gogs is run with Service.RequireSignInView=true

psaavedra:
Check if the deploy key can access to a repository. A deploy key
doesn't represent an gogs user, so in a site with Service.RequireSignInView
activated we should give read access only in the repositories where
this deploy key is in use. In other case, a deploy service or system
using an active deploy key can get read access to all the repositories
in a Gogs service.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -259,7 +259,8 @@ func runServ(c *cli.Context) error {
 					user.Name, requestedMode, repoPath)
 			}
 		}
-	} else  { // if public and read
+	} else  {
+                // if public and read
                 // Check if the key can access to the repository in case of it is a deploy key (a deploy keys != user key).
                 // A deploy key doesn't represent a signed in user, so in a site with Service.RequireSignInView activated
                 // we should give read access only in repositories where this deploy key is in use. In other case, a server
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -260,15 +260,15 @@ func runServ(c *cli.Context) error {
 			}
 		}
 	} else  {
-                // if public and read ...
-                // Check if the key can access to the repository in case of it is a deploy key (a deploy keys != user key).
-                // A deploy key doesn't represent a signed in user, so in a site with Service.RequireSignInView activated
-                // we should give read access only in repositories where this deploy key is in use. In other case, a server
-                // or system  using an active deploy key can get read access to all the repositories in a Gogs service.
+		// if public and read ...
+		// Check if the key can access to the repository in case of it is a deploy key (a deploy keys != user key).
+		// A deploy key doesn't represent a signed in user, so in a site with Service.RequireSignInView activated
+		// we should give read access only in repositories where this deploy key is in use. In other case, a server
+		// or system  using an active deploy key can get read access to all the repositories in a Gogs service.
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks for t he PR!

I think we should reuse [this part of code](https://github.com/gogits/gogs/blob/master/cmd/serve.go#L215-L233), maybe as a function, to reduce duplicated code blocks.

psaavedra:
Code refactoring done

Unknwon:
Please response to review comments.

psaavedra:
I don't see comments for the latest push. Anyway, the code now uses 2 functions wich encapsulate duplicated logic wrote in the previous one. Feel free to make comments in this patch now.

Unknwon:
Oh my bad, it is confusing, I thought was posted but actually I have to click on `**ubmit Review**...

cmd/serve.go
```diff
@@ -41,6 +41,7 @@ var CmdServ = cli.Command{
 
 func setup(logPath string) {
 	setting.NewContext()
+	setting.NewService()
```
psaavedra: in order to load in `struct Service` the values from the configuration file. In particular, I was looking for a way to get `RequireSignInView`. 

Looks like `newService()` is the function which load this information but, I want load just this values not all the initialization done by `NewServices()` so I think, get a public function just for this  is the right way to get it.

cmd/serve.go
```diff
@@ -250,7 +259,15 @@ func runServ(c *cli.Context) error {
 					user.Name, requestedMode, repoPath)
 			}
 		}
-	}
+	} else  { // if public and read
```
Unknwon: Should be a easy fix for this one..?

cmd/serve.go
```diff
@@ -250,7 +259,16 @@ func runServ(c *cli.Context) error {
 					user.Name, requestedMode, repoPath)
 			}
 		}
-	}
+	} else  {
+                // if public and read
```
Unknwon: Doesn't seem like code is correctly formatted?

psaavedra: Why not, I've followed the same style than the others if-else structures?

Unknwon: It's very obvious to tell that comment and code below don't have same indentation.

cmd/serve.go
```diff
@@ -250,7 +259,16 @@ func runServ(c *cli.Context) error {
 					user.Name, requestedMode, repoPath)
 			}
 		}
-	}
+	} else  {
+                // if public and read ...
+                // Check if the key can access to the repository in case of it is a deploy key (a deploy keys != user key).
+                // A deploy key doesn't represent a signed in user, so in a site with Service.RequireSignInView activated
+                // we should give read access only in repositories where this deploy key is in use. In other case, a server
+                // or system  using an active deploy key can get read access to all the repositories in a Gogs service.
```
Unknwon: It is very obvious to tell the comments is not having same indentation as code below...

psaavedra: I see. My bad. My editor is replacing tabs by spaces.

psaavedra: formatted and pushed again

Unknwon:
Thanks!

Issue #4084, opened 2017-02-04
LibreJS compatibility


jorgesumle:
https://freepo.st/post/imtju3v7bj#comments

It would be great if Gogs could be compatible with [LibreJS](https://www.gnu.org/software/librejs/). I've started working in making this possible for NotABug.org, but it would be better if I could do that to this repo. Should I make a pull request with this feature here?

It would mean having to distribute not just the minified JavaScript files, but also the source code (for examble, semantic-2.2.1.min.js **and semantic.js**) and creating a link in the footer template pointing to the JavaScript license information table. It could be a hidden link (not visible).

> `<a hidden href="/librejs" data-jslicense="1">JavaScript licenses</a>`

The JavaScript license table would be something like this (assuming it would be in the `gogs/templates` directory). We could call that file `librejs.tmpl`:

```
{{template "base/head" .}}

<table id="jslicense-labels1">
  <tr>
    <td><a href="/js/libs/jquery.are-you-sure.js">jquery.are-you-sure.js</a></td>
    <td>
      <a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a>
      <br>
      <a href="http://www.gnu.org/licenses/gpl-2.0.html">GNU-GPL-2.0-or-later</a>
    </td>
    <td><a href="/js/libs/jquery.are-you-sure.js">jquery.are-you-sure.js</a></td>
  </tr>
  <tr>
    <td><a href="/js/semantic-2.2.1.min.js">semantic-2.2.1.min.js</a></td>
    <td><a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a></td>
    <td><a href="/js/semantic-2.2.1.js">semantic-2.2.1.js</a></td>
  </tr>
  <tr>
    <td><a href="/js/gogs.js">gogs.js</a></td>
    <td><a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a></td>
    <td><a href="/js/gogs.js">gogs.js</a></td>
  </tr>
  <tr>
    <td><a href="/js/libs/">gogs.js</a></td>
    <td><a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a></td>
    <td><a href="/js/gogs.js">gogs.js</a></td>
  </tr>
  <tr>
    <td><a href="/plugins/highlight-9.6.0/highlight.pack.min.js">highlight.pack.min.js</a></td>
[...]

PR #4070, opened 2017-01-31
Ipython notebook support

prinsherbert:
I'm sorry for making a mess of the conversation (this being the 3rd PR for the same feature)

I was more careful merging with newer stuff in the develop branch, I think I have done it right this time. Please let me know if anything is still off.

- https://github.com/gogits/gogs/pull/4068
- https://github.com/gogits/gogs/pull/4067

PR #4092, opened 2017-02-06
Add LibreJS support

jorgesumle:
https://github.com/gogits/gogs/issues/4084 I've finished the librejs template file, but I need some help with routes. 

I'm not sure how do templates routes work and I don't want to make something wrong. I'm trying to make gogsbaseurl.com/librejs to contain the LibreJS page, but can make it work and I'm afraid to break something. Once that is done, I'll have no problem adding a link to the footer template. Could one active/experienced programmer in this project help me?

Sorry for my ignorance, I never programmed in Go and I'm having a hard time in understanding how the code works.

### DIFF
--- a/public/assets/librejs/librejs.html
+++ b/public/assets/librejs/librejs.html
@@ -1,0 +1,76 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>JavaScript license information</title>
+  </head>
+  <body>
+    <table id="jslicense-labels1">
+      <tr>
+        <td><a href="/js/libs/jquery.are-you-sure.js">jquery.are-you-sure.js</a></td>
+        <td>
+          <a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a>
+          <br>
+          <a href="http://www.gnu.org/licenses/gpl-2.0.html">GNU-GPL-2.0-or-later</a>
+        </td>
+        <td><a href="/js/libs/jquery.are-you-sure.js">jquery.are-you-sure.js</a></td>
+      </tr>
+      <tr>
+        <td><a href="/js/semantic-2.2.7.min.js">semantic-2.2.1.min.js</a></td>
+        <td><a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a></td>
+        <td><a href="https://github.com/Semantic-Org/Semantic-UI/archive/2.2.1.tar.gz">semantic-UI-2.2.1.tar.gz</a></td>
+      </tr>
+      <tr>
+        <td><a href="/js/gogs.js">gogs.js</a></td>
+        <td><a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a></td>
+        <td><a href="/js/gogs.js">gogs.js</a></td>
+      </tr>
+      <tr>
+        <td><a href="/js/libs/clipboard-1.5.9.min.js">clipboard-1.5.9.min.js</a></td>
+        <td><a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a></td>
+        <td><a href="https://github.com/zenorocha/clipboard.js/archive/v1.5.9.tar.gz">clipboard-1.5.9.tar.gz</a></td>
+      </tr>
+      <tr>
+        <td><a href="/js/libs/emojify-1.1.0.min.js">emojify-1.1.0.min.js</a></td>
+        <td><a href="http://www.freebsd.org/copyright/freebsd-license.html">Expat</a></td>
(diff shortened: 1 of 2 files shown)


Unknwon:
Does the URL has to be `gogsbaseurl.com/librejs`?

jorgesumle:
No, it could be in gobaseurl.com/otherfolder/librejs or in another place, but it should always be in the gobaseurl.com website.

Unknwon:
Hmm, if it's not a fixed path, how does LiberJS know where to find this page?

jorgesumle:
In the [footer.tmpl](https://github.com/gogits/gogs/blob/master/templates/base/footer.tmpl) template, you must create a link pointing to that URL and also use `data-jslicense="1"`. I didn't do that yet, because first I want to make the URL to work.

    <a hidden href="/librejs" data-jslicense="1">JavaScript licenses</a>

Unknwon:
I see. So for the HTML page you're trying to put, I think it is purely static and only for the purpose of LiberJS? 

Then we could put it under `public/assets/liberjs/liberjs.html`, and make it a self-contained HTML page. To access that page by **gobaseurl.com/assets/liberjs/liberjs.html**. Sounds good?

strk:
@jorgesumle I love this PR: would you mind sending a similar one for https://github.com/go-gitea/gitea ?

jorgesumle:
Done. Thanks for your help @Unknwon. I'll watch this repo and update the LibreJS table every time a new JavaScript library is updated or added.

Apart from the changes you suggested, I changed the link of the source file to the tar.gz file because I was told that that was the correct approach in the LibreJS mailing list: http://lists.gnu.org/archive/html/help-librejs/2017-02/msg00003.html.
bildschirmfoto vom 2017-02-07 22-09-04
bildschirmfoto vom 2017-02-07 22-09-37

Unknwon:
Looks good!

> I'll watch this repo and update the LibreJS table every time a new JavaScript library is updated or added.

You might have some work soon regarding this PR: https://github.com/gogits/gogs/pull/4070 It's going to be merged with two new JS libs.

jorgesumle:
> @jorgesumle I love this PR: would you mind sending a similar one for https://github.com/go-gitea/gitea ?
I don't think I have time right now. I would have to check its JavaScript files, install Gitea and test it. I made this contribution because I was a NotABug user and I use Gogs every day, but it's very tedious to make a pull request like this.
> You might have some work soon regarding this PR: #4070 It's going to be merged with two new JS libs.
No problem with that. :)

Unknwon:
FYI, the #4070 is merged, it's good to merge that PR before this one so you only need to verify your work once at this point.

jorgesumle:
OK, I added `/plugins/marked-0.3.6/marked.min.js` and `/plugins/notebookjs-0.2.6/notebook.min.js` from #4070 to the table of LibreJS. I used `git push -f` so that there is only one clean commit.

Unknwon:
>I used git push -f so that there is only one clean commit.

Actually we don't need to do that anymore because GitHub allows squash commits while merging.

Another question, since we're now link the license page for each JS lib, do we need to keep the license file (e.g. https://github.com/gogits/gogs/tree/master/public/plugins/marked-0.3.6) in our codebase anymore?

jorgesumle:
No. Here is an example from LibreJS documentation:
  ```
  <tr>
        <td><a href="all.min.js">all.min.js</a></td>
        <td>
            <a href="http://www.gnu.org/licenses/gpl-3.0.html"
                >GPL-3.0+</a>
            <br />
            <a href="http://www.apache.org/licenses/LICENSE-2.0"
                >Apache-2.0</a>
        </td>
        <td>
            <a href="gpl-script.js">gpl-script.js</a>
            <br />
            <a href="apache-script.js">apache-script.js</a>
        </td>
    </tr>
```
As you can see the license links are external. However, I chose to link to the license file directly because it's inside Gogs (/plugins/marked-0.3.6/LICENSE). There is an example of that also in the LibreJS documentation: 

> More information on JavaScript Web Labels is detailed here: https://www.gnu.org/software/librejs/free-your-javascript.html#step3 and here: https://www.gnu.org/licenses/javascript-labels.html.

Unknwon:
Thanks for the explanation!

In fact I would like to delete license files of JS libs from Gogs codebase and point to its license in **liberjs.html** file.

I can do that after merge the PR, or if you would like to help a bit and do it within the PR.

jorgesumle:
Can you please do that?

Unknwon:
OK, I'll do it. Thanks again for the PR!

Unknwon:
Just realized I keep typing **liberjs** where should be **librejs**... Anyway, it's now online on https://try.gogs.io/assets/librejs/librejs.html

alexgleason:
Looks like a few scripts have been updated now but the Javascript licenses page no longer reflects that. I wonder if there's a way to automate this process during build? Gogs is broken in LibreJS.

Unknwon:
Would love to see another PR to fix this... :trollface:

Issue #4107, opened 2017-02-10
SSH cipher support


spacetourist:
Hi All, I'm considering adding support for configurable SSH ciphers into the Gogs built-in server. Having reviewed the existing code it looks to be a simple change.

The reason for this request is that I am running Gogs in a PCI-DSS environment and for my needs I would like to disable the weak ciphers arcfour256 and arcfour128.

I plan to modify the SSL module code in Gogs to check the config for a new settings (perhaps SSH_SERVER_CIPHERS) and where it exists I would like to implement it when ssh.ServerConfig is instantiated in ssh.go.

Here is the default cipher support list which we are currently falling back to:

**golang.org/x/crptyo/ssh/common.go**
```
var supportedCiphers = []string{
	"aes128-ctr", "aes192-ctr", "aes256-ctr",
	"aes128-gcm@openssh.com",
	"arcfour256", "arcfour128",
}
```

Please let me know if the team would be happy to consider this as a pull request once implemented? Thanks

spacetourist:
#4109

Unknwon:
Close by https://github.com/gogits/gogs/pull/4109.

PR #4109, opened 2017-02-10
Configurable SSH cipher suite

spacetourist:
Update provides user with option to configure the cipher suite used for the built in SSH server

### DIFF
--- a/conf/app.ini
+++ b/conf/app.ini
@@ -116,6 +116,8 @@ SSH_LISTEN_HOST = 0.0.0.0
 SSH_LISTEN_PORT = %(SSH_PORT)s
 ; Root path of SSH directory, default is '~/.ssh', but you have to use '/home/git/.ssh'.
 SSH_ROOT_PATH =
+; Choose the ciphers to support for SSH connections
+SSH_SERVER_CIPHERS = aes128-ctr, aes192-ctr, aes256-ctr, aes128-gcm@openssh.com, arcfour256, arcfour128
 ; Directory to create temporary files when test publick key using ssh-keygen,
 ; default is system temporary directory.
 SSH_KEY_TEST_PATH =
--- a/modules/setting/setting.go
+++ b/modules/setting/setting.go
@@ -425,8 +425,8 @@ func NewContext() {
 		LandingPageURL = LANDING_PAGE_HOME
 	}
 
-	SSH.ServerCiphers = sec.Key("SSH_SERVER_CIPHERS").Strings(",")
 	SSH.RootPath = path.Join(homeDir, ".ssh")
+	SSH.ServerCiphers = sec.Key("SSH_SERVER_CIPHERS").Strings(",")
 	SSH.KeyTestPath = os.TempDir()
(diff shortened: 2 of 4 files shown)


modules/setting/setting.go
```diff
@@ -424,6 +425,7 @@ func NewContext() {
 		LandingPageURL = LANDING_PAGE_HOME
 	}
 
+	SSH.ServerCiphers = sec.Key("SSH_SERVER_CIPHERS").Strings(",")
```
Unknwon: To keep in order, please move this line of code below the line `SSH.RootPath = path.Join(homeDir, ".ssh")`.

conf/app.ini
```diff
@@ -116,6 +116,8 @@ SSH_LISTEN_HOST = 0.0.0.0
 SSH_LISTEN_PORT = %(SSH_PORT)s
 ; Root path of SSH directory, default is '~/.ssh', but you have to use '/home/git/.ssh'.
 SSH_ROOT_PATH =
+; Choose the ciphers to support for SSH connections
+SSH_SERVER_CIPHERS = aes128-ctr, aes192-ctr, aes256-ctr, aes128-gcm@openssh.com, arcfour256, arcfour128
```
Unknwon: Have you found any references that declares what ciphers that Go's ssh package supports?

spacetourist: Hi Unknwon - nice to meet you!

Issue #4107 contains an excerpt from the go source for the ssh package, this is the reference i have been working with, i presumed it to be comprehensive.

Let me know if theres anything else and please see revised commit below

Unknwon:
Thanks!

lunny:
@spacetourist could you also send a PR to github.com/go-gitea/gitea ?

spacetourist:
Hi lunny, yeah, happy to do that but i'm new here. Does that mean cloning your repo, making the changes and repeating the process?

lunny:
I think you can add a new remote gitea ref to github.com/go-gitea/gitea and create a new feature branch on your fork and cherry-pick to this new feature branch.

spacetourist:
Phew, i think i've done it - not as straightforward as one might have hoped - you've drifted away from the tree! 

P.S. what are the core differences between gogs and gitea? Thats not exactly clear from your readme

lunny:
Thanks

aaronjwood:
Can I suggest that you prioritize AES-GCM above everything else? I think an authenticated encryption mode should be first on everyone's list.

It might be nice to prevent the use of the arcfour suites entirely. Since most modern chips have the AES-NI instruction set to accelerate AES operations most people would see better performance too.

spacetourist:
I wholly agree with Aaron, certainly that was the driver behind the change in the first place and I doubt there are many cases where the arc ciphers are preferred by anyone other than those involved in downgrade attacks... Having the facility to define the suite used on a system allows those people who might wish to use the arc ciphers to add them back in however as a new contributor I certainly didn't want to impose my priorities on other users of the platform.

PR #4116, opened 2017-02-12
Use very strong ciphers

aaronjwood:
This sets strong ciphers/curves for when TLS is used.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -663,7 +663,18 @@ func runWeb(ctx *cli.Context) error {
 	case setting.SCHEME_HTTP:
 		err = http.ListenAndServe(listenAddr, m)
 	case setting.SCHEME_HTTPS:
-		server := &http.Server{Addr: listenAddr, TLSConfig: &tls.Config{MinVersion: tls.VersionTLS10}, Handler: m}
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for the PR, but CI is failing on Go 1.5:

```
cmd/web.go:674: undefined: tls.TLS_RSA_WITH_AES_256_GCM_SHA384
```

aaronjwood:
You're welcome. Production binaries are currently built with 1.5? Is there any reason not to move to the latest version? I know that the latest release has some big performance gains due to use of SSA in their backend. More information about their use of SSA here https://blog.golang.org/go1.7

You can also support HTTP/2 (which I think is important these days) if you're on anything greater than or equal to 1.6.

If you're not open to moving to the latest release I can adjust the ciphers for 1.5, they just won't be as strong :)

Unknwon:
Builtin HTTPS functionality is convenient  but use proxy (e.g. NGINX, Caddy) to handle HTTPS is more general solution. So it is not worth dropping support of Go 1.5.

aaronjwood:
Sure, I can understand having people use a reverse-proxy instead. It's what I do on my servers too. I've removed that cipher to be compatible with 1.5.

Unknwon:
Thanks!

PR #4121, opened 2017-02-13
Icon redesign

egonelbre:
Icon redesign with .svg.


Unknwon:
This is awesome, thank you very much!

Unknwon:
I'm going to update all documents.

Unknwon:
All places should be updated now. Thanks again!

jorgesumle:
Great work @egonelbre! @Unknwon, you should also update the screenshots in https://gogs.io/

Unknwon:
@jorgesumle will do in next release :D

jxsl13:
Old one was better :c

egonelbre:
@jxsl13 any particular reason you don't like the new version?

jxsl13:
The old one was unique, cute and appealing and you would directly know what it means, especially because of the go language mascot. You could say, it is a "unique selling point", the new one in comparison is just plain, nothing special, it could be the logo of anything, but not this project.

egonelbre:
@jsxl13 Yeah, I agree with the cuteness part. The previous was not legible at smaller sizes, which is why I decided to go with a simpler design. Sure, I agree that there might be designs that I haven't considered that would be a better fit. I'll think about it, maybe I'll have some other better ideas.

Unknwon:
See my response at [forum](https://discuss.gogs.io/t/discussion-what-does-the-community-think-about-the-new-logo/908/2?u=unknwon).

PR #4126, opened 2017-02-15
Fixed tab indentation in file viewer

jjaksic:
In code viewer, for some reason code container's left padding was implemented not with padding, but with ```:before { content: ' ' }```.
Adding a space character at the beginning of each line is harmful, because it's an actual space character and it messes up indentation.
This doesn't matter in space-indented files, but it matters in tab-indented ones (the first tab appears shorter than the following tabs, because there's already a space in front of it), and especially in files with mixed tabs and spaces (space-indented lines get indented one space further than tab-indented ones).

This fix replaces the spaces with a 2px left padding, which fixes the indentation issue.

### DIFF
--- a/public/css/gogs.css
+++ b/public/css/gogs.css
@@ -1431,7 +1431,7 @@ footer .ui.language .menu {
 .repository.file.list #file-content .code-view .lines-code .hljs {
   background-color: white;
   margin: 0;
-  padding: 0 !important;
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks your PR, but please read https://github.com/gogits/gogs/wiki/Contributing-Code#coding-guidelines No.3

Issue #2199, opened 2015-12-15
Syntax highlight class for e.g CMake (and others)


xor-gate:
I see gogs uses highlight.js and CMake is supported (build framework) used in C/C++ style projects.

When gogs sees one of the following files it could highlight with CMake markup: 
* `CMakeLists.txt` (with correct casing)
* `*.cmake`

I search through the gogs sourcecode but as far as I can tell you let highlight.js automatic select the syntax highlighting and dont set the div class attribute as specified here: https://highlightjs.org/usage/

Unknwon:
How does a normal `CMakeLists.txt` looks like?

xor-gate:
Example from https://highlightjs.org/static/demo:
```cmake
cmake_minimum_required(VERSION 2.8.8)
project(cmake_example)

# Show message on Linux platform
if (${CMAKE_SYSTEM_NAME} MATCHES Linux)
    message("Good choice, bro!")
endif()

# Tell CMake to run moc when necessary:
set(CMAKE_AUTOMOC ON)
# As moc files are generated in the binary dir,
# tell CMake to always look for includes there:
set(CMAKE_INCLUDE_CURRENT_DIR ON)

# Widgets finds its own dependencies.
find_package(Qt5Widgets REQUIRED)

add_executable(myproject main.cpp mainwindow.cpp)
qt5_use_modules(myproject Widgets)
```

Unknwon:
Thanks, so `CMakeLists.txt` this name is unique? Like `Makefile`, `Dockerfile`, etc.?

xor-gate:
Yes it is, it works also for `*.cmake` files

Unknwon:
Latest `master` and demo site should be able to show highlight for cmake files correctly, any example repository to test on demo site?

xor-gate:
Awesome it seems to work:
https://try.gogs.io/admin1/mpack/src/cmake/CMakeLists.txt
https://try.gogs.io/admin1/cmaketest/src/2015.06/CommonQtSupport.cmake

Thanks for your time and effort!

Unknwon:
Thank your confirmation!

bouffa:
Syntax highlighting no longer works for `CMakeLists.txt` files, see https://try.gogs.io/admin1/mpack/src/cmake/CMakeLists.txt

Unknwon:
@bouffa highlight of `.txt` has disabled by default.

bouffa:
[...]

PR #4130, opened 2017-02-15
Enable syntax highlighting for CMakeLists.txt files #2199

bouffa:
Enable syntax highlighting for CMakeLists.txt files #2199


Unknwon:
Thanks!

PR #4146, opened 2017-02-18
Update TRANSLATORS

reekoheek:
I dont know if it's right, if it is not, please let me know.

I've got email from joe2010xtmf because I contribute Indonesian language for gogs, and I can open a pull request to file https://github.com/gogits/gogs/blob/master/conf/locale/TRANSLATORS.

Should I pull request this to develop branch? (because I see "master" from the blob url file.


Unknwon:
Thank you!

Issue #4145, opened 2017-02-18
Up to date docker image : based on alpine 3.5 + go 1.8


yaourt:
Are you interested in updating the Docker image to up to date version of alpine (3.5) & go (1.8) ?
I have a not yet fully tested Dockerfile to do so.

It is based on the current Dockerfile + https://github.com/docker-library/golang/blob/132cd70768e3bc269902e4c7b579203f66dc9f64/1.8/alpine/Dockerfile

Unknwon:
Hi, as long as current version works and stable, I don't think worth taking risk to upgrade.

Besides I failed to `go get github.com/gogits/gogs` on alpine3.5 not sure what went wrong.

yaourt:
Sorry, I was not so clear ...

> BTW alpine3.5 comes with Go 1.7.3

I have a Dockerfile based on alpine 3.5, that builds Go 1.8, then with Go 1.8 builds Gogs.
The result is then a Docker image based on alpine 3.5, with Gogs (built with Go 1.8).

I'm going to send a PR so that you'll have a better view of what I'm talking about ...

> Besides I failed to go get github.com/gogits/gogs on alpine3.5 not sure what went wrong.

I think the issue is related to this : https://github.com/niemeyer/gopkg/issues/50
The workaround consists in changing git default config : 
`git config --global http.https://gopkg.in.followRedirects true`

Unknwon:
>I think the issue is related to this : niemeyer/gopkg#50
The workaround consists in changing git default config :
git config --global http.https://gopkg.in.followRedirects true

Good point! Looking forward to the PR.

BTW, does the image size reminds at "30M+" range?

yaourt:
I don't know how the "Compressed size" is computed on hub.docker.com, but here it is what I have locally.

- "mygogs" being my docker images with based on alpine 3.5 / Gogs 0.9.41 (built with Go 1.8)

- "gogs/gogs" being the docker image available on hub.docker.com

```
$ docker images
REPOSITORY                                          TAG                 IMAGE ID            CREATED             SIZE
mygogs                                              1.8                 a990323a427b        52 minutes ago      91.9 MB
gogs/gogs                                           0.9.141             0efbc5cc2550        7 days ago          90 MB
```

Unknwon:
Thanks for the info! Size looks good.

But I think "Gogs 0.9.41" fairly too old? Why not use latest `master` or `develop` to test?

yaourt:
[...]

PR #4150, opened 2017-02-18
 [ #4145 ] Provide an updated Dockerfile and docker/build.sh

yaourt:
Provide an updated Dockerfile and docker/build.sh so that the resulting
Docker image is based on up to date versions.

In order to have an up to date Docker image based on alpine 3.5
  and having Gogs built with Go 1.8

Refs: [ #4145 ]

The Gogs version (tag / branch name) to build can be selected by editing the following block in Dockerfile:

```
# Branch name
#ENV GOGS_VERSION v0.9.141
ENV GOGS_VERSION develop

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,14 @@
-FROM alpine:3.3
-MAINTAINER jp@roemer.im
+FROM alpine:3.5
+
+# Components versions
+ENV GOLANG_VERSION 1.8
+ENV GOLANG_SRC_URL https://golang.org/dl/go$GOLANG_VERSION.src.tar.gz
+ENV GOLANG_SRC_SHA256 406865f587b44be7092f206d73fc1de252600b79b3cacc587b74b5ef5c623596
+
+# Branch name
+#ENV GOGS_VERSION v0.9.141
+ENV GOGS_VERSION develop
+
 
 # Install system utils & Gogs runtime dependencies
 ADD https://github.com/tianon/gosu/releases/download/1.9/gosu-amd64 /usr/sbin/gosu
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,5 +1,14 @@
...
+FROM alpine:3.5
+
+# Components versions
+ENV GOLANG_VERSION 1.8
+ENV GOLANG_SRC_URL https://golang.org/dl/go$GOLANG_VERSION.src.tar.gz
+ENV GOLANG_SRC_SHA256 406865f587b44be7092f206d73fc1de252600b79b3cacc587b74b5ef5c623596
+
+# Branch name
+#ENV GOGS_VERSION v0.9.141
+ENV GOGS_VERSION develop
```
Unknwon: This is not accurate, should just remove this block (line 8-10)

Unknwon: Docker Hub doesn't know which branch it is on when building the image.

Unknwon:
Thanks for the PR!

I actually prefer the `build.sh` way because the code is more clean maintainable, not tight all commands together and very hard to read.

Dockerfile
yaourt: This tag / branch name is used in docker/build.sh script.
To speed up the build process, only this branch / tag is cloned see docker/build.sh / line 12

git clone --single-branch **--branch ${GOGS_VERSION}** --depth 1 https://github.com/gogits/gogs ${GOPATH}/src/github.com/gogits/gogs

Unknwon: The problem is you don't know which branch it is on.

yaourt: In the build-sh script I replaced the `go get ...` command by the `git clone ...` one (as underlined above) on purpose so that the desired Gogs version can be chosen.

Instead of relying on the local Gogs version, the one from where the Dockerfile is coming, this `git clone ...` allows to have it decoupled.

The Dockerfile is only using the content from the local `docker` folder, everything else, the whole gogs sources are coming from github while building the image.

So the GOGS_VERSION variable is then used to know the branch / tag to be cloned (only this branch / tag is retrieved to speedup the build).

My goal was to have a Docker build more or less independent from the local source tree, of course I can understand that this new behavior may not suit you.

Unknwon: >In the build-sh script I replaced the go get ... command by the git clone ... one (as underlined above) on purpose so that the desired Gogs version can be chosen.

You can choose with Docker Hub automated builds, this file is used for that. Not meant for manual build.

yaourt:
See other PR (#4151) on same issue

PR #4153, opened 2017-02-19
this commit fix #3958

### LINKED ISSUE #3958: Deal with '/' in Wiki page title
The issue will be closed without any reasons if it does not satisfy any of following requirements:

- Gogs version (or commit ref): 0.9.97-1472903388.491
- Git version: 2.10.2
- Operating system: Debian 8
- Can you reproduce the bug at https://try.gogs.io: Yes

## Description

I know the support for hierarchical wiki is cooking, but if I try to create a new wiki page with title 'A/B', website throws a 404. I think its a good idea to check for `/` in the wiki title name before POST.

huqiangit:
Deal with '/' in Wiki page title.

### DIFF
--- a/routers/repo/wiki.go
+++ b/routers/repo/wiki.go
@@ -218,7 +218,7 @@ func NewWikiPost(ctx *context.Context, form auth.NewWikiForm) {
 		return
 	}
 
-	ctx.Redirect(ctx.Repo.RepoLink + "/wiki/" + models.ToWikiPageURL(form.Title))
+	ctx.Redirect(ctx.Repo.RepoLink + "/wiki/" + models.ToWikiPageURL(models.ToWikiPageName(form.Title)))
 }
 
 func EditWiki(ctx *context.Context) {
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for the PR... but this solution seems somewhat... weird? Is it just removing all `/`s?

huqiangit:
it not remove any '/'. i just  want to redirect url to the right  wiki page.

Unknwon:
[`models.ToWikiPageName`](https://github.com/gogits/gogs/blob/master/models/wiki.go#L34-L37) replaces all `/` to a ` `(space).

huqiangit:
sorry i am not familiar with golang。
according to this line https://github.com/gogits/gogs/blob/master/models/wiki.go#L102
i think function ToWikiPageName produce each new wiki page url , so, just use this function redirect url.

github also don't support '/'.this commit just deal with 404 fault,when we create a wiki page with '/' in title.

Unknwon:
Indeed it will work, just looks a bit strange. Thank you anyway!

huqiangit:
thank you~

PR #4168, opened 2017-02-21
Some translation changes relates to Ukrainian language.

Argentumbolo:
- Add Ukrainian translation into templates/home.tmpl
    - Fix Ukrainian language name in conf/app.ini
          because on Ukrainian 'language' is female.
    - Add me as translator into conf/locale/TRANSLATORS


Unknwon:
Thank you! I used Google Translator... good to have something to blame :trollface:

PR #4181, opened 2017-02-23
Fix packaging

crohr:
This pull request fixes the packaging for https://packager.io/gh/pkgr/gogs, which started to fail about 2 weeks ago. It switches to the newest version of the go buildpack and adjusts some configuration related to the packaging.

I also ran into an issue while running govendor, with the commit 8be5635c28f40a85ce8e8f65c6118b4ee4e548e9 not longer existing in the macaron repo, hence the change. Let me know if you think I'm doing something wrong here.

The package properly builds on my branch at https://packager.io/gh/pkgr/gogs/build_runs/604

### DIFF
--- a/.pkgr.yml
+++ b/.pkgr.yml
@@ -19,6 +19,8 @@ targets:
   centos-6: &el
     build_dependencies:
       - pam-devel
+      # required for go buildpack
+      - perl-Digest-SHA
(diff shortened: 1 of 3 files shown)


Unknwon:
Thank you!

Issue #4176, opened 2017-02-23
MSSQL attempting to use LIMIT expression instead of TOP


JakeAM:
- Gogs version (or commit ref): 0.10 RC
- Git version: 2.8.1.windows.1
- Operating system: Windows Server 2012
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
  - [X] MSSQL
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [X] Not relevant
- Log gist (usually found in `log/gogs.log`): https://gist.github.com/JakeAM/8fdecbf4aadb32d945616aabdf034cb4

## Description

Unless you have admin privileges, when you try and access an organisation it throws the error found in the gist above. I have investigated it with SQL profiler and it turns out it's trying to use the LIMIT keyword which doesn't exist in SQL Server. Below is the query it tries to run if the user is not an admin.

```
SELECT repository.* FROM repository
	INNER JOIN team_repo 
	ON team_repo.repo_id = repository.id
	WHERE (repository.owner_id = 1 AND repository.is_private = 0) OR team_repo.team_id IN (5)
	GROUP BY repository.id
	ORDER BY updated_unix DESC
	LIMIT 15 OFFSET 0
```

I would guess this is actually an issue with underlying ORM?

Unknwon:
Thanks, didn't know that MSSQL doesn't have `limit`.

Unknwon:
This is claimed to be fixed by merging #4194, please help test on `develop` branch.

Unknwon:
Please notify if fix doesn't work, closing now.

PR #4194, opened 2017-02-26
Replaced raw SQL queries for reading repositories of organization

dlob:
This PR replaces the raw SQL queries for reading repositories of organizations. It resolves issue #4176.

Has been successfully tested with MSSQL Server.


Unknwon:
Thank you!

PR #4238, opened 2017-03-04
Update README.md

naileakim:
As far as I understood, the git user is created in the `build.sh` script?


Unknwon:
Yes, forgot to update README when scripts changed, thank you!

PR #4248, opened 2017-03-06
Checking if protocol is FCGI for suburl is removed

raxetul:
Checking if protocol is FCGI for suburl is removed to support reverse proxies with suburls.

1. Please make sure you are targeting the `develop` branch.
2. Please read contributing guidelines:
4. ... if it is not related to any particular issues, explain why we should not reject your pull request.

**You MUST delete above content including this line before posting; too lazy to take this action considered invalid pull request.**

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -118,9 +118,7 @@ func newMacaron() *macaron.Macaron {
 	if setting.EnableGzip {
 		m.Use(gzip.Gziper())
 	}
-	if setting.Protocol == setting.SCHEME_FCGI {
-		m.SetURLPrefix(setting.AppSubUrl)
-	}
+	m.SetURLPrefix(setting.AppSubUrl)
 	m.Use(macaron.Static(
 		path.Join(setting.StaticRootPath, "public"),
 		macaron.StaticOptions{


Unknwon:
Hi, can you explain more? Also, please read the PR template.

raxetul:
I'm using Gogs behind nginx reverse proxy with URLs https://localhost/Foo (another app) and https://localhost/gogs (I ve configured app.ini ROOT_URL ending with gogs.) Before my commit gogs did not accept  https://localhost/gogs as root and looks for https://localhost/install instead of https://localhost/gogs

raxetul:
I will also check PR template, sorry.

Unknwon:
>I'm using Gogs behind nginx reverse proxy with URLs https://localhost/Foo (another app) and https://localhost/gogs (I ve configured app.ini ROOT_URL ending with gogs.) Before my commit gogs did not accept https://localhost/gogs as root and looks for https://localhost/install instead of https://localhost/gogs

It sounds like more about wrong reverse proxy config, and the changes in this PR doesn't make sense to me. I don't understand code changed under what purpose?

raxetul:
Same reverse config works with my other app, may be fcgi should be prefered
as you said. So don't pull then, if you see something wrong.

7 Mar 2017 ÖÖ 1:10 tarihinde "无闻" <notifications@github.com> yazdı:

I'm using Gogs behind nginx reverse proxy with URLs https://localhost/Foo
(another app) and https://localhost/gogs (I ve configured app.ini ROOT_URL
ending with gogs.) Before my commit gogs did not accept
https://localhost/gogs as root and looks for https://localhost/install
instead of https://localhost/gogs

It sounds like more about wrong reverse proxy config, and the changes in
this PR doesn't make sense to me. I don't understand code changed under
what purpose?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/gogits/gogs/pull/4248#issuecomment-284549988>, or mute
the thread

Unknwon:
>Same reverse config works with my other app

Coherence doesn't equivalent to causing... thank you anyway.

raxetul:
@Unknwon ,
I never think the way  "if it is correct for something, it has to be correct for other things". That not case. I thought  I shouldn't use reverse proxy, I should use FCGI but after some research I see that I'm not wrong also. Choosing http(s) reverse proxy is dependable.

http://stackoverflow.com/questions/4003714/nginx-and-perl-fastcgi-vs-reverse-proxy-psgi-starman
https://www.shimmercat.com/en/blog/articles/talk-to-your-application/
https://forum.xojo.com/15475-comparison-of-speed-between-cgi-standalone-and-reverse-proxy/0

Now I know that my way is not wrong also. It's up to you now.

Issue #3088, opened 2016-05-13
Explore doesn't include private but accessible repos


Dortamur:
Currently, there is no easy way to find a repo by searching, when there are many private repos in organisations. In our setup we heavily use Organisations and Teams where 99% of the Repos are Private, and none of the Private repos in an Organisation show up under Explore, either browsing or searching.

It would be really useful to include _all_ repos that a User has access to, under Explore.

By extension, the API /api/v1/repos/search should also return all accessible Repos, including Private repos in an Organisation that the User has access to via a Team.

de-live-gdev:
My *Explore -> Repositories* is also empty, all repos on my server are set to private. Nothing shows up, doesn't matter if I'm logged in  as admin, or not, got access to the organization, repo owner, member, etc..

bkcsoft:
I can have at look at this

bkcsoft:
Which version of Gogs are you running ? I can't reproduce this :sweat:

de-live-gdev:
0.9.27.0512

![unbenannt](https://cloud.githubusercontent.com/assets/6735650/15326252/970304e2-1c4d-11e6-9f1f-ddafc59e1140.png)

Dortamur:
Mine is 0.9.15.0323.

If I go into Settings for a repo owned by an Organisation, and change Visibility to Not Private, it shows up in Explore. Changing it back to Private, and it no longer shows in Explore. I am in a Team with access to the Repo - and I'm also Admin.

I've also just tested with Repos owned by myself, and the same occurs. Private, and it's not shown in Explore. Not Private, and it is shown.

Dortamur:
I've now upgraded to 0.9.25.0506 (using the .deb packages from packager.io), and Explore is still not showing Private-but-allowed repos.

bkcsoft:
Nvm, I can reproduce it. Working on it :D

maxleistner:
any progress on this? @bkcsoft

master-d:
@Unknwon would you still be willing to examine a pull request for this issue?  I noticed you assigned it to yourself on 7/2/2016 but I was curious and wanted to see if I could fix this issue locally.

Unknwon:
@master-d yes you can try, it is somewhat a complicated query.

Just keep in mind there is no guarantee that any proposed PR will be merged.

master-d:
[...]

PR #4273, opened 2017-03-10
Issue #3088 Allow SearchRepos to return all repos accessible by user

master-d:
Added a new search query and field "UserID" to the SearchRepoOptions struct so that models/repo.go SearchRepositoryByName() function can retrieve all repostiories that the logged in user has access to on the "Explore" page.

The search will not include other people's private repositories even if the logged in user is an admin however it could be modified to behave this way if desired.


models/repo.go
```diff
@@ -1599,9 +1610,10 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
 	if err != nil {
 		return nil, 0, fmt.Errorf("Count: %v", err)
 	}
+	log.Info("Count %d", count)
```
Unknwon: Remove this line.

models/repo.go
```diff
@@ -1583,14 +1584,24 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
 	}
 
 	repos = make([]*Repository, 0, opts.PageSize)
-
-	// Append conditions
-	sess := x.Where("LOWER(lower_name) LIKE ?", "%"+opts.Keyword+"%")
-	if opts.OwnerID > 0 {
-		sess.And("owner_id = ?", opts.OwnerID)
+	sess := x.Alias("r")
```
Unknwon: `repo` is better.

models/repo.go
```diff
@@ -1583,14 +1584,24 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
...
+	// Attempt to find repositories that opts.UserId has access to
+	// This does not include other people's private repositories even if opts.UserId is an admin
+	if !opts.Private && opts.UserID > 0 {
+		sess.Join("LEFT", []string{"team_repo", "tr"}, "tr.repo_id=r.id")
```
Unknwon: Use `access` table to get not-owned, private and has access to repositories is needed, here you didn't consider the case where user is collaborator of private repositories.

So what we to do:

1. To include repos user directly owned
2. To find private repos that has record in `access` table (has such record means user has at least read permission)

master-d: ahh you're right, I forgot to consider collaborators. So you're saying the access table will contain a row for collaborators, and members of a team that has access to a private repository?  If that's true then I don't need to access team_repo and team_user at all then and can get all this information from the access table?

models/repo.go
master-d: agreed. will change

models/repo.go
master-d: will remove sorry about that. I was using the logger before I figured out how to get delve to work with vscode and forgot about that line.

models/repo.go
Unknwon: >So you're saying the access table will contain a row for collaborators, and members of a team that has access to a private repository?

`access` table contains a record if:

1. Repo is not owned by the user
2. User has at least read access to the repo

Unknwon: Therefore, I think we only need to join `repository` and `access` tables.

master-d: gotcha. I'll work on rewriting this query

master-d: Query has been rewritten to use the access table now.  I did 2 separate pushes, but I realize now that that could be confusing so from now on I'll just do one for all code changes.

models/repo.go
```diff
@@ -1583,14 +1584,22 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
...
-
-	// Append conditions
-	sess := x.Where("LOWER(lower_name) LIKE ?", "%"+opts.Keyword+"%")
-	if opts.OwnerID > 0 {
-		sess.And("owner_id = ?", opts.OwnerID)
+	sess := x.Alias("repo")
+	// Attempt to find repositories that opts.UserId has access to
+	// This does not include other people's private repositories even if opts.UserId is an admin
+	if !opts.Private && opts.UserID > 0 {
+		sess.Join("LEFT", []string{"access", "acc"}, "acc.repo_id = repo.id")
```
Unknwon: 1. Is there a particular reason you use `LEFT` not `INNER` for join?
2. Rename `access` to `acc` is over simplifying, just use `access` is fine and obvious.

models/repo.go
```diff
@@ -1583,14 +1584,22 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
...
+		sess.Where("repo.lower_name like ? and (repo.owner_id=? or acc.user_id=? or repo.is_private=?)",
```
Unknwon: All SQL keywords should be in upper case, e.g. `AND`, `OR`.

models/repo.go
master-d: Yes. An inner join wouldn't work here since that would cause the query only to return repositories that had a record in the access table which would exclude public repositories and the users own private repositories. A left join here allows us to return all repositories that match the search string and one or more of the 3 optional conditions I have in parentheses

Unknwon: Got it, thanks for the explanation.

Unknwon:
Might not be related, have you check if user is not searching, can s/he see the private but accessible repositories in explore?

models/repo.go
master-d: I uppercased all keywords and removed the alias for the 'ACCESS' table.

master-d:
I checked the 'Add Repository' function for adding a team repository and that would only return repositories owned by the organization because it didn't pass the 'UserID' field I added to 'SearchRepoOptions' struct.  basically if you don't pass a UserID, the function should return the same results as it did before.  Is there anywhere else I should check?

Unknwon:
By PR itself, it's pretty good now!

Unknwon:
Thank you!

master-d:
No problem.  Thank you for allowing me to make my first contribution to your project.... It's also my first time working with go :)  I follow this project now and might look for additional bugs/features from the mailing list

PR #4289, opened 2017-03-15
Modify GetRecentUpdatedRepos to include all accessible repositories

master-d:
small change for issue #3088 

## models/repo.go  
modfied SearchRepositoryByName() so that an empty string keyword search is allowed  
      
## routers/home.go  
modified RepoSearch functions so that the "Explore" page loads by running an empty string search and will allow a manual empty string search by the user.  An empty string search will return all the repositories that the logged in user has access to (up to the page-size limit).  If the user is not logged in, then only public repos are returned.

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -1555,7 +1555,7 @@ func GetUserMirrorRepositories(userID int64) ([]*Repository, error) {
 // GetRecentUpdatedRepositories returns the list of repositories that are recently updated.
 func GetRecentUpdatedRepositories(page, pageSize int, userID int64) (repos []*Repository, err error) {
 	repos, _, err = SearchRepositoryByName(&SearchRepoOptions{
-		Keyword:  "%",
+		Keyword:  "",
 		UserID:   userID,
 		OrderBy:  "updated_unix DESC",
 		Private:  false,
@@ -1589,9 +1589,6 @@ type SearchRepoOptions struct {
 // SearchRepositoryByName takes keyword and part of repository name to search,
 // it returns results in given range and number of total results.
 func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int64, _ error) {
-	if len(opts.Keyword) == 0 {
-		return repos, 0, nil
-	}
 	opts.Keyword = strings.ToLower(opts.Keyword)
 
 	if opts.Page <= 0 {
@@ -1600,6 +1597,7 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
 
 	repos = make([]*Repository, 0, opts.PageSize)
 	sess := x.Alias("repo")
+
 	// Attempt to find repositories that opts.UserID has access to,
 	// this does not include other people's private repositories even if opts.UserID is an admin.
 	if !opts.Private && opts.UserID > 0 {
--- a/models/repo.go
+++ b/models/repo.go
@@ -1072,14 +1072,14 @@ func CountUserRepositories(userID int64, private bool) int64 {
 	return countRepositories(userID, private)
 }
 
-func Repositories(page, pageSize int) (_ []*Repository, err error) {
+func Repositories(page, pageSize int, _ int64) (_ []*Repository, err error) {
 	repos := make([]*Repository, 0, pageSize)
 	return repos, x.Limit(pageSize, (page-1)*pageSize).Asc("id").Find(&repos)
 }
 
 // RepositoriesWithUsers returns number of repos in given page.
 func RepositoriesWithUsers(page, pageSize int) (_ []*Repository, err error) {
-	repos, err := Repositories(page, pageSize)
+	repos, err := Repositories(page, pageSize, -1)
 	if err != nil {
 		return nil, fmt.Errorf("Repositories: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


Unknwon:
Hi, I do not think allow search by empty string is a good idea (you can argue with references).

Instead, we should make `GetRecentUpdatedRepos` to handle private repos.

master-d:
Ok. I'll make the argument that if ```GetRecentUpdatedRepos``` is modified to return all accessible repositories, then these will be loaded on page load.  Then presuming the user performs a number of non-empty searches, they would be able to retrieve the original list of repos at any time by clearing the search.  If empty string search isn't allowed the user would just have to click on the "Explore" link again to get the original list.  I don't really have a strong opinion either way so if you'd like empty string search eliminated I'll disable it again and make modifications to  ```GetRecentUpdatedRepos```

Unknwon:
I would prefer make changes in `GetRecentUpdatedRepos` because empty string is same as not performing search.

master-d:
Ok I'll work on this tonite.

master-d:
I was working on a fix for this tonite, but it looks like I can't easily change the method signature for ```GetRecentUpdatedRepos``` to include the logged in UserID because this function is set to the ```ranger``` field of the passed in struct ```RepoSearchOptions```.  I thought I could just change the method signature of ```ranger``` field in the struct, but if I do, then routers/admin/repo.go won't compile because it passes the ```models.Repositories``` function as the ranger value and there's no need to include a UserID field for ```models.Repositories``` since it's just used to get a list of all public/private repositories.  I guess Go doesn't support optional function parameters so the only way I can think to get around this would  be if ```ranger``` function accepted a single struct with all search options as an argument or if I don't use ```ranger``` in routers/home.go at all and just do the search manually which is what is being done currently if a UserID is available.  I'm inclined to do the latter unless you have a better idea for this?

master-d:
Just a thought and not really related to the issue above, but I noticed there are a number of Repository search functions available in repo.go and alot of them run their own SQL statements to get results.  It might be better to have one "master" search function that takes a struct of search options and returns results based on the options passed to it.  You could still have multiple "overloaded" search functions, but they would ultimately just make a call to the master function and not run any SQL themselves.

Unknwon:
There are only two places (explore and admin) using `RenderRepoSearch`, I think it's OK to add another argument `userID`.

master-d:
Ok I can do it that way however the ```models.repositories``` function used by the admin page would ignore the UserID parameter passed to it if we go that route.  I'll see if I can use the blank identifier as a function parameter in lieu of an optional argument.

master-d:
I made changes to update the method signature for the ranger property of ```RepoSearchOptions``` and it appears to work as expected however it will allow an empty string search since this is picked up by the ranger like it always has.  I might need to update the "CountRepositories" methods as well to properly count the number of user accessible repos.  The repo count is being returned, but I don't see anywhere where this information is used on the explore page.

models/repo.go
```diff
@@ -1553,9 +1553,18 @@ func GetUserMirrorRepositories(userID int64) ([]*Repository, error) {
 }
 
 // GetRecentUpdatedRepositories returns the list of repositories that are recently updated.
-func GetRecentUpdatedRepositories(page, pageSize int) (repos []*Repository, err error) {
-	return repos, x.Limit(pageSize, (page-1)*pageSize).
-		Where("is_private=?", false).Limit(pageSize).Desc("updated_unix").Find(&repos)
+func GetRecentUpdatedRepositories(page, pageSize int, userID int64) (repos []*Repository, err error) {
+	repos, _, err = SearchRepositoryByName(&SearchRepoOptions{
+		Keyword:  "%",
```
Unknwon: Not tested, but I supposed this could hurt performance badly.

Unknwon: We can do something like `buildAccessibleRepositoriesQuery` similar to [`buildIssuesQuery`](https://github.com/gogits/gogs/blob/master/models/issue.go#L884), to reuse the same part of both `GetRecentUpdatedRepositories` and `SearchRepositoryByName`.

master-d: yeah I'm not sure how this would affect performance either.  Would you consider passing empty string keyword and have ```SearchRepositoryByName``` remove the like clause in the query if an empty string is passed?

master-d: currently empty string searches are handled by the "ranger" function which happens to be ```GetRecentUpdatedRepositories``` for the Explore page which is how it was always handled so empty string search will be enabled either way, but I could change the query and eliminate the like clause for empty string searches to eliminate potential performance problems.

models/repo.go
```diff
@@ -1555,7 +1555,7 @@ func GetUserMirrorRepositories(userID int64) ([]*Repository, error) {
 // GetRecentUpdatedRepositories returns the list of repositories that are recently updated.
 func GetRecentUpdatedRepositories(page, pageSize int, userID int64) (repos []*Repository, err error) {
 	repos, _, err = SearchRepositoryByName(&SearchRepoOptions{
-		Keyword:  "%",
+		Keyword:  "",
```
master-d: I decided to re-enable empty string search for ```GetRecentUpdatedRepositories``` because I couldn't remove the LIKE clause easily without greatly complicating the code in ```SearchRepositoryByName```.  I profiled 3 different versions of the search query on my machine.  
One with LIKE '%%'
One with LIKE '%%%'
and one with the LIKE clause removed and didn't see much of a difference between the 3 queries however I don't have a very large database of repositories either.  If you want I could send you a sql script if you wanted to profile the queries on gogs.io to see if there's any difference.

master-d:
Ok no problem.  My pervious posts may have been confusing, but what I was trying to say was that empty string search has always been enabled through the "ranger" function as [seen here](https://github.com/gogits/gogs/blob/master/routers/home.go#L70).  After I discovered this behavior already existed I didn't try to remove it, but instead changed the ranger function to return user repositories.

Unknwon:
Closed by https://github.com/gogits/gogs/commit/cac7af2c783e12b68e8263b57a150d19f085f4da.

Issue #3963, opened 2016-12-16
Assignee menu text is overwritten when typing name


Corben78:
- Gogs version (or commit ref): last stable 0.9.71.0809
- Git version: 
- Operating system: OpenSuSE
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes
  - [ ] No
  - [ ] Not relevant
- Log gist:

## Description

When creating a new issue, and assigning somebody to it, the text "Assignee" is overwritten by the entry, when typing his name (or at least the first letter of his name).  
Then the assignee is written twice in that place. Once where the text "Assignee" stood before, and the assignee himself below.  
When clearing the assignee entry again, still the "Assignee" text before with the gear symbol next to it is gone. Either the formerly assigned assignee is still written there, or the text "clear assignee"

Especially when a lot of people are working on a project this list can be quite long, so typing is the faster way than searching in the list manually.

markainick:
The same behaviour can also be seen with Labels and Milestone Menus.

Unknwon:
>the text "Assignee" is overwritten by the entry
What entry?

Hi, can you provide some screenshot because I'm pretty much confused.

markainick:
Here's a gif:

![gogsbug_gif](https://cloud.githubusercontent.com/assets/15358419/22417045/69c43cd0-e6d2-11e6-8afb-760168f7b037.gif)

Basically what you try to select by typing overwrites the controller name (in this case "Labels").

Unknwon:
@markainick thanks, is it on demo site?

Corben78:
Yes, for both labels and assignees.

Unknwon:
What browser are you using? I can't reproduce this on demo site using Chrome 56 on Mac OS X.

Corben78:
I'm using mainly Firefox on Linux, but just tried it in Chrome and it happens there too.
Just cross checked, it's the same behavior with Firefox and Internet Explorer on Windows.

Maybe our explanation wasn't accurate enough, sorry about that. I'll try again:
When clicking on the gear symbol to open the list of labels, milestones or assignees, you have two possibilities to select entries: selecting with mouse, or starting to type the first characters of an entry on your keyboard.
[...]

PR #4294, opened 2017-03-15
Assignee menu text is overwritten when typing name #3963

spectre013:
Added the select attribute to the dropdown settings. This prevents the selected option from changing the main label text when typing to select the label.

### DIFF
--- a/public/js/gogs.js
+++ b/public/js/gogs.js
@@ -1233,10 +1233,11 @@ $(document).ready(function () {
 
     // Semantic UI modules.
     $('.ui.dropdown').dropdown({
-        forceSelection: false
+        forceSelection: false,
+        action:'select'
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for the PR! Have you tested other places where we do need this behavior, e.g. selecting owner in create new repository?

Unknwon:
Hmm, combining two independent features/functions together in one PR is not very good... As I will squash the commits, this could cause losing credits from you (if you don't mind about this effect, you don't have to split into two PRs).

spectre013:
yeah, in the process of fixing that now.

spectre013:
removed the other PR there is just one now.

Unknwon:
OK, thank you!

Issue #4277, opened 2017-03-12
Repositories panel in personal settings


Unknwon:
List all accessible repositories and their sizes.

Related: https://github.com/gogits/gogs/issues/1158#issuecomment-246634622

master-d:
I could work on this if @DblK doesn't want it

Unknwon:
Blocked by https://github.com/gogits/gogs/pull/4289.

Unknwon:
Closed by https://github.com/gogits/gogs/pull/4312, can be tested on https://try.gogs.io/user/settings/repositories.

PR #4312, opened 2017-03-19
Repositories panel in personal settings

master-d:
This is my first attempt at adding a personal repositories panel under "User Settings" (Issue #4277).  
One of the things that needs to be reviewed is my use of the ```[admin]``` localization variables since these do not exist under ```[settings]``` yet.  I'm not sure what the process is for adding to the localization files or if you even want to bother with that for this PR or not.  

Currently, it will only display repositories that the logged in user owns and will only allow the logged in user to delete his own repositories.

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -255,7 +255,10 @@ func runWeb(ctx *cli.Context) error {
 			m.Get("", user.SettingsOrganizations)
 			m.Post("/leave", user.SettingsLeaveOrganization)
 		})
-
+		m.Group("/repos", func() {
+			m.Get("", user.SettingsRepos)
+			m.Post("/delete", user.SettingsDeleteRepo)
+		})
 		m.Route("/delete", "GET,POST", user.SettingsDelete)
 	}, reqSignIn, func(ctx *context.Context) {
 		ctx.Data["PageIsUserSettings"] = true
--- a/routers/user/setting.go
+++ b/routers/user/setting.go
@@ -456,6 +456,7 @@ func SettingsLeaveOrganization(ctx *context.Context) {
 func SettingsRepos(ctx *context.Context) {
 
 	ctx.Data["Title"] = ctx.Tr("admin.repositories")
+	ctx.Data["PageIsSettingsRepositories"] = true
 
 	keyword := ctx.Query("q")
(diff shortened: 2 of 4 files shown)


Unknwon:
Hi, can you do rebase instead of merge?

Unknwon:
Can you attach a screenshot?

master-d:
repositories-gogs

Unknwon:
Thanks for the screenshot!

1. Do not 
1. Delete should not allowed.
2. Not checked code but all accessible repos should be listed.
3. Panel heading should be same as navbar, in English it's **Repositories**, make sure use the same i18n locale string.
4. Navbar should be highlighted.
5. I more prefer the way GitHub does:

image

master-d:
Gotcha.  I can make these adjustments. I'm assuming you don't want the delete button available even for user owned repositories?  And you only want this formatting for the User Settings repositories page right and not the admin page?

Unknwon:
>I'm assuming you don't want the delete button available even for user owned repositories?

Correct.

>And you only want this formatting for the User Settings repositories page right and not the admin page?

This page is no relation to admin page.

master-d:
I updated some the html format to match the github template a little better, but haven't hooked up the "leave" button yet and had a few questions about this button:

1. Should the "leave" button appear for shared team repos as well as collaborative repos or only for collaborative repos?
2. What action should be performed when the user clicks "leave" and clicks "yes" to confirmation popup? (this kinda depends on the answer to 1)
3. Should I display the "forked from" information shown in your github example if the repository has been forked?

repositories-gogs

Unknwon:
Thank you!

master-d:
@Unknwon I wasn't done with this PR yet because the leave button isn't wired up.  We're you gonna do this yourself?  I don't mind doing it, but was hoping you could answer the questions I made above.

Unknwon:
Ah... that's what I thought, I'm working on it currently.

master-d:
Ok, also check the condition that determines whether the leave button is displayed or not.  I think it is currently displaying on all non-public repositories that you don't own and that may be incorrect.

Unknwon:
Thanks, I've noticed, almost done.

PR #4330, opened 2017-03-23
Fix language names

sikmir:
This PR fixes the follow issues:

1. wrong language name: "Suomalainen" is incorrect name for Finnish language, should be "suomi".
2. wrong capitalization of language names: different languages have different conventions for capitalization.


Fastidious:
This will not be accepted.

sikmir:
@Fastidious Why? "Suomalainen" is totally wrong name for language.

Fastidious:
Oh, I might have gotten confused, and be wrong, after all.

Unknwon:
Thank you!

@Fastidious changes to `app.ini` can be modified via PR :wink:

PR #4343, opened 2017-03-26
Small JS improvement: Use jQuery .find()

amouhzi:
$('#id test') is a shortcut for $('#id').find('test')

By using find(), we are eliminating some function calls. which is always good ;)


Unknwon:
>By using find(), we are eliminating some function calls. which is always good ;)

Good point, but why this change is necessary?

amouhzi:
It is not necessary, both are working

- `$('#edit-title-input input');` is working and is fast
- `$('#edit-title-input').find('input');` is the same but faster.

As I noticed, it is just a small improvement, not a fix.

Unknwon:
OK...

PR #4345, opened 2017-03-26
Corrected Mirror.NextUpdate not set

cez81:
Mirror.NextUpdate is not properly set because the switch-case has the wrong column name. Changed from next_update**d**_unix to next_update_unix.


Unknwon:
Thank you!

PR #4353, opened 2017-03-28
Fix heading in README.md

davidjb:
The unicode non-breaking space character was breaking the heading format; this switches that character for an ASCII space.


Unknwon:
Thank you!

PR #4361, opened 2017-03-29
Handle mysql error in migration

kylemcc:
This patch adds error handling for mysql in addition to the existing sqlite error check. Without the additional error check, I'm seeing the following error:

`Fail to initialize ORM engine: migrate: do migrate: Error 1146: Table 'gogs.protect_branch_whitelist' doesn't exist`


Unknwon:
Thank you!

Issue #2855, opened 2016-03-18
LDAP admin right after first login


shomodj:
When using LDAP with admin filter gogs will create a "local clone" LDAP user with admin parameter (modules/login.go:296) but if the user is not in the admin group at the time of the first login he/she is created without admin privileges.

But if you add that user later in admin group in LDAP, gogs will check if user is in admin group but it will not sync the data, so user will never be admin even if you add it in LDAP later.

Now this is a common case in big companies, first you create an user in LDAP, and then admin rights are added later per managers approval.

jaydio:
This is indeed a problem. The admin filter should be applied on-demand meaning every time a user logs in. That's a typical use case for accessing a directory service as its optimized for read access.

Lafriks:
Partially my pull request #3062 that does sync users with LDAP fixes it

JonLundy:
I have checked the LDAP logs when authenticating. It actually makes a query with the admin filter at the time of every login. But does not appear to use that result.

Unknwon:
@JonLundy good point..!

DMWOO7:
That's a real problem which confused me too. Seems like there are currently only to ways to work around this bug: 
1. Manually add the admin permission. An acceptable workaround, but make the great LDAP filter for this relatively useless, since you have to change permissions on two places. This can result in errors, when someone forget it. 
2. Deleting the user in gogs: Not a real solution at all, since the user is already using Gogs

Unknwon:
This is claimed to be fixed by merging #4405, please help test on `develop` branch.

matthijsvdr:
Is this working for anyone?
When i add an admin filter it always returns this error:

> [ERROR] [...kg/auth/ldap/ldap.go:304 SearchEntry()] LDAP: Admin search failed: 0 entries

This even happens when the admin filter is the same as the user filter

User filter works, admin does not
```
user: (&(uid=%s)(memberOf=cn=users,cn=groups,cn=accounts,dc=EXAMPLE,dc=COM))
admin: (&(uid=%s)(memberOf=cn=admins,cn=groups,cn=accounts,dc=EXAMPLE,dc=COM))
```

Unknwon:
@matthijsvdr hi, based on the https://gogs.io/docs/features/authentication, I think currently admin filter does not recognize `%s` placeholder.

matthijsvdr:
@Unknwon a that sucks :(

PR #4405, opened 2017-04-08
Add check for user for create/update on LDAP

JonLundy:
This is a fix to #2855. 

The first time a user logs in with LDAP the database saves if they have the admin flag or not. Because of this if the LDAP source changes the user will still have the flag set for admin. This fix does a check to see if the user already exists in the database. If they exist then update, else create.

I think it could be probably improved with some way of checking that the user being updated has the same login source.


Unknwon:
Thank you!

PR #4417, opened 2017-04-11
Better explenation for team membership and granted access

rbukovansky:
As discussed here https://crowdin.com/translate/gogs/1/enus-cs#378 I'm suggesting this update to explanation of team membership and granting of access.


Unknwon:
Thank you!

PR #4421, opened 2017-04-12
Support using CROWD/JIRA for authorization

kkkbird:
support authorization with CROWD/JIRA user server by CROWD REST API.
for api detail, please refer to https://developer.atlassian.com/display/CROWDDEV/Crowd+Developer+Documentation

it is tested on JIRA 7.3.0


Unknwon:
Thanks for the PR!

Because I don't use CROWD/JIRA, is it possible to have a test instance myself to test this feature?

kkkbird:
yes, you can simply start a JIRA instance with images from docker
community, like cptactionhank/atlassian-jira.
please refer to
https://store.docker.com/community/images/cptactionhank/atlassian-jira.

after start it, just use the default configuration to set it up and you may
need to apply for a trial license from atlassian during the installation.

when the setup is completed, login and navi to JIRA administrator > User
management > JIRA user server and click "add application" to add an
application authorization setting for gogs. then you can test the feature.

Unknwon:
Thanks for the info!

PR #4423, opened 2017-04-13
Made the webhook timeout duration longer.

BartWillems:
A 5 seconds default timeout is too low and can cause random timeouts in certain setups.

15 seconds is a nice default timeout, it's not too long, but it has a much lower chance of triggering the timeout.


Unknwon:
Thank you!

Issue #4433, opened 2017-04-17
download empty file -> got 1024 bytes data


wtnb75:
- Gogs version (or commit ref): 0.11.4.0405
- Git version: 2.1.4
- Operating system: Raspbian GNU/Linux 8 (jessie)
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [x] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log gist (usually found in `log/gogs.log`):

## Description

- create new repository
- create empty file (like guide message)
  - touch README.md
  - git init
  - git add README.md
  - git commit -m "first commit"
  - git remote add origin http://your-url/account/repository.git
  - git push -u origin master
- download README.md with raw url
  - curl -i http://your-url/account/repository/raw/master/README.md
  - got `Content-Length: 1024`

maybe this `n > 0` should be `n >= 0`
https://github.com/gogits/gogs/blob/master/routers/repo/download.go#L21

Unknwon:
This is claimed to be fixed by merging #4436, please help test on `develop` branch or https://try.gogs.io.

wtnb75:
I tested on current `develop` branch. (612a7e76f1a47d0159d3b6fa5558c37397bbca10)
it was fixed👍 
thank you. 

```
# curl -i http://localhost:3000/aaaaa/test/raw/master/README.md
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Set-Cookie: lang=en-US; Path=/; Max-Age=2147483647
Set-Cookie: i_like_gogits=e45418e23f88b7ec; Path=/; HttpOnly
Set-Cookie: _csrf=iLNsck0rwot_d6b6ha-eJuSzD486MTQ5NjI0MTc5MjU1MTE3MDQzMg%3D%3D; Path=/; Expires=Thu, 01 Jun 2017 14:43:12 GMT; HttpOnly
Date: Wed, 31 May 2017 14:43:12 GMT
Content-Length: 0

```

PR #4436, opened 2017-04-19
Fix for downloading zero bytes files.

cez81:
Allocated buffer served to client and not properly truncated to number of bytes read.

Fix for #4433


Unknwon:
Thank you!

PR #4440, opened 2017-04-20
Fixes #4439

### LINKED ISSUE #4439: Emoji not rendered on commits view
- Gogs version (or commit ref): 0.11.4.0405
- Git version: n/a
- Operating system: Ubuntu 
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] MSSQL
  - [ ] SQLite
  - [x] Yes, see https://try.gogs.io/c12/emoji-test/commits/master
  - [ ] No
  - [ ] Not relevant
- Log gist (usually found in `log/gogs.log`): n/a

## Description

I am not sure if this is or isn't related to #3023, however I am running Gogs version `0.11.4.0405` and have noticed that on the commit history page emoji aren't rendered even though the `<td>` element has the `has-emoji` class.

Having had a look at [gogs.js](https://github.com/gogits/gogs/blob/6ea9642d6415073c6312c80bf6533a5070630ad2/public/js/gogs.js#L1252) the implementation of emojify looks to be correct.

Looking at [commits_table.tmpl])(https://github.com/gogits/gogs/blob/9e3c83372f19853ac1dc164e2fcf58d9779ac0f0/templates/repo/commits_table.tmpl#L43) a fix is to place the `has-emoji` class onto the `<span>` rather than on the `<td>`.

I am going to fork and create a pull request the above effect.

carbontwelve:
This is a basic pull request to fix #4439; the `has-emoji` class seems to need to be on the html element that contains the emoji. In this case the html element that has the emoji is a child of the element with the `has-emoji` class, it was being ignored.

What I have done here is move the `has-emoji` class from the parent `<tr>`, attaching it to the child `<span>` element that contains the commit message with the potential of having emoji.

Before:
image

After:
image

### DIFF
--- a/templates/repo/commits_table.tmpl
+++ b/templates/repo/commits_table.tmpl
@@ -40,14 +40,14 @@
 							{{end}}
 						</td>
 
-						<td class="message collapsing has-emoji">
+						<td class="message collapsing">
 							{{/* Username or Reponame doesn't present we assume the source repository no longer exists */}}
 							{{if not (and $.Username $.Reponame)}}
 								<span class="ui sha label">{{ShortSHA1 .ID.String}}</span>
 							{{else}}
 								<a rel="nofollow" class="ui sha label" href="{{AppSubURL}}/{{$.Username}}/{{$.Reponame}}/commit/{{.ID}}">{{ShortSHA1 .ID.String}}</a>
 							{{end}}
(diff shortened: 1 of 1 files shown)


Unknwon:
Thank you!

PR #4474, opened 2017-05-02
Rename app.ini attachment toggle from ENABLE to ENABLED

magnetikonline:
An update to the attachment enable/disable toggle to match language used with other config sections (e.g. `[repository.upload]`, `[release.attachment]` or `[mailer]`) - namely:

```
[attachment]
ENABLE = false
```

Becomes...

```
[attachment]
ENABLED = false
```

Understand this might be a backward breaking change for some - and documentation/wiki would require updates too. Happy to assist where possible.

  Peter


Unknwon:
Thank you!

PR #4484, opened 2017-05-04
Exposes some more API for Repo: IsBare (as empty), BaseRepo (as parent), Size

guyzmo:
title says it all.

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -293,6 +293,7 @@ func (repo *Repository) APIFormat(permission *api.Permission) *api.Repository {
 		Description:   repo.Description,
 		Private:       repo.IsPrivate,
 		Empty:         repo.IsBare,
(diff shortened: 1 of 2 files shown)


Unknwon:
The vendor does not seem to match what you propose to gogits/go-gogs-client, I need to hijack this PR.

Unknwon:
Closed by https://github.com/gogits/gogs/commit/c69fa18434f9377557cb3e417d5469e0740e48d1.

guyzmo:
Ah! Indeed I did not update the PR with the solution I implemented for gitea:

https://github.com/go-gitea/gitea/pull/1687/files

the idea is to allow for Parent to be `nil`, and also avoid potential huge dumps in case of long project hierarchies. (which is why the dual `innerAPIFormat`/`APIFormat`.

PR #4517, opened 2017-05-19
Fix pointer mismatch with pam plugin

rjsalts:
I was trying to authenticate via the pam backend plugin and it failed with an exception:
[Macaron] PANIC: runtime error: cgo argument has Go pointer to Go pointer
Someone pointed me to
https://github.com/pmoody-/golang-pam/commit/e5f9028af25fd791b54a23e762ed4fc0505f59ab
which is a similar library. Making the equivalent changes seems to have fixed my issue.

### DIFF
--- a/vendor/github.com/msteinert/pam/transaction.c
+++ b/vendor/github.com/msteinert/pam/transaction.c
@@ -39,13 +39,12 @@ error:
 	return PAM_CONV_ERR;
 }
 
-struct pam_conv *make_pam_conv(void *appdata_ptr)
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks for the PR!

I've updated vendor from its upstream via https://github.com/gogits/gogs/commit/0778d7de80ebbd6b1b7c7d7d5676655b7706f1f4. Please help check if problem solved.

rjsalts:
I'm not seeing any problems with the upstream code either. Thanks

PR #4519, opened 2017-05-19
don't run ssh-keygen check if disabled

### LINKED ISSUE #4507: SSH Key import fails on CentOS/RHEL 5
- Gogs version (or commit ref): 0.14.1
- Git version: 1.7.4.1
- Operating system: CentOS 5
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [x] SQLite
  - [ ] No
  - [x] Not relevant

## Description

SSHKeyGenParsePublicKey fails with OpenSSH_4.3, shipped with CentOS 5.

This is because as per the Windows check in the method, OpenSSH 4..3 also does not output the key type, therefore there are only 3 fields returned and the check fails.

Therefore adding any key fails, making SSH useless.
```
[user@host ~]$ ssh -V
OpenSSH_4.3p2, OpenSSL 0.9.8e-fips-rhel5 01 Jul 2008
[user@host ~]$ ssh-keygen -l -f /home/user/id_rsa.pub 
4096 04:87:56:4a:55:c6:97:a3:b7:98:92:96:be:ae:00:43 /home/user/id_rsa.pub
```

Gibheer:
This pull request somewhat fixes #4507 by moving the the config variable check to a point before the loading of the key is done. That way we avoid the problem of ssh-keygen returning weird stuff and returning an error.

The second commit is a small cleanup of two comments, where the x/crypto/ssh library did not support ed25519 keys, which it does now. The comments were removed and the proper constant was put into place.

### DIFF
--- a/models/ssh_key.go
+++ b/models/ssh_key.go
@@ -23,9 +23,9 @@ import (
 	"golang.org/x/crypto/ssh"
 	log "gopkg.in/clog.v1"
 
-	"github.com/gogits/gogs/pkg/tool"
 	"github.com/gogits/gogs/pkg/process"
 	"github.com/gogits/gogs/pkg/setting"
+	"github.com/gogits/gogs/pkg/tool"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


Unknwon:
Thank you!

Issue #4287, opened 2017-03-15
search repository by description keywords


dxp909:
up

Unknwon:
Reason and use case?

dxp909:
因为在仓储比较多的时候，没办法把所有仓储名字都记得很清楚，这个时候就可以通过描述里的信息来快速定位具体的仓储，比如描述中写着XXX项目中文的名称，搜索的时候就可以直接通过项目名称进行快速定位

carbontwelve:
@Unknwon the company I work for has a use case where this is very useful. We have projects that have job numbers attached and have been including these within the repository description rather than the name because they can often change or one project can have more than one associated job code. 

When exploring repositories it would be useful to be able to enter the job code and the search results include repositories that match.

Without delving into the source code it looks as though the explore search functionality just looks at the repositories names, the additional functionality that both @dxp909 and myself are looking for is for that search functionality to query _both_ the repository name _and_ its description.

carbontwelve:
Hi @Unknwon I notice that this has been added to the 0.12 milestone; since we needed this feature to help with discovering repositories at work I have taken the initiative and made the required small amend.

Please see this pull request with more details: #4548

PR #4548, opened 2017-06-05
Modified repository keyword search to include description for #4287

carbontwelve:
Hi,

This pull request modifies the behavior of the repository search to include searching within the repositories description. This is as per the issue #4287.

I have tested this in our production environment and no issues have been found.

### DIFF
--- a/models/repo.go
+++ b/models/repo.go
@@ -1646,7 +1646,7 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
 		}
 	}
 	if len(opts.Keyword) > 0 {
-		sess.And("repo.lower_name LIKE ?", "%"+strings.ToLower(opts.Keyword)+"%")
+		sess.Where("repo.lower_name LIKE ? OR repo.description LIKE ?", "%"+strings.ToLower(opts.Keyword)+"%", "%"+strings.ToLower(opts.Keyword)+"%")
 	}
 	if opts.OwnerID > 0 {
 		sess.And("repo.owner_id = ?", opts.OwnerID)
--- a/models/repo.go
+++ b/models/repo.go
@@ -1646,7 +1646,7 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
 		}
 	}
 	if len(opts.Keyword) > 0 {
-		sess.Where("repo.lower_name LIKE ? OR repo.description LIKE ?", "%"+strings.ToLower(opts.Keyword)+"%", "%"+strings.ToLower(opts.Keyword)+"%")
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for your PR!

models/repo.go
```diff
@@ -1646,7 +1646,7 @@ func SearchRepositoryByName(opts *SearchRepoOptions) (repos []*Repository, _ int
 		}
 	}
 	if len(opts.Keyword) > 0 {
-		sess.And("repo.lower_name LIKE ?", "%"+strings.ToLower(opts.Keyword)+"%")
+		sess.Where("repo.lower_name LIKE ? OR repo.description LIKE ?", "%"+strings.ToLower(opts.Keyword)+"%", "%"+strings.ToLower(opts.Keyword)+"%")
```
Unknwon: Can you confirm this works? Because I think you should not replace `And` by `Where`, XORM sometimes doesn't behave correct.

carbontwelve: Hi @Unknwon While writing this small PR I looked into the source code of XORM for the `And` and `Where` methods to find that they end up resolving to the same `session.Statement.And` function:

```
// Where provides custom query condition.
func (session *Session) Where(query interface{}, args ...interface{}) *Session {
	session.Statement.Where(query, args...)
	return session
}

// And provides custom query condition.
func (session *Session) And(query interface{}, args ...interface{}) *Session {
	session.Statement.And(query, args...)
	return session
}
```

As you can see the XORM `And` method resolves to `session.Statement.And` and the `Where` method resolves to `session.Statement.Where`. The later looks like the below:

```
// Where add Where statement
func (statement *Statement) Where(query interface{}, args ...interface{}) *Statement {
	return statement.And(query, args...)
}
```

The `where` and `and` method appear to resolve to the same `session.Statement.And` function meaning they have the same functionality.

I have tested that this works in our production environment. You can see below a screen grab from our production environment showing a repository found by its description:

image


I believe that it will work the same if I replace the `where` for  `and`. Thinking about it, considering this is an SQL `AND( a OR b)` it may be more expressive of what is going on if I do replace it.

What do you think?

Unknwon: Have you tested a logged in user, who is a collaborator of other private repositories?

Because at one time I was troubled by `Where` that it discards previous conditions and stats over. Similiar situation here: https://github.com/gogits/gogs/pull/4548/files#diff-d4e2553a87eb8b329efd7335f028903cR1641

carbontwelve: @Unknwon I haven't done so, or at least not sure if I have. How would I configure a test account to do so?

Looking at the xorm code `Where` returns for all intents and purposes an identical code path to `And`: https://github.com/gogits/gogs/blob/master/vendor/github.com/go-xorm/xorm/statement.go#L151

I have modified this pull request to replace my usage of `Where` with `And` (19c7bbb) and have tested that it still works as expected when searching for repositories by their name or by their description.

Unknwon:
Thank you!

PR #4584, opened 2017-06-26
Do not include line numbers in diff selection

eigan:
Marking text in a diff will include the line numbers, resulting in copy pastes like this:

```
43	37	
     /**
44		
-     * @return array
38	
+     * @inheritDoc
45	39	
      */
```

This pull request will tell browser to not include the line numbers when selecting text. Copy paste will then behave like the standard code-view. This is not a problem in code-view, since the structure of the table rendering the code is a bit different. 

This css property is [widely supported](http://caniuse.com/#feat=user-select-none). IE9 and lower is the only (most used) browsers not supporting it.

I could not find a relevant issue.

### DIFF
--- a/public/css/gogs.css
+++ b/public/css/gogs.css
@@ -2138,6 +2138,7 @@ footer .ui.language .menu {
 .repository .diff-file-box .code-diff .lines-num {
   border-right: 1px solid #d4d4d5;
   padding: 0 5px;
+  user-select: none;
 }
 .repository .diff-file-box .code-diff .lines-num.lines-num-old,
 .repository .diff-file-box .code-diff .lines-num.lines-num-new {


Unknwon:
Thank you!

Unknwon:
Hmm, just tested on demo site, it does not seem to work and found https://css-tricks.com/almanac/properties/u/user-select/. 

>However it's a bit buggy. ... WebKit still allows the text to be copied if you select elements around it.

eigan:
noo :(

Github puts the line number in an attribute on `<td>`. And uses css with the [attr](https://developer.mozilla.org/en/docs/Web/CSS/attr) property.

```
HTML
<td .... data-line-number="22">
 <span class="blob-num"></span>

CSS
.blob-num::before {
    content: attr(data-line-number);
}
```

I not familiar with the template syntax. But I guess one could modify the [section_unified.tmpl](https://github.com/gogits/gogs/blob/7e15ff9486cde36b76f3d39b7db23f89278acb7d/templates/repo/diff/section_unified.tmpl) and test this..

Unknwon:
Thanks for the tip! Problem solved.  :+1: 

Patch has been pushed to `develop` branch.

PR #4633, opened 2017-07-19
api: fix nil pointer dereference on repo/forks

barracks510:
Sets BaseRepo value for each resulting fork in GetForks().
Finds correct permissions for parent repository in APIFormat().

The output JSON is a little redundant, perhaps we should offer a
different view of the api.Repository without the Parent attribute?


Unknwon:
Thank you!

Issue #4684, opened 2017-08-10
LDAP group verification doesn't work when using 'dn' as user attribute


dev-dsp:
- Gogs version (or commit ref):  0.11.28.0718
- Operating system: CentOS 7
- Database:
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] MSSQL
  - [ ] SQLite
  - [x] Not relevant
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log:
```
2017/08/10 03:17:24 [TRACE] Fetching attributes 'uid', 'givenName', 'sn', 'mail', 'dn' with filter '(uid=testuser)' and base 'uid=testuser,ou=people,dc=example,dc=tk'
2017/08/10 03:17:24 [TRACE] LDAP: Fetching groups 'member' with filter '(cn=gogs-users)' and base 'ou=groups,dc=example,dc=tk'
2017/08/10 03:17:24 [TRACE] LDAP: Group membership test failed [username: testuser, group_member_uid: member, user_uid: 
```

## Description
Group membership verification doesn't work correctly with following options set:
```
Group Attribute Containing List of Users: member
User Attribute Listed in Group: dn
```
'dn' is just not used and 'user_uid' is blank, as you can see in log above
```
$ >>> ldapsearch -x "(cn=gogs-users)" -b 'ou=groups,dc=example,dc=tk' member -LLL
dn: cn=gogs-users,ou=groups,dc=example,dc=tk
member: uid=testuser,ou=people,dc=example,dc=tk
```

aboron:
One bug I see from my pull request a while back is that I introduced a typo on line 272 of the file gogs/pkg/auth/ldap/ldap.go.

this:
`		} else if len(sr.Entries) < 1 {`
should be:
`		} else if len(srg.Entries) < 1 {`

This would affect group membership checking, since it would not be looking at the correct result set.

aboron:
So, it looks like the GetAttributeValue() function will not return a value for an attribute called 'dn' since it considers that to be the identifier of the root object.  So the specific case of searching for group members by 'dn' will have to be tested and handled differently.

Looking into this now.

aboron:
Proposed fix available in develop branch of https://github.com/aboron/gogs - just changed the one [ldap.go](https://raw.githubusercontent.com/aboron/gogs/develop/pkg/auth/ldap/ldap.go) file: 
[...]

PR #4688, opened 2017-08-11
Fixes to group membership search handling

aboron:
Fix for the case where group members are listed by member 'dn' - Issue (#4684)
Also, fixed typo in group member list return size check.


Unknwon:
Thank you!

PR #4715, opened 2017-08-22
Add option to send the user to a login box for the landing page

acr31:
**Summary**: allow another configuration choice for the default landing page.

**Changes**: adds a new configuration choice `login` for the LandingPage option in `pkg/setting/setting.go`

**Rationale**: I send users directly to the login page when they visit the site rather than requiring them to click 'Sign in'. I am sharing this PR in case you thought that would be useful to others.


duckbrain:
I believe you can accomplish the same thing with

```ini
REQUIRE_SIGNIN_VIEW = true
```

acr31:
Ah. Even better. Thanks - I'll do that instead.

Issue #4725, opened 2017-08-28
go-import meta tag is wrong


MehSha:
- Gogs 0.11.29 (or commit ref): 
- Git version: 
- Operating system: Linux
- Database (use `[x]`):
  - [x] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [ ] Not relevant

refs #1878 , 
when trying to go get my.org/me/repo/subdir, it clones whole my.org/me/repo.git into the subdir directory, I expect only contents of subdir to be there.
reason:
go get expects a meta tag, that in it's content we have following:  "GOPATH_install_dir VCS VCS_Address". 
I checked on github, and for example meta tag for cockroachdb sub-pkgs are:
request: curl https://github.com/cockroachdb/cockroach/tree/master/pkg/sql?go-get=1
response: meta name="go-import" content="github.com/cockroachdb/cockroach git https://github.com/cockroachdb/cockroach.git"

while gogs 0.11.29 generates:
meta name="go-import" content="my.org/me/repo/subdir git https://my.org/me/repo.git"

the subdir at the end of "my.org/me/repo/subdir" is extraneous and leads to incorrect behavior.
please fix.

dolmen:
I have the same issue.
To sum-up: the meta `go-import` should always point to the root of the repo, not to the subdir requested.

dolmen:
There is a fix in #4726.

sdboyer:
glad to see this is being fixed - we've had a couple gogs users show up in the dep queue with issues arising from this problem: golang/dep#1213 golang/dep#1120.

the bug makes it impossible to use dep (and likely at least difficult for other tools) to reference non-root packages in repositories stored with gogs. the fix can't land soon enough 😄 

also, just to reiterate the OP's information by quoting from the [official go toolchain docs](https://golang.org/cmd/go/#hdr-Remote_import_paths):

> If the import path is not a known code hosting site and also lacks a version control qualifier, the go tool attempts to fetch the import over https/http and looks for a <meta> tag in the document's HTML <head>.
>
> The meta tag has the form:
>
> ```
> <meta name="go-import" content="import-prefix vcs repo-root">
> ```
[...]

PR #4726, opened 2017-08-28
fixed go-import meta refs #4725

MehSha:


Unknwon:
Ah... already fixed, but thank you anyway!

PR #4762, opened 2017-09-07
travis: update go versions

dvrkps:


Unknwon:
Ah... updated already, but thank you anyway!

PR #4773, opened 2017-09-15
Dingtalk webhook support

athurg:
Add support for Dingtalk as webhook, which is noticed on [Gogs's Discussion](https://discuss.gogs.io/t/webhook/1277/2) .

I just test all the functions on my mac. It'll be welcome to be checked on more enviroments. And if there're any problems, please contact me to fix.

Thanks a lot!


Unknwon:
Thank you!

PR #4776, opened 2017-09-16
Allow setting UID and GID when running a docker container

### LINKED ISSUE #3520: Allow UID for git user in docker container to be specified via ENV variable
- Operating system: Docker 
- Feature Request:
  ## Description
  It would be helpful for users who may use NFS or other shared storage for the gogs volume to have the UID of the git user specified in the add user command in build.sh.  Basically, a variable such as GIT_UID, if set would invoke the add user command with the speceific user id, if not set, then don't run the command with a UID specified. 

Thanks

robertbeal:
Fix #3520 by allowing the UID and GID of the `git` user to be changed by passing the environment variables `PUID` and `GUID` in as part of the docker run command.

This is done using `usermod` and `groupmod` (included in the `shadow` package in the alpine testing repo). 

Prior to this pull request, gogs files are written using uid/gid 1000:1000 as these are the id's of the docker `git` user. Which in most cases, may not map to a meaningful user on the host. 

```
sudo docker create \
    --name=gogs \
    -p 22:22 \
    -p 3000:3000 \
    -e PUID=$(id $USER -u) -e PGID=$(id $USER -g) \
    -v /var/gogs:/data \
    -t gogs/gogs
```
Apologies for changing so much of the finalize.sh script. I had to in order for the container to build correctly (beforehand it would build, but /app/gogs/docker was empty) so that I could test it works.

I've also added a usermod command when creating the `git` user in a second commit. I believe simply doing `passwd -u` leaves the user in an insecure state. The usermod command means the user is unlocked but with password disabled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,19 +3,33 @@ FROM alpine:3.5
 # Install system utils & Gogs runtime dependencies
 ADD https://github.com/tianon/gosu/releases/download/1.9/gosu-amd64 /usr/sbin/gosu
 RUN chmod +x /usr/sbin/gosu \
- && apk --no-cache --no-progress add ca-certificates bash git linux-pam s6 curl openssh socat tzdata
+  && echo http://dl-2.alpinelinux.org/alpine/edge/community/ >> /etc/apk/repositories \
+  && apk --no-cache --no-progress add \
+    bash \
+    ca-certificates \
+    curl \
+    git \
+    linux-pam \
+    openssh \
+    s6 \
+    shadow \
+    socat \
+    tzdata
 
 ENV GOGS_CUSTOM /data/gogs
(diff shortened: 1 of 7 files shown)


robertbeal:
I ended up writing a fresh Dockerfile. I've got the above working, alpine based, using s6 overlay, su-exec (instead of gosu as it's 4kb instead of 1.8mb). The image (compressed) is 16mb smaller than the gogs one (although may be missing a few bits). 

But... best of all the container can run in` `--read-only` mode (don't think the current gogs one can) making it much more prod friendly. 

https://github.com/robertbeal/docker-gogs/blob/master/Dockerfile

Unknwon:
Thank you!

Issue #4778, opened 2017-09-17
Basic functionality needs JavaScript


rambo-wears-a-pinky-thong:
- Gogs version (or commit ref):  dev
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes
  - [ ] No
  - [ ] Not relevant

## Description

Gogs works more or less well without javascript, but there are some features that are not accessible because the top menu requires javascript. These are 

- create new organization
- create new mirror
- change user settings
- logout

could the menus on the top bar please  be replaced with CSS/HTML menus only?

jorgesumle:
There are some dropdowns in the issue page and in the pull request page which don't work when JavaScript is disabled. The branch selectors don't work neither.

rambo-wears-a-pinky-thong:
An easy solution for this is the `button:hover > .dropdown-menu { display: block; }` selector in a `<noscript/>` tag. Or alternatively with a hidden checkbox as shown [here](https://www.freakspot.net/barra-de-navegacion-de-bootstrap-sin-javascript-en.html).

rambo-wears-a-pinky-thong:
If dropdown menus can be made to work with only CSS/HTML, it would be great. In general, removing as much javascript as possible is usually a good idea.

Unknwon:
This is claimed to be fixed by merging #4780, please help test on `develop` branch or https://try.gogs.io.

PR #4780, opened 2017-09-17
Make dropdowns menus work when JavaScript is disabled

jorgesumle:
Attempt to make https://github.com/gogits/gogs/issues/4778 better. It localizes the message which appears for people who have JavaScript disabled and makes the dropdowns work with CSS (only for people who have JS disabled, using the `<noscript>` HTML tag)


Unknwon:
Thank you!

Issue #4514, opened 2017-05-18
Protected branch can be deleted from the web after merge request


mgxian:
- Gogs version (or commit ref): 0.11.6.0407
- Git version: 2.11.1
- Operating system:  docker on Centos6 (use docker image gogs/gogs:0.11.4)
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [x] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log gist (usually found in `log/gogs.log`):

## Description
I protect master branch, preproduction branch and production branch. but when I agree a pull request(merge master branch to preproduction branch), and then there has a delete button, when I click it, my master branch was deleted. I think this is a bug.

jakejanson:
Can confirm this is still a bug in version 0.11.23.0625 with MySQL database. I would make a pull request to fix this, but unfortunately I don't know Go. (If anyone could tell me roughly what and where I have to modify, I could do it.)

bayangan1991:
@jakejanson if you tell me what binary you need I can compile it for you.

Unknwon:
This is claimed to be fixed by merging #4803, please help test on `develop` branch or https://try.gogs.io.

PR #4803, opened 2017-09-29
Disallow web ui to delete protected branch

bayangan1991:
Hi,

Addresses issue #4514 

Simple pull request that remove the ability for users to delete protected branches after a successful merge.

I have compiled and tested and it works as intended for me on Windows and Linux/amd64


Unknwon:
Thank you!

PR #4807, opened 2017-10-01
Fix the language name for zh_TW

yan12125:
台湾 is in Simplified Chinese (zh_CN), and 臺灣 is the correct version for Traditional Chinese (zh_TW).

Both 臺灣 and 台灣 are commonly used in Taiwan. According to [1][2], 臺灣 should be used in official documents, so I choose this version.

[1] http://news.ltn.com.tw/news/focus/paper/451629
[2] https://zh.wikipedia.org/wiki/%E8%87%BA%E7%81%A3#.E3.80.8C-.7B.E8.87.BA.7D-.E3.80.8D.E8.88.87.E3.80.8C-.7B.E5.8F.B0.7D-.E3.80.8D


Unknwon:
Thank you!

PR #4813, opened 2017-10-06
Add CLI subcommands for all admin dashboard actions

knedlsepp:
This PR adds CLI subcommands for all operations available in the web admin panel.
The `gogs admin` command now looks like this:
```
NAME:
   Gogs admin - Allow using internal logic of Gogs without hacking into the source code
to make automatic initialization process more smoothly

USAGE:
   Gogs admin command [command options] [arguments...]

COMMANDS:
     create-user                  Create a new user in database
     delete-inactive-users        Delete all inactive accounts
     delete-repository-archives   Delete all repositories archives
     delete-missing-repositories  Delete all repository records that lost Git files
     collect-garbage              Do garbage collection on repositories
     rewrite-public-keys          Rewrite '.ssh/authorized_keys' file (caution: non-Gogs keys will be lost)
     resync-hooks                 Resync pre-receive, update and post-receive hooks
     reinit-missing-repositories  Reinitialize all repository records that lost Git files

OPTIONS:
   --help, -h  show help
```

### DIFF
--- a/cmd/admin.go
+++ b/cmd/admin.go
@@ -6,6 +6,8 @@ package cmd
 
 import (
 	"fmt"
+	"reflect"
+	"runtime"
 
 	"github.com/urfave/cli"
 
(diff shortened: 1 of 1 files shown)


Unknwon:
Thank you!

knedlsepp:
@Unknwon I just realized that these new subcommands only work when a `gogs web` instance is running. Is there a quick way to fix this?

Unknwon:
@knedlsepp I think it depends on which subcommands you're looking for executing without web instance.

Issue #3585, opened 2016-09-02
Use of HTML tags in issues' comments


pot4u8divhduh:
- Gogs version (or commit ref):  try.gogs.io
- Git version: try.gogs.io
- Operating system: try.gogs.io
- Database (use `[x]`): try.gogs.io
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes    https://try.gogs.io/rrrrrrrr/rrrrrrrr/issues
  - [ ] No
  - [ ] Not relevant
- Log gist:
## Description

When using tags in the comments section of issues, for example _iframe_ or _script_, all the text after them is no longer rendered. If you write a comment with this text for example

```
Use iframe:

<iframe>

more text....
```

[only the first line is displayed](https://try.gogs.io/rrrrrrrr/rrrrrrrr/issues/1)

Unknwon:
It does what exactly needs to be done. because tag is not complete. https://try.gogs.io/rrrrrrrr/rrrrrrrr/issues/1#issuecomment-1072

bkcsoft:
@Unknwon maybe we should use HTMLEscape instead of HTMLStrip for this? Even though GitHub doesn't  show them. (Always nice to be better than GitHub :laughing: )

```

Foobar
<script>alert("foobar");</script>
Foobar
```
Foobar
<script>alert("foobar");</script>
Foobar

Unknwon:
>should use HTMLEscape instead of HTMLStrip

Where?

makarchuk:
There is no HTMLStrip here, unclosed tag is stripped by markdown renderer. It's possible to escape html before passing to markdown renderer. It worked fine for me. 

![screenshot from 2017-10-14 16-51-40](https://user-images.githubusercontent.com/13999738/31575425-1bfa435c-b100-11e7-955e-b15693042bcf.png)

Should I create PR?

Unknwon:
@makarchuk sure! We can take a look and merge if works good.

makarchuk:
Sorry my previous screenshot was missleading. I only sanitized issue titles and forgot about Issue comments over there. So comment looked fine (while if you sanitized it doesn't).  
I was only able to fix it partially (line after unclosed `<iframe>`) is visible now).   
I can fix it completely but It'll break markdown standard and I'm not really comfortable with that.  
[...]

PR #4825, opened 2017-10-15
Losen HTML checks so unclosed tags won't hide the insides

makarchuk:
Improvements for #3585. 
screenshot from 2017-10-15 15-13-10


pkg/markup/markdown.go
```diff
@@ -119,6 +119,7 @@ func (options *MarkdownRenderer) ListItem(out *bytes.Buffer, text []byte, flags
 // RawMarkdown renders content in Markdown syntax to HTML without handling special links.
 func RawMarkdown(body []byte, urlPrefix string) []byte {
 	htmlFlags := 0
+	htmlFlags |= blackfriday.HTML_SKIP_HTML
```
Unknwon: How about HTML tables? Will table still be rendered with this flag?

PR #4835, opened 2017-10-23
Add API endpoint /api/v1/user/orgs

syndbg:
# Why is this needed?

We're more than pleased with Gogs and want to automate completely a migration from GitLab to Gogs.
We need API to create organizations.

# Why we're not using `/api/v1/admin/users/<username>/orgs` ?

Although the same can be achieved with the above mentioned endpoint, 
I don't feel it's correct since it's an admin endpoint and requires you to specify the `username` that will be the owner of the repo.

Also there's a feature request at https://github.com/gogits/go-gogs-client/issues/77.

~~Once this is merged I'll expose Client API for `go-gogs-client`.~~ Exposed `go-gogs-client` in https://github.com/gogits/go-gogs-client/pull/84

# Tests

I haven't invested time into better code coverage due to the fact that we need to ship this soon. I will add tests in a separate PR later if that's okay with you.

@Unknwon

### DIFF
--- a/routes/api/v1/org/org.go
+++ b/routes/api/v1/org/org.go
@@ -13,6 +13,34 @@ import (
 	"github.com/gogits/gogs/routes/api/v1/user"
 )
 
+func CreateOrgForUser(c *context.APIContext, apiForm api.CreateOrgOption, user *models.User) {
+	if c.Written() {
+		return
+	}
+
+	org := &models.User{
+		Name:        apiForm.UserName,
+		FullName:    apiForm.FullName,
+		Description: apiForm.Description,
+		Website:     apiForm.Website,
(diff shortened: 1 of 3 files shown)


routes/api/v1/org/org.go
```diff
@@ -31,6 +59,11 @@ func ListMyOrgs(c *context.APIContext) {
 	listUserOrgs(c, c.User, true)
 }
 
+// https://github.com/gogits/go-gogs-client/wiki/Organizations#create-your-organization
```
syndbg: This wiki does not exist yet! Once it's merged I'll add it :100:

Unknwon:
Thank you!

PR #4862, opened 2017-11-10
Fix English grammar for delete repository dialog

33mhz:
The grammar for the "repository delete dialog" in English is incorrect.


Unknwon:
Thank you! But for en-GB, the fixes will only take effect if you fix them on https://crowdin.com/project/gogs.

PR #4878, opened 2017-11-20
Add utf8mb4 to connStr for mysql. Visit (#4131)

### LINKED ISSUE #4131: Can't create issue with 🐞 in title
- Gogs version (or commit ref): 
- Git version: 
- Operating system: 
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] SQLite
  - [ ] No
  - [ ] Not relevant
- Log gist:

## Description
This is tested in `try.gogs.io`. Can't provide link because issues doesn't get created. Easily reproducible, simply create a new issue, starting the topic with an emoji, and it will render a `500`.
...

m2nlight:
Just add utf8mb4 to connStr for mysql.

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -134,10 +134,10 @@ func getEngine() (*xorm.Engine, error) {
 	switch DbCfg.Type {
 	case "mysql":
 		if DbCfg.Host[0] == '/' { // looks like a unix socket
-			connStr = fmt.Sprintf("%s:%s@unix(%s)/%s%scharset=utf8&parseTime=true",
+			connStr = fmt.Sprintf("%s:%s@unix(%s)/%s%scharset=utf8mb4,utf8&parseTime=true",
 				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
 		} else {
-			connStr = fmt.Sprintf("%s:%s@tcp(%s)/%s%scharset=utf8&parseTime=true",
+			connStr = fmt.Sprintf("%s:%s@tcp(%s)/%s%scharset=utf8mb4,utf8&parseTime=true",
 				DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
 		}
 	case "postgres":


Unknwon:
Thanks for the PR!

Just to confirm, if the host does not support utf8mb4, it falls back to utf8 (as what it's currently is)?

m2nlight:
@Unknwon Yes, this charset can [fallback to utf8](https://github.com/gogits/gogs/blob/develop/vendor/github.com/go-sql-driver/mysql/README.md#charset). But, I got a ERROR at gogs install page:
```
Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'utf8' at line 1
```
So, I need remove `,utf8` from charset to fix that. Then it's work well.
New code:
```
case "mysql":
	if DbCfg.Host[0] == '/' { // looks like a unix socket
		connStr = fmt.Sprintf("%s:%s@unix(%s)/%s%scharset=utf8mb4&parseTime=true",
			DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
	} else {
		connStr = fmt.Sprintf("%s:%s@tcp(%s)/%s%scharset=utf8mb4&parseTime=true",
			DbCfg.User, DbCfg.Passwd, DbCfg.Host, DbCfg.Name, Param)
	}
```

Unknwon:
But the new code will fail to fallback to utf8? Then old server could not start anymore?

m2nlight:
New code only `charset=utf8mb4`.

utf8mb4 required mysql version 5.5.3+

Unknwon:
@m2nlight I see Gogs requires at least MySQL 5.5.3, so I think won't be a problem.

One question remain: what will happen if Gogs tries to connect a MySQL that was using utf8 but now utf8mb4? Does it matter at all?

m2nlight:
@Unknwon utf8mb4 compatible utf8. Using utf8mb4 connected utf8 database is ok.
Wanna emoji charactors, the DBManager can manual 'alter' those column type to utf8mb4.

Unknwon:
OK, lgtm.

PR #4880, opened 2017-11-21
 Update models.go. charset=utf8mb4 for mysql. 

m2nlight:
Remove `,utf8` from `charset=utf8mb4,utf8`


Unknwon:
Hi... this is already changed.

PR #4888, opened 2017-11-22
Fix overextending of variable name

Androbin:
Before: `outPath=./output_$outPlattform_$outArch`
Variable referenced is `outPlattform_`

The shell includes the subsequent underscore as part of the variable name.

Before: `outPath=./output_${outPlattform}_$outArch`
Variable referenced is `outPlattform`


Unknwon:
Thanks!

PR #4908, opened 2017-12-02
Update TRANSLATORS

feikname:
I read a message in Crowdin (sent Feb 11th, but I didn't login again for quite some time) that said I could open a pull request to include myself in the translators list, so here am I hehe.


Unknwon:
Thank you!

PR #4934, opened 2017-12-12
Use identicon image as the default gravatar.

Arnie97:
Inspired by go-gitea/gitea#2767:
> Use geometric pattern image as default gravatar. It is easier to distinguish different people.
>
> Before:
> 
> After:
>


Unknwon:
Thank you!

PR #4965, opened 2017-12-31
Fix file commit history pagination

douglaswth:
- The `commits_table.tmpl` pagination uses a `$.FileName` variable in
  order to generate the next/previous URLs, but it seems like that
  variable was no longer being populated in `renderCommits`.


Unknwon:
Thank you!

PR #4966, opened 2017-12-31
Update year in footer

STayinloves:
Happy new year!


Fastidious:
If the year is to be updated each year, calling up a `go` built-in function to print current year would be the cleanest and best solution, and not this. Alas, I don't know how to do it, but I am sure someone else who knows `go` knows.

Unknwon:
:joy:

PR #4976, opened 2018-01-05
Issuelist labels

JamesGKent:
this pull request adds issue labels (as normally appear in the /repo/issues tab) to the /issues page.
it is not related to an open issue in the tracker, but it does standardize the appearance between the places where lists of issues are shown.


Unknwon:
Hi... I don't think it would work? Because in the dashboard, issues are coming from different repositories and labels are different.

JamesGKent:
when you say you don't think this would work? I have this set up on my gogs install and it seems to work fine?
is there anything you think i should do to test? anything you think wouldn't work or would break it?

JamesGKent:
how it looks for me:
test

Unknwon:
Ah, I understand now. My bad.

PR #4985, opened 2018-01-09
Fix repo-clone-url input size

juliengk:
image


Unknwon:
Hi, can you provide more detail where the size doesn't work, and why it now works based on your fix?

PR #5003, opened 2018-01-18
[Docker][RFC] Use upstream go image

J0WI:
There is no need to build go from source. Docker is already doing this and provides a well supported base image. Those official images are also continuously maintained and scanned for security issues.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/


Unknwon:
Hi, how is the final image size built based on official Go image?

J0WI:
It's 391MB on my amd64 machine. The golang base image (1.8) is 78MB, so there are still many ways to improve dependencies.
1.8 is no longer supported btw.

Unknwon:
It seems way too large than the current size (60MB)...

J0WI:
How do you measure it? When I build the current master it's much bigger than this (I get the size from `docker images`).

Unknwon:
In case of local size of current image (just pulled):

```
➜  ~ docker images
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
gogs/gogs                      latest              f23a2efe3e3f        About an hour ago   156MB
```

Nols1000:
I think the reason for the size of the image is that the official golang:alpine image contains the source of the packages to build go applications. I think the best way to build a minimized image would be to compile a binary with all dependencies built in (not linked) and run it in a "scratch"-image or for a little more script support in "busybox" or "alpine". Check out this blog post [https://blog.codeship.com/building-minimal-docker-containers-for-go-applications/](https://blog.codeship.com/building-minimal-docker-containers-for-go-applications/).

J0WI:
> I think the reason for the size of the image is that the official golang:alpine image contains the source of the packages to build go applications.

I think that's expected that others can use this image as a base for their own changes.

Unknwon:
The base image of Go will make Gogs's image way too large and since that's not necessary, I don't think we should change our based image.

Unknwon:
By a quick look, isn't it the way it's been done right now?

J0WI:
I don't think so. There is an [example](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds) for a go application, that looks much simpler that the current procedure with build shell scripts. The upstream image also supports a range of different architectures out of the box.

Unknwon:
I the current multi-stage Docker build is doing a good job.

J0WI:
Note: this may not work for official images: https://github.com/docker-library/official-images/issues/3383

Issue #4180, opened 2017-02-23
Dockerfile and docker hub images for the Raspberry Pi are 5 months old


praticamentetilde:
As said above, the images in [this docker hub repo](https://hub.docker.com/r/gogs/gogs-rpi/) are quite old. I've tried to `docker build` manually, too. `Dockerfile.rpi` seems not up to date:

- the parent image is quite old. There are new Alpine 3.5 based images now (such as [this](https://hub.docker.com/r/resin/raspberrypi2-alpine/));
- it does not `RUN` `build-go.sh` nor `finalize.sh`;
- from `master`, `RUN` `build.sh` fails at line 16 (`cd ${GOPATH}/src/github.com/gogits/gogs`).

Hope the information above is helpful.

Unknwon:
@praticamentetilde Hi, is it possible that you could make a PR to adjust changes in `Dockerfile.rpi`?

praticamentetilde:
I've already tried to make some adjustments to the Dockerfile, but I hadn't any luck. I will continue to try.

atomi:
I've been using `armhf/alpine:edge` and I build gogs directly on the rpi and not in a docker container, then I just `ADD` the gogs directory in a separate Dockerifle. It takes about 15 minutes to build an on rpi1.

Spice-King:
I just copied Dockerfile to Dockerfile.rpi and changed the FROM directive to `armhf/alpine:3.5` and changed gosu to grab the armhf build. Took a while to build, but it built and ran on my rpi3 just fine. If you want a PR for that, I can do it up.

atomi:
I can vouch for the `armhf/alpine` images. They work on armv6 as well :+1:

Unknwon:
@atomi thanks for the info, the main problem here is continuous building and pushing the arm image, I might have time to investigate this weekend.

atomi:
Yeah if you have an rpi you could use https://github.com/drone/drone since it supports arm now but it would take a little bit of work to set up.

Unknwon:
@atomi might be overkill just for building Docker images...

Spice-King:
Someone has gotten basic QEMU emulation of the Pi working. The terminal works fine, just don't expect a desktop experience to run fast on slower hardware, have GPIO pins, or the VideoCore GPU. How you might want to chase that and build a CI system for it is up to you. And like everything else in life, there are many ways to skin this cat.

atomi:
[...]

PR #5058, opened 2018-02-24
Enable Automated builds for aarch64 ( ARM64 ) platforms

julianxhokaxhiu:
With these commits you will be able to use the Docker Hub to Automate the builds for ARM64 platforms.

As a proof, you may take a look at https://hub.docker.com/r/julianxhokaxhiu/gogs/builds/ ( aarch64 tag )

Related to #4180


Unknwon:
Hi, hmm. It is good to have automated builds, but if I'm not missing anything, you use cross build which doesn't have SQLite3 support.

julianxhokaxhiu:
At the current status I backported the Dockerfile using the concept you use to build on native platforms.

Can you please point me where is the sqlite3 missing?

Unknwon:
Based on [Build with Tags](https://gogs.io/docs/installation/install_from_source#build-with-tags) and [Install from binary](https://gogs.io/docs/installation/install_from_binary), ARM binary should have tags `sqlite`, `pam` and `cert` applied/enabled when compile.

cruz210:
Can you help enable builds

julianxhokaxhiu:
I see, then I would need to do another PR for the other file too. Can you please merge this so I can open another PR and I patch both armv7 and aarch64 :)

julianxhokaxhiu:
Actually analyzing both Dockerfile I see I'm using build scripts inside the `docker` directory.

Maybe this is something you can fix it on your own? As is nothing related to the dockerfiles themself. Or am I missing something?

PR #5068, opened 2018-02-28
Disable password autocomplete

adoreste:
As gogs can be linked with PAM it is recommended to disable autocomplete on the password input.

### DIFF
--- a/templates/user/auth/login.tmpl
+++ b/templates/user/auth/login.tmpl
@@ -15,7 +15,7 @@
 					</div>
 					<div class="required inline field {{if .Err_Password}}error{{end}}">
 						<label for="password">{{.i18n.Tr "password"}}</label>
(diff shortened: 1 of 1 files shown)


Unknwon:
Good to have that, but Chrome just doesn't care. :rofl:

rubenv:
This actually promotes worse security, as it forces people to type their password, which in case causes them to use trivial passwords. It's also highly annoying.

It's a good thing Chrome and password managers ignore it :-)

Unknwon:
:joy: fair point

adoreste:
If you have it on production servers that will have to pass security checks it is clearly an issue(That's how we realised about it). Unfortunately not everybody uses Chrome and we catch some credentials with autocomplete from browsers in our red teaming engagements. If you don't want to type your password it might be better to use a password manager.

atomi:
The autocomplete is a password manager. Bummer.

PR #5083, opened 2018-03-07
Update pkgr spec to build on Debian 9 (Stretch)

moepman:
Update pkgr specifications to enable builds on Debian 9 (Stretch) since the Debian 8 (Jessie) .deb files will not work out of the box on a clean Debian 9 install.


Unknwon:
Thank you!

Issue #2838, opened 2016-03-16
Gogs page load very slow.


wenerme:
## Problem
```
© 2016 Gogs 当前版本: 0.9.0.0306 页面: 44119ms 模板: 20ms
```

## Config
```
APP_NAME = <REMOVED>
RUN_USER = root
RUN_MODE = prod

[database]
DB_TYPE = sqlite3
HOST = 127.0.0.1:3306
NAME = gogs
USER = root
PASSWD =
SSL_MODE = disable
PATH = data/gogs.db

[repository]
ROOT = ./repositories

[server]
DOMAIN = <REMOVED>
HTTP_PORT = 3000
ROOT_URL = <REMOVED>
DISABLE_SSH = false
SSH_PORT = 3022
OFFLINE_MODE = true
ENABLE_GZIP= true
[mailer]
ENABLED = false

[service]
REGISTER_EMAIL_CONFIRM = false
ENABLE_NOTIFY_MAIL = false
DISABLE_REGISTRATION = false
ENABLE_CAPTCHA = true
REQUIRE_SIGNIN_VIEW = false

[picture]
DISABLE_GRAVATAR = true

[session]
PROVIDER = file

[log]
MODE = file
LEVEL = Info

[security]
INSTALL_LOCK = true
SECRET_KEY = <REMOVED>
```

Download from main site.

## System status

```
# cat /etc/redhat-release
CentOS release 6.5 (Final)
```

```
Tasks: 252 total,   1 running, 251 sleeping,   0 stopped,   0 zombie
Cpu(s): 25.2%us,  0.5%sy,  0.0%ni, 74.2%id,  0.0%wa,  0.0%hi,  0.0%si,  0.0%st
Mem:  10127364k total,  9907780k used,   219584k free,    82332k buffers
Swap: 16777208k total,   270980k used, 16506228k free,  7425212k cached
```

## Request log

```
[Macaron] Started GET <REMOVED> for <REMOVED>
2016/03/15 22:57:57 [D] Session ID: d7d1dad0f55f2384
2016/03/15 22:57:57 [D] CSRF Token: 7U3nCPkyzHhXpMxifg5i3Nb2rqU6MTQ1ODAwNDAyMzY2MjY0ODc0OA==
2016/03/15 22:58:41 [D] Template: repo/home
[Macaron] Completed /<REMOVED> 200 OK in 44.125287281s
[Macaron] Started GET /js/jquery-1.11.3.min.js for <REMOVED>
[...]

Issue #980, opened 2015-02-22
Very slow responses.


nd2s:
Response time is very bad here. I have to wait from multiple seconds to multiple minutes on nearly every request (a few requests return almost instantly though). System seems to be fully functional, though. Everything was slow from the very beginning ( = /install).

I don't know what part is causing the problem. I can't see anything in logs. There are no errors reported. I have the feeling it has something to do with db access, but that's pretty much a shot in the dark.

Running version with latest commit from master (d85366930c94145c8f1119ff0b6bb95d77c5200e) on FreeBSD with PostgreSQL, serving with nginx.

Any ideas what could cause that or how I can debug?

Unknwon:
Thanks for your feedback!

- is it slow when you execute git command in terminal?
- do you have virus check app or something similar?
- what's your system hardware?

nd2s:
It's running on a production server: i7-4770 haswell quad-core, 32gb ddr3 ram, 1gb/s connection. There is certainly no antivirus or virus on a FreeBSD server :).

All systems are running well, it has to be an issue with this daemon.

Unknwon:
Hmm... looks like there is no reason to be slow on this machine. :joy:

- What is the time show in the bottom of the page?
- Do you say it is slow even for the installation page?

nd2s:
Page and template times are normal. 0-35ms mostly. It takes some (long) time before I can see the request in the (stdout) logs, after that the page immediately gets processed and returned.

Yes, it was slow on installation page. When saving settings. That's why I was suggesting it could have been related to the db connection.

But nevermind, I just solved it: I checked nginx logs again after I got some gateway timeouts and saw the the localhost ipv6 address being used. Seems nginx randomly tried using ipv6 for connecting to the daemon listening on ipv4 because I used "localhost" as a hostname. I changed it to 127.0.0.1 and it's working now.

Sry for the trouble and thanks for the quick response.

Unknwon:
Wow, great! I should add this to troubleshooting, thanks you!

AysadKozanoglu:
i have also the problem that gogs web und git push are very slow (waiting of something)

on fresh debian 8.8 jessie i started ./gogs from the extracted .tar.gz than i open the regular link http://:3000

gogs needs Page: 40071ms Template: 30031ms
Gogs Version: 0.11.19.0609 Page: 40071ms Template: 30031ms

i can not solve the issue with the performance. debian and gogs are fresh installed (nothing else) all settings are by default.

[...]

Issue #914, opened 2015-02-07
Gogs painfully slow when browsing repos


Stefan-Code:
I am facing the problem that browsing repositories is dead-slow. Other stuff like browsing the dashboard, Users, Config etc. works fine but accessing the repo Overview or viewing files is painfully slow. Console output here:
<pre><code>
    C:\Go\stuff\src\github.com\gogits\gogs>gogs.exe web
    2015/02/07 05:29:00 [T] Custom path: C:/Go/stuff/src/github.com/gogits/gogs/custom
    2015/02/07 05:29:00 [T] Log path: C:/Go/stuff/src/github.com/gogits/gogs/log
    2015/02/07 05:29:00 [I] Gogs: Go Git Service 0.5.12.0202 Beta
    2015/02/07 05:29:00 [I] Log Mode: File(Trace)
    2015/02/07 05:29:00 [I] Redis Enabled
    2015/02/07 05:29:00 [I] Memcache Enabled
    2015/02/07 05:29:00 [I] Cache Service Enabled
    2015/02/07 05:29:00 [I] Session Service Enabled
    2015/02/07 05:29:08 [I] Run Mode: Production
    2015/02/07 05:29:08 [I] Listen: http://0.0.0.0:3000
    [Macaron] Started GET /Stefan/test/src/master/README.md for 127.0.0.1
    [Macaron] Completed /Stefan/test/src/master/README.md 200 OK in 19.7081272s
</code></pre>
As you can see the request to view the readme took 20 seconds! I also activated caching using redis, however nothing changed. I am using Windows 7 64bit, Go 1.4.1 and built the latest Gogs from source. 
I also had a look at the MySQL log files, and no slow queries show up there.

Unknwon:
Thanks your feedback!

What's your repository looks like? Maybe upload to https://try.gogs.io ? 

Because right now your description is like black box to us.

Thanks.

Stefan-Code:
Hi, thanks for the quick response, the repo is an empty one created with gogs and contains 3 files: README, LICENSE and .gitignore with python preset, nothing else

Unknwon:
What's time consumption at the bottom of page?

Stefan-Code:
Same as console, right now: 19583ms

Unknwon:
I see, I would still recommend you upload the repository to https://try.gogs.io

Stefan-Code:
I;m on it right now

Stefan-Code:
[...]

PR #5126, opened 2018-03-29
Default to disabling federated avatar lookup

9999years:
It's a fun feature but `ENABLE_FEDERATED_AVATAR` being true by default leads to confusion and huge slowdowns (see #2838, #980, #914) — it's unexpected that Gogs requires DNS configured (?) for avatars.  

Avatar lookup is a niche feature that I don't believe should be enabled by default given how prone it is to causing problems.

Changes `ENABLE_FEDERATED_AVATAR` in `conf/app.ini` to be `false`.


Unknwon:
Thank you!

PR #5168, opened 2018-04-13
Fix Mac OS X → macOS

Nicolapps:
This pull requests fixes the _Mac OS X_ spelling used in the docs, which was replaced by _macOS_ in 2016.


Unknwon:
Thank you!

PR #5169, opened 2018-04-16
Fixes code editor responsiveness

florianwiech:
On desktop browser clients, resizing the code editor view under 760px destroys the layout.

This issue is fixed with these less lines.

### DIFF
--- a/public/less/_repository.less
+++ b/public/less/_repository.less
@@ -422,13 +422,18 @@
 
 				table {
 					width: 100%;
+					tbody {
+						tr {
+							padding: 0 !important;
+						}
+					}
 				}
 				.lines-num {
 					vertical-align: top;
 					text-align: right;
 					color: #999;
 					background: #f5f5f5;
-					width: 1%;
+					width: 42px;
 
 					span {
 						line-height: 20px;
@@ -439,7 +444,8 @@
 				}
 				.lines-num,
 				.lines-code {
-					padding: 0;
+					display: table-cell !important;
+					padding: 0 !important;
 					pre,
 					ol,
 					.hljs {


rokuu:
I don't see any broken layout even at 1024x600. Can you include screenshots of before and after the fix? and your browser/OS info?

florianwiech:
Before and After
I hope the screenshot helps to understand the issue.
menubar_and_florianwiech_recogest__this_website_was_created_by_students_as_part_of_their_bachelor_project_module__-_gogs_and_florianwiech_recogest__this_website_was_created_by_students_as_part_of_their_bachelor_project_module__-_gogs__go_gi

Unknwon:
Thanks for the PR, however:

image

florianwiech:
According two the second point on screenshot: Yes, it is a Fix for smaller browser resolutions.
It is still handy if you have to scale down your browser window.

I would love to see this fix, but it is up to you @Unknwon :)

Unknwon:
Since the solution has came up, I'll checkout and as long as it doesn't break UI on regular resolution, I'll merge.

Unknwon:
@rokuu changes merged to `develop` branch, not yet in `master` branch yet.

rokuu:
Yeah, I think I worded my comment poorly. I manually pulled the PR locally and Gogs. Didn't do anything for me. Layout still broke.

@unknwon did you test this PR? OP tested it on 11.34 as per the screenshots. I'm not able to replicate the fix on 11.44. Did something change in between the releases?

Unknwon:
@rokuu I realized you have to generate `.css` yourself, the changes only include `.less`.

rokuu:
Generated the css too. It works :D 

But a few pages in the admin panel render weirdly under the supported resolution, here are a few examples.

screenshot from 2018-04-18 08-29-49
screenshot from 2018-04-18 08-30-28

@florianwiech got a fix for this?

florianwiech:
@rokuu thanks for checking it out!

Yes, I have the same layout breaks under the supported resolution in the admin panel. Semantic UI just breaks all their tables like that :/
A solution would be a new Pull Request ;)

rokuu:
Semantic UI has a stackable/unstackable variation for tables

https://semantic-ui.com/collections/table.html
http://jsfiddle.net/ah8hyhg2/1/

Would this work?

florianwiech:
@rokuu I think you should open a new issue for that kind of problem. Then we can discuss it there :)
(And yes, both issues are outside of the supported resolution, but I hope a new opened issue does not get closed immediately.)

rokuu:
@florianwiech I agree. Opened a new issue here https://github.com/gogits/gogs/issues/5179

Unknwon:
Checked, thank you!

PR #5181, opened 2018-04-19
Fixes overflow of brand icon hyperlink

rokuu:
rsz_1rsz_screenshot_from_2018-04-19_12-39-03

rsz_1screenshot_from_2018-04-19_12-40-28


Unknwon:
Hi, you need edit `.less` file and generate `.css`, not editing the `.css` directly.

Issue #2097, opened 2015-12-03
add support for Asciidoc as an alternative to Markdown


sluedecke:
Our projects are using Asciidoc as the language for technical documentation, including the obligatory README files in the root of a repository.  Gogs appears to support Markdown (just like github), but I'd be happy to see Asciidoc here as well (.adoc as an extension)

tboerger:
I think with https://godoc.org/github.com/VonC/asciidocgo it should be doable quite easy.

Unknwon:
@tboerger last update was 2 years ago, anyone has used that package (at least usable)?

nanoant:
:+1: but please add such thing always as a TAGGED feature, just like `sqlite` so we don't need the pull all of them by default :)

joneepenk:
no one work on this?

TrixiePenworth:
Asciidoc support would be awesome to have.

sluedecke:
Indeed it would.  Folks at  [hugo.io](https://github.com/gohugoio/hugo) could use native (go) asciidoc support as well, but as far as I know there is no native implementation in Go available right now.

There is a rejected pull request at hugo: https://github.com/gohugoio/hugo/pull/3764.  It makes use of an externally installed asciidoc toolchain.

Someone started it https://github.com/bep/asciidocgo.git, but the last commit was 2015.

HTH someone who wants to pick up the breadcrumbs.

Mizuho32:
How about using [Asciidoctor.js](https://github.com/asciidoctor/asciidoctor.js?files=1)?
It converts asciidoc text to html dynamically so maybe slow but I think it is enough for this use.

I created very rough sketch of [this](https://github.com/Mizuho32/gogs/blob/asciidoctor-test/templates/repo/view_file.tmpl) and it works.
![gogs-adoc](https://user-images.githubusercontent.com/5453161/39408270-a70b22a0-4c0e-11e8-8059-c17fd4c16e21.png)

```adoc
= Hello

== First

[red]#*test*#
http://asciidoctor.org[*Asciidoctor*]

[TIP]
====
Hoge
====
```

As you can see, it destroys styling but css scoping by ShadowDOM might help.

coreagile:
Any chance this will ever get finished? I just started playing with Gogs, and given the number of repos I have that use AsciiDoc, this would be a huge boon!

Mizuho32:
[...]

PR #5205, opened 2018-04-30
Asciidoc README support

Mizuho32:
Asciidoc format README support. https://github.com/gogits/gogs/issues/2097

gogs-adoc2

```adoc
= Hello

== First

[red]#*test*#
http://asciidoctor.org[*Asciidoctor*]

[TIP]
====
Hoge
====
```


Makefile
```diff
@@ -5,6 +5,10 @@ DATA_FILES := $(shell find conf | sed 's/ /\\ /g')
 LESS_FILES := $(wildcard public/less/gogs.less public/less/_*.less)
 GENERATED  := pkg/bindata/bindata.go public/css/gogs.css
 
+ADOCJS		 := node_modules/asciidoctor.js/dist/asciidoctor.min.js
```
Unknwon: Do not check in node_modules, only import final .min.js and .min.css under `public/plugins` directory. Look for plugins already under that directory for example.

Makefile
```diff
@@ -19,6 +23,10 @@ GOVET = go tool vet -composites=false -methods=false -structtags=false
 
 .IGNORE: public/css/gogs.css
 
+adocjsupdate: $(NODE_MODULES)
```
Unknwon: No need to update in makefile, download a specific version is all we need.

Mizuho32:
I removed asciidoctor update code from Makefile and moved asciidoctor code to `plugins/asciidoctor-1.5.7/`

pjanx:
Does this fix links to images?
```
image::picture.png[align="center"]
```
Something like [in Markdown](https://github.com/gogs/gogs/blob/master/pkg/markup/markup.go#L175). I haven't tried your changes first-hand to see if it does but it seems to me that you've omitted this handling, which would be a big shame.

By the way https://github.com/bytesparadise/libasciidoc is getting somewhere.

Mizuho32:
For the images on the web, the answer is Yes.
gogs_adoc_image

source:
```adoc
= Hello

== Asciidoctor

image::https://user-images.githubusercontent.com/79351/32431984-fb09a580-c292-11e7-9a10-553d46dba2b3.png[align="center"]

== First

[red]#*test*#
http://asciidoctor.org[*Asciidoctor*]

[TIP]
====
Hoge
====
```

The script Asciidoctor.js I used in this PR, is a browser side script and converts the adoc document to the html at the browser client side on load.
So, I don't know the detail of the markup.go and the function `wrapImgWithLink`, it doesn't need such kind of implementation and is independent of the server side implementation, languages.

For the pictures on the server local (e.g. `/home/hoge/test.png`),  it is impossible to access.

pjanx:
I'm talking about linking to pictures that are in the same repository as the viewed Asciidoc file.

Mizuho32:
If those pictures have the urls like `https://user-images.githubusercontent.com/79351/32431984-fb09a580-c292-11e7-9a10-553d46dba2b3.png`, it is possible to link but maybe you want to link the pictures in the simple, relative path style like `image::picture.png[align="center"]` so, it is not possible.

pjanx:
Okay, at least that's been answered. But it is very possible, not solving it is a bug:
* you can always post-process the output in JS for one,
* or set a default value for the `imagesdir` attribute
   
  > If the image target is an absolute path or URI, the imagesdir prefix is not added to the path.

Mizuho32:
I guess, if the pictures' ulrs have the rules like below, post-processing or attributes like `imagesdir` will make sense.
```bash
├── README.adoc
├── chakku.png
└── img
    └── poi.png
```
* `http://raw.poioichakku.com/chakku.png` for `chakku.png`
* `http://raw.poioichakku.com/img/poi.png` for `poi.png`

For the case that the pictures' ulrs are like `https://user-images.githubusercontent.com/79351/32431984-fb09a580-c292-11e7-9a10-553d46dba2b3.png`, including random string, I have no idea....

Fortunately, gogs's url rule is the first one(?), so I'll try (but I have no time to do now 😭)

Issue #5190, opened 2018-04-21
Repository description field in database should me more than 256 chars length


sergey-dryabzhinsky:
- Gogs version (or commit ref): 
  - 0.11.46.0418
- Git version: 
  - 2.17
- Operating system: 
  - Ubuntu Linux 18.04 x86 (LXC)
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL 5.7
  - [ ] MSSQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log gist (usually found in `log/gogs.log`):
`2018/04/21 19:34:50 [ERROR] [...g/context/context.go:171 ServerError()] UpdateRepository: updateRepository: update: Error 1406: Data too long for column 'description' at row 1`
## Description
If you fill description field in edit repository form with UTF-8 long string (is russian as example) and try to save it - gogs fails with 500 error.

## Suggestion:
1. do a migration that changes fields in DB like `repository`.`description` to `TEXT` type. Where it apropriate.
2. limit available length of characters in textarea form inputs (via javascript).

sergey-dryabzhinsky:
A new PR: #5219
Work in progress.

PR #5219, opened 2018-05-09
Update repository description field to contain more than 256 symbols

sergey-dryabzhinsky:
For issue #5190 

- update repository model - description field now is `TEXT` and limited by 4000 symbols
- new migration
- add description to html forms - repo creation and repo settings
- add translation for description


sergey-dryabzhinsky:
There is some questions left:
- is here migration number stored somewhere? when db is created.
- is here some way to determine of DB engine used?
- how to migrate SQLite db? Is the only way to recreate table if we change column type?

Unknwon:
What is the point to have 4000 chars in description? Why not 1024?...

sergey-dryabzhinsky:
It's not about amount of bytes.
It's just enough to describe repository.
Usually.

Since `description` stored as `TEXT` it can be 65k symbols long.
But we really don't need that much.

I just hope it will look fine at project page.

sergey-dryabzhinsky:
I tested it. Enter really long text.
And it looks ugly.

4000 symbols is too big.
512 is enough.

sergey-dryabzhinsky:
Now it looks like:

sergey-dryabzhinsky:
And new repo:

sergey-dryabzhinsky:
@Unknwon You don't mind if I add markdown support to render description on repo page?

sergey-dryabzhinsky:
And by guide I need to change only en-US locale.

Unknwon:
>You don't mind if I add markdown support to render description on repo page?

Do one thing in a PR.

sergey-dryabzhinsky:
@Unknwon Then it ready for review.

models/migrations/v18.go
```diff
@@ -0,0 +1,34 @@
...
+
+func updateRepositoryDescriptionField(x *xorm.Engine) error {
+	exist, err := x.IsTableExist("repository")
+	if err != nil {
+		return fmt.Errorf("IsTableExist: %v", err)
+	} else if !exist {
+		return nil
+	}
+	if (setting.UseMySQL) {
+		_, err = x.Exec("ALTER TABLE `repository` MODIFY `description` TEXT")
```
Unknwon: Why not change to `VARCHAR(512)`?

public/plugins/autosize-4.0.2/dist/autosize.js
```diff
@@ -0,0 +1,288 @@
+/*!
```
Unknwon: `autosize.min.js` is the only file we need, other files should not be added.

models/migrations/v18.go
sergey-dryabzhinsky: Changed to `VARCHAR(512)`.
`TEXT` used in previous version for `4000` long string.

public/plugins/autosize-4.0.2/dist/autosize.js
sergey-dryabzhinsky: Removed.

sergey-dryabzhinsky:
@Unknwon Please review.

models/migrations/v18.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2017 The Gogs Authors. All rights reserved.
```
Unknwon: s/2017/2018/

models/migrations/v18.go
```diff
@@ -0,0 +1,34 @@
...
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+package migrations
+
+import (
+	"fmt"
+
+	"github.com/go-xorm/xorm"
+	"github.com/gogs/gogs/pkg/setting"
```
Unknwon: Prepend an empty line right before this line.

models/migrations/v18.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func updateRepositoryDescriptionField(x *xorm.Engine) error {
+	exist, err := x.IsTableExist("repository")
+	if err != nil {
+		return fmt.Errorf("IsTableExist: %v", err)
+	} else if !exist {
+		return nil
+	}
+	if (setting.UseMySQL) {
```
Unknwon: You can use a `switch ... case:` struct to have cleaner code flow.

models/migrations/v18.go
sergey-dryabzhinsky: Done

models/migrations/v18.go
sergey-dryabzhinsky: Done

models/migrations/v18.go
sergey-dryabzhinsky: Done?

models/migrations/v18.go
```diff
@@ -0,0 +1,34 @@
...
+	switch {
+		case setting.UseMySQL:
+			_, err = x.Exec("ALTER TABLE `repository` MODIFY `description` VARCHAR(512);")
+		case setting.UseMSSQL:
+			_, err = x.Exec("ALTER TABLE `repository` ALTER COLUMN `description` VARCHAR(512);")
+		case setting.UsePostgreSQL:
+			_, err = x.Exec("ALTER TABLE `repository` ALTER COLUMN `description` TYPE VARCHAR(512);")
+		case setting.UseSQLite3:
+			// Sqlite3 uses TEXT field by default for any string type field.
+		default:
```
Unknwon: In general, we would return error if not matched to known list, but since it is migration and we know what we have, `default` is not needed at all. What do you think?

sergey-dryabzhinsky: Yes, it's useless here. We check all available options here.

Unknwon:
Thanks, looks like in a good state to merge in code level. I'll pull and test locally ASAP.

Unknwon:
Thank you!

Issue #4973, opened 2018-01-04
public access to issues and wiki doesn't work


JamesGKent:
currently running gogs 0.11.34 on raspbian jessie.
I have gogs configured to require the user to log on before anything is visible (Require Sign In View)  
However in this setting if a repo is set to private then other users cannot see the repo, the issue tracker or the wiki, even when public access to issue tracker and wiki is enabled

JamesGKent:
correction, they can be accessed by trying the URL, but searching through the site will not show the repo at all, this is not how i would expect this combination of settings to behave

Fastidious:
Can you replicate (as the original issue template requests) on https://try.gogs.io/, and share the link, please? Disregarding the original issue template has caused many issues to simply be closed.

JamesGKent:
its a bit more complicated than sharing a link, but as an example i have a repo here: https://try.gogs.io/JamesGKent/test  
the repo is private and public access to wifi and wiki are turned on.
if you go directly to that url without commit permissions (not logged in or as a user other than me) you should only be able to access the issues or wiki tab and not the files or pull requests tab.
that part works fine.

however try finding that repo through either the search or explore and you can't.
it's only accessible via direct link, i'm not saying that is an undesirable feature (in some cases i'd say thats preferable), however i'd argue that hiding the repo from search or explore should be a separate setting (in same location as private maybe?)

I've tried to look through the codebase to work out where the list of repos is created to try work out how to influence that behavior, but i'm new to go and so far can't seem to work out where the list is made or how the code decides which repos to include and which to leave out.

Fastidious:
@JamesGKent the name you picked for your repo makes it hard to test (no pun intended). A repo that is set as private should not show under a search, nor whilst using explore. That is by design.

I see more like a bug the ability to be able to see a private repo issues, and wiki. If you search around, I think that has been on the issues before.

JamesGKent:
you say it's by design but in the repo settings, there are options to enable the issue tracker and the wiki, and along with those options is the option to make either of those public.
these are the settings to which i'm referring:
![settings issue](https://user-images.githubusercontent.com/11315221/34577475-239c05a0-f179-11e7-8ab8-1e192cf12491.png)

as per the name, new repo here:
https://try.gogs.io/JamesGKent/PublicWikiPrivateCode

[...]

PR #5234, opened 2018-05-19
make repo show if any part is public

JamesGKent:
This pull request addresses issue #4973 

essentially if a repo is set to private and either (or both) the wiki or issue tracker are set to allow public access, they cannot be accessed without knowing the URL. The repo doesn't show in search or explore.

With these changes if the repo is private and either issue tracker or wiki are set to public then the repo will show in search and explore, but only the public pages can be accessed.

I'd suggest that maybe this behaviour should be enabled through a config option (per repo or for whole installation) but i'm not sufficiently capable with go to be able to implement that.


Unknwon:
Thank you!

PR #5239, opened 2018-05-21
fix #5229

### LINKED ISSUE #5229: Webhook not cleaned up when delete repository
models/repo.go
DeleteRepository 
漏了个 webhook

@Unknwon

naiba:
Webhook and hook_task not cleaned up when delete repository


Unknwon:
I can only merge this later because we also need to do migration which "gc"s old unlinked records in database table.

naiba:
@Unknwon wen ru gou

naiba:
@Unknwon #5277 clean unlinked webhook and hook_tasks

Unknwon:
You're are so good...

Issue #4822, opened 2017-10-11
Interpret markdown in repository descriptions


alexanderadam:
It is pretty nice like GitHub does it.

sergey-dryabzhinsky:
@alexanderadam Can you provide feedback on PR #5242 ?

alexanderadam:
done. Thank you very much. Although I probably switch to [gitea](https://github.com/go-gitea/gitea) (because of some other issues). But I don't know yet whether gitea has this feature already. Please ping me in case it does not yet and you make a PR. :wink: 

Thank you and have a wonderful weekend!

sergey-dryabzhinsky:
Well, there now simple nl2br filter for description.
Not markdown yet.
But finaly description could be not a big unreadable string.

PR #5242, opened 2018-05-24
Simple html filter for repository description

sergey-dryabzhinsky:
For issue #4822 

- add new template filter function
- use it in repo home template

### DIFF
--- a/pkg/template/template.go
+++ b/pkg/template/template.go
@@ -65,6 +65,7 @@ func NewFuncMap() []template.FuncMap {
 		"Safe":             Safe,
 		"Sanitize":         bluemonday.UGCPolicy().Sanitize,
 		"Str2html":         Str2html,
+		"Md2html":          Md2html,
 		"TimeSince":        tool.TimeSince,
 		"RawTimeSince":     tool.RawTimeSince,
 		"FileSize":         tool.FileSize,
--- a/pkg/template/template.go
+++ b/pkg/template/template.go
@@ -65,7 +65,7 @@ func NewFuncMap() []template.FuncMap {
 		"Safe":             Safe,
 		"Sanitize":         bluemonday.UGCPolicy().Sanitize,
 		"Str2html":         Str2html,
-		"nl2br":            nl2br,
+		"NewLine2br":       NewLine2br,
 		"TimeSince":        tool.TimeSince,
 		"RawTimeSince":     tool.RawTimeSince,
 		"FileSize":         tool.FileSize,
@@ -132,7 +132,7 @@ func Str2html(raw string) template.HTML {
 }
 
 // Simple filter, converts newline symbols to <br>
-func nl2br(raw string) template.HTML {
+func NewLine2br(raw string) template.HTML {
 	return template.HTML(strings.Replace(template.HTMLEscapeString(raw), "\n", "<br>", -1))
 }
 
--- a/pkg/template/template.go
+++ b/pkg/template/template.go
@@ -133,7 +133,7 @@ func Str2html(raw string) template.HTML {
 
 // Simple filter, converts newline symbols to <br>
 func NewLine2br(raw string) template.HTML {
-	return template.HTML(strings.Replace(template.HTMLEscapeString(raw), "\n", "<br>", -1))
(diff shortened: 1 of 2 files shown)


sergey-dryabzhinsky:
How it looks:

sergey-dryabzhinsky:
Simple enough.
Ready for review.

Unknwon:
😂 Is it overkill? Description is meant for at most one paragraph, otherwise it should be considered as "document".

sergey-dryabzhinsky:
I don't think it's overkill.
Sometimes we want to describe project more "structured", like:
```
This is my awesome project.
- php: 7.0
- engine: yii-1
- db: mongo
```
But now it can't even "emulate" MD - newlines will be stripped, all text will be merged in one long string.

If you have suggestions how to use here some minimal markup I'll rethink things.

Unknwon:
If the main concern is the new line, then we could just unescape `\n`s just like links, no need to make it fully markdown-able.

pkg/template/template.go
```diff
@@ -130,6 +131,10 @@ func Str2html(raw string) template.HTML {
 	return template.HTML(markup.Sanitize(raw))
 }
 
+func Md2html(raw string) template.HTML {
+	return template.HTML(markup.RawMarkdown([]byte(raw), ""))
```
MichZia: why to put a comment here

sergey-dryabzhinsky:
Now it looks like that:

alexanderadam:
[I like it](https://github.com/gogits/gogs/issues/4822) :wink:

Unknwon:
I think link and list are the only two things we should support in the repository description, anything beyond that should be put in a sophisticated document (i.e. README.md).

pkg/template/template.go
```diff
@@ -65,6 +65,7 @@ func NewFuncMap() []template.FuncMap {
 		"Safe":             Safe,
 		"Sanitize":         bluemonday.UGCPolicy().Sanitize,
 		"Str2html":         Str2html,
+		"nl2br":            nl2br,
```
Unknwon: s/nl2br/NewLine2br/ feels better?

sergey-dryabzhinsky:
Well, I can guess how to do that.
1. Make new limited sanitizer with custom policy - allow only `p, br, span, ul, ol, li`
2. Make new filter which do `str -> md -> sanitizer`

It worth a new issue I think.

Unknwon:
That's overthinking, what you have done on this PR right now is sufficient already.

sergey-dryabzhinsky:
> That's overthinking, what you have done on this PR right now is sufficient already.

I mean if we will do limited markdown filter then it will be another PR.
Not this PR.

templates/repo/home.tmpl
```diff
@@ -5,7 +5,7 @@
 		{{template "base/alert" .}}
 		{{if .PageIsRepoHome}}
 			<p id="repo-desc">
-				{{if .Repository.Description}}<span class="description has-emoji">{{.Repository.Description | Str2html}}</span>{{else}}<span class="no-description text-italic">{{.i18n.Tr "repo.no_desc"}}</span>{{end}}
```
Unknwon: You can actually use `.Repository.Description | NewLine2br | Str2html`

pkg/template/template.go
```diff
@@ -130,6 +131,11 @@ func Str2html(raw string) template.HTML {
 	return template.HTML(markup.Sanitize(raw))
 }
 
+// Simple filter, converts newline symbols to <br>
+func NewLine2br(raw string) template.HTML {
+	return template.HTML(strings.Replace(template.HTMLEscapeString(raw), "\n", "<br>", -1))
```
Unknwon: I think it would be more clear that just do `strings.Replace(raw, "\n", "<br>", -1)`

templates/repo/home.tmpl
sergey-dryabzhinsky: Done

pkg/template/template.go
sergey-dryabzhinsky: Done

Unknwon:
Thanks!

PR #5245, opened 2018-05-25
Fix avatar email update error

wei:
Updating avatar email yields "Uploaded file is not a image." error on https://try.gogs.io/user/settings/avatar.

image

Taken from go-gitea/gitea#3793 and go-gitea/gitea#3988


Unknwon:
Thank you!

PR #5246, opened 2018-05-25
semantic version not aligned

freedombone:
breaks librejs


Unknwon:
Are you sure it is 2.2.1 not 2.3.1? https://github.com/gogits/gogs/tree/master/public/js

freedombone:
Yes, because in https://try.gogs.io/assets/librejs/librejs.html it clearly says 2.2.1 but redirects to 2.2.7.

If you need to update those versions you should change it there too otherwise it breaks LibreJS.

Unknwon:
Then the PR is incomplete, "/js/semantic-2.2.1.min.js" does not exist.

PR #5260, opened 2018-06-03
Fix #1253

### LINKED ISSUE #1253: Label repos
It'd be cool if you could apply labels to repositories, and then filter repos by label, view all labels (with a button to show/hide labeled repos), etc.

Then make them cloneable as "label:<name>" (actually, idk the name restrictions on paths, so maybe this wouldn't work), which would result in a git repo with all labeled repos as submodules.

isundil:
Hello,

This pull request implements the feature #1253.
The only remaining thing to do is translating langs.

Regards,
isundil

### DIFF
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -304,6 +304,10 @@ func runWeb(c *cli.Context) error {
 			m.Get("/stars", user.Stars)
 		})
 
+		m.Group("/:username/labels", func() {
+			m.Get("/:labelID", user.Profile)
+		})
+
 		m.Get("/attachments/:uuid", func(c *context.Context) {
 			attach, err := models.GetAttachmentByUUID(c.Params(":uuid"))
 			if err != nil {
--- a/cmd/web.go
+++ b/cmd/web.go
@@ -231,6 +231,7 @@ func runWeb(c *cli.Context) error {
 			m.Get("", user.SettingsRepos)
 			m.Post("/leave", user.SettingsLeaveRepo)
 			m.Post("/labels", user.SettingsAddOrEditRepoLabel)
+			m.Post("/labels/delete", user.SettingsDeleteRepoLabel)
 		})
 		m.Group("/organizations", func() {
 			m.Get("", user.SettingsOrganizations)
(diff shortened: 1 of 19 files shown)


Unknwon:
Thanks for the PR!

Do you have screenshots?

isundil:
Here are some screenshots from the updated pages

User / org home page
2018-06-05-215222_1108x732_scrot

Repository pages
2018-06-05-215233_1081x828_scrot

Repository setting page
2018-06-05-215329_1079x857_scrot

User setting page
2018-06-05-215345_1037x545_scrot

Unknwon:
Thanks! Although color seems fancy, but I personally prefer the why GitHub "topic" does, feels more nature, how do you think?

Unknwon:
But I think it is too colorful compare to the simple color scheme of Gogs main interface.

isundil:
Yeah, I agree with you. I'll update this PR and send updated screenshot around next week.

Thanks

SoniEx2:
can you nest these? (sub-labels)

if not, it doesn't fix #1253

isundil:
No they are not nesteable as it would imply breaking change on the router. I feel pointless to have sub-labels and I don't know any system allowing this feature

SoniEx2:
gmail

SoniEx2:
on the repo, just "A/B/C".

on the label list, print it as a tree (similar to the `tree` unix command output).

Unknwon:
I personally think no real hierarchy needed, just allow `/` in the label name is sufficient enough.

SoniEx2:
Why nesting matters:

```
$ tree -dfi -L 2
.
./bitbucket
./bitbucket/eirc3
./bitbucket/eventbus
./bitbucket/libtransfer
./bitbucket/love
./bitbucket/lua
./bitbucket/openwithmenu
./bitbucket/rcmodrevive
```

```
$ tree -d -L 2
.
└── bitbucket
    ├── eirc3
    ├── eventbus
    ├── libtransfer
    ├── love
    ├── lua
    ├── openwithmenu
    └── rcmodrevive
```

isundil:
I can manage the nesting simpling by storing the full name.

The pro will be an easy search access and just a minimal change in the label list instead of seaking in a database tree ;
the cons are we would need one more query when renaming or removing a label (to check their children)

SoniEx2:
display it as a tree. and have correct removal semantics.

but for simplicity don't store it as a tree in the DB.

ibearth2025:
Please , How can I implement those new features in normal Gogs , I means the version which is not using Docker ?

PR #5315, opened 2018-06-26
require autosize when repo/settings/options is rendered

### LINKED ISSUE #5314: After failing to delete repo, can't delete without clicking to another page
- Gogs version (or commit ref): 029b33c6508b9f320300e68e38184b2eb17ee1d5 (master, right now)
- Git version: git version 2.15.2 (Apple Git-101.1)
- Operating system: macOS 10.13.5
- Database (use `[x]`):
  - [ x] PostgreSQL 9.5.2
  - [ ] MySQL
  - [ ] MSSQL
  - [ ] SQLite
  - [ x] Yes (provide example URL) https://try.gogs.io/s12chung/test-delete (you need to delete your own repo)
  - [ ] No
  - [ ] Not relevant

## Description

After failing to delete repo, can't delete without clicking to another page--clicking again and refresh failed---because the delete repo button is disabled (probably due to validation error).

delete-repo

I can try to fix.

s12chung:
https://github.com/gogs/gogs/blob/dfd494c113cfa2a25ee0c2f5a12c6403f24742a4/public/js/gogs.js#L1267 requires autosize to execute, which executes for "repo/settings/options". without this change, you get a JS error.

Technically, autoresize is only required for code with `SETTINGS_OPTIONS` rendering, but I found it kinda messy handling each case, so I guess browsers will take a minor JS loading hit. Let me know if this is good enough.

fixes #5314

### DIFF
--- a/routes/repo/setting.go
+++ b/routes/repo/setting.go
@@ -44,6 +44,7 @@ func Settings(c *context.Context) {
 func SettingsPost(c *context.Context, f form.RepoSetting) {
 	c.Title("repo.settings")
 	c.PageIs("SettingsOptions")
+	c.RequireAutosize()
 
 	repo := c.Repo.Repository
 


routes/repo/setting.go
```diff
@@ -42,6 +42,8 @@ func Settings(c *context.Context) {
 }
 
 func SettingsPost(c *context.Context, f form.RepoSetting) {
+	c.RequireAutosize()
```
Unknwon: This line should be added right after line 48 without empty line.

s12chung: done.

Unknwon:
Thank you!

PR #5319, opened 2018-06-27
[security] Add 'rel=noopener noreferrer' to <a> tags

cezar97:
Details at https://github.com/go-gitea/gitea/issues/315 and https://github.com/go-gitea/gitea/pull/4328.
Original credits to @denji .

I didn't touch any locales files as it seems to be only internal links, but I didn't digged in too much also because new locales are often added and there should be consistency...


Unknwon:
Thank you!

PR #5325, opened 2018-07-02
Adding headers to support spring config server

PhaneendraGoutham:
Spring cloud config server dependency spring cloud config monitor looks for X-Github-Event condition PropertyPathEndpoint.class -> GithubPropertyPathNotificationExtractor.class if ("push".equals(headers.getFirst("X-Github-Event"))) {...}

PropertyPathEndpoint -> notifyByPath() -> PropertyPathNotification notification = this.extractor.extract(headers, request); -> GithubPropertyPathNotificationExtractor -> extract -> has condition if ("push".equals(headers.getFirst("X-Github-Event"))) {...}

If we add these headers as part of webhook it will resolve the issue and we can integrate with spring config server.


Unknwon:
Thank you!

PR #5332, opened 2018-07-07
Apply hardening for systemd unit

khomutsky:
Disallow modifications to the system, disallow access to devices and separate temporary directory.

ProtectSystem=full # Make /boot, /etc and /usr read-only
PrivateDevices=yes # Disallow access to devices
PrivateTmp=yes # Make separate /tmp directory for gogs.


Unknwon:
Thank you!

CRCinAU:
I would suggest one more:

NoNewPrivileges=true

Unknwon:
@CRCinAU a PR would great!

CRCinAU:
I think I've done that right - first time doing a PR on github.

PR #5337, opened 2018-07-12
修正了一处描述错误

merlone:
更正了一处描述错误


Unknwon:
Thanks for your PR! However I have to close it because https://github.com/gogs/gogs/wiki/Contributing-Code#those-we-do-not-accept

PR #5338, opened 2018-07-12
Add default Authentication Switch

luhaixun:
The changes will add a default authentication switch so that user doesn't need select their authentication source every time on the login page. This will save engineers time and money.

For example, if this select action from dropdown list will cost 1s and Gogs has 1,000,000 users, it will then cost 11.6 days on switching different authentication method. If one second will cost developer 1 cent, it will then save $10,000 money!

### DIFF
--- a/models/login_source.go
+++ b/models/login_source.go
@@ -133,6 +133,7 @@ type LoginSource struct {
 	Type      LoginType
 	Name      string          `xorm:"UNIQUE"`
 	IsActived bool            `xorm:"NOT NULL DEFAULT false"`
+	IsDefault bool            `xorm:"DEFAULT NULL"` // for backward compatible
 	Cfg       core.Conversion `xorm:"TEXT"`
 
 	Created     time.Time `xorm:"-" json:"-"`
@@ -257,6 +258,11 @@ func CreateLoginSource(source *LoginSource) error {
 	}
 
 	_, err = x.Insert(source)
+	if err != nil {
+		return err
+	} else if source.IsDefault {
+		return ResetNonDefaultLoginSources(source.ID)
+	}
 	return err
 }
 
@@ -291,22 +297,32 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
 	return source, nil
 }
 
+// ResetNonDefaultLoginSources clean other default source flag
+func ResetNonDefaultLoginSources(id int64) error {
+	_, err := x.NotIn("id", []int64{id}).Cols("is_default").Update(&LoginSource{IsDefault: false})
+	return err
+}
+
 // UpdateLoginSource updates information of login source to database or local file.
 func UpdateLoginSource(source *LoginSource) error {
 	if source.LocalFile == nil {
 		_, err := x.Id(source.ID).AllCols().Update(source)
+		if err != nil && source.IsDefault {
+			return ResetNonDefaultLoginSources(source.ID)
+		}
 		return err
 	}
 
 	source.LocalFile.SetGeneral("name", source.Name)
 	source.LocalFile.SetGeneral("is_activated", com.ToStr(source.IsActived))
+	source.LocalFile.SetGeneral("is_default", com.ToStr(source.IsDefault))
 	if err := source.LocalFile.SetConfig(source.Cfg); err != nil {
 		return fmt.Errorf("LocalFile.SetConfig: %v", err)
 	} else if err = source.LocalFile.Save(); err != nil {
 		return fmt.Errorf("LocalFile.Save: %v", err)
 	}
-
 	localLoginSources.UpdateLoginSource(source)
+
 	return nil
 }
 
@@ -361,10 +377,13 @@ func (s *LocalLoginSources) ActivatedList() []*LoginSource {
 		if !s.sources[i].IsActived {
 			continue
 		}
-
 		source := &LoginSource{}
 		*source = *s.sources[i]
-		list = append(list, source)
+		if s.sources[i].IsDefault {
+			list = append([]*LoginSource{source}, list...)
+		} else {
+			list = append(list, source)
+		}
 	}
 	return list
 }
@@ -394,7 +413,8 @@ func (s *LocalLoginSources) UpdateLoginSource(source *LoginSource) {
 	for i := range s.sources {
 		if s.sources[i].ID == source.ID {
 			*s.sources[i] = *source
-			break
+		} else if source.IsDefault {
+			s.sources[i].IsDefault = false
 		}
 	}
 }
@@ -429,6 +449,7 @@ func LoadAuthSources() {
 			ID:        s.Key("id").MustInt64(),
 			Name:      s.Key("name").String(),
 			IsActived: s.Key("is_activated").MustBool(),
+			IsDefault: s.Key("is_default").MustBool(),
 			LocalFile: &AuthSourceFile{
 				abspath: fpath,
 				file:    authSource,
--- a/models/login_source.go
+++ b/models/login_source.go
@@ -410,8 +410,7 @@ func (s *LocalLoginSources) UpdateLoginSource(source *LoginSource) {
 	for i := range s.sources {
 		if s.sources[i].ID == source.ID {
 			*s.sources[i] = *source
-		} else if source.IsDefault {
-			s.sources[i].IsDefault = false
+			break
 		}
 	}
 }
--- a/models/login_source.go
+++ b/models/login_source.go
@@ -261,7 +261,7 @@ func CreateLoginSource(source *LoginSource) error {
 	if err != nil {
 		return err
 	} else if source.IsDefault {
-		return ResetNonDefaultLoginSources(source.ID)
+		return ResetNonDefaultLoginSources(source)
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


Unknwon:
Thanks for the PR!

But this does not work well if admin sets authentication source via config file. So I think the better approach is to set a default authentication source ID, if not exist then use whatever the first authentication source found.

luhaixun:
@Unknwon I added a default source radio each column to turn on the default source from the admin console. 

As far as authentication source with the config file, will it show in UI after application up? If not, we need to reconsider it, thanks for your feedback!

models/login_source.go
```diff
@@ -133,6 +133,7 @@ type LoginSource struct {
 	Type      LoginType
 	Name      string          `xorm:"UNIQUE"`
 	IsActived bool            `xorm:"NOT NULL DEFAULT false"`
+	IsDefault bool            `xorm:"DEFAULT NULL"` // for backward compatible
```
Unknwon: s/DEFAULT NULL/DEFAULT false/

and no need for extra comment

models/login_source.go
```diff
@@ -257,6 +258,11 @@ func CreateLoginSource(source *LoginSource) error {
 	}
 
 	_, err = x.Insert(source)
+	if err != nil {
+		return err
+	} else if source.IsDefault {
+		return ResetNonDefaultLoginSources(source.ID)
+	}
 	return err
```
Unknwon: In this case, you can directly `return nil` here.

models/login_source.go
```diff
@@ -291,22 +297,32 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
...
+func ResetNonDefaultLoginSources(id int64) error {
+	_, err := x.NotIn("id", []int64{id}).Cols("is_default").Update(&LoginSource{IsDefault: false})
+	return err
+}
+
 // UpdateLoginSource updates information of login source to database or local file.
 func UpdateLoginSource(source *LoginSource) error {
 	if source.LocalFile == nil {
 		_, err := x.Id(source.ID).AllCols().Update(source)
+		if err != nil && source.IsDefault {
```
Unknwon: Mixed error checking with a flag checking does not smell good to me.

models/login_source.go
```diff
@@ -361,10 +377,13 @@ func (s *LocalLoginSources) ActivatedList() []*LoginSource {
...
 		}
-
 		source := &LoginSource{}
 		*source = *s.sources[i]
-		list = append(list, source)
+		if s.sources[i].IsDefault {
+			list = append([]*LoginSource{source}, list...)
+		} else {
+			list = append(list, source)
+		}
```
Unknwon: What will be the effect if we do not make the "default" login source as the first element in the slice?

Unknwon:
>As far as authentication source with the config file, will it show in UI after application up?

Yes, they will show on the same list (in admin panel) as if you add a login source via Web UI.

models/login_source.go
luhaixun: sure

models/login_source.go
luhaixun: I will expand the logic.

models/login_source.go
Unknwon: When responding to the client, the default will be made, so no need to care about order here.

https://github.com/gogs/gogs/pull/5338/files#diff-d34865bc980845207ee9456ffb644282R116

models/login_source.go
Unknwon: I didn't see any new changes?

models/login_source.go
```diff
@@ -291,22 +297,37 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
...
+	_, err := x.NotIn("id", []int64{id}).Cols("is_default").Update(&LoginSource{IsDefault: false})
+	return err
+}
+
 // UpdateLoginSource updates information of login source to database or local file.
 func UpdateLoginSource(source *LoginSource) error {
 	if source.LocalFile == nil {
-		_, err := x.Id(source.ID).AllCols().Update(source)
-		return err
+		if _, err := x.Id(source.ID).AllCols().Update(source); err != nil {
```
Unknwon: Why error occurs, usually means the update wasn't succeed, so `ResetNonDefaultLoginSources` at this time is not a good idea. I think we should only do `ResetNonDefaultLoginSources` if `err == nil`.

luhaixun: Yes, that's a bug that should be addressed in the implementation, thanks I've included the fix in the PR.

models/login_source.go
```diff
@@ -291,22 +297,33 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
...
 	source.LocalFile.SetGeneral("is_activated", com.ToStr(source.IsActived))
+	source.LocalFile.SetGeneral("is_default", com.ToStr(source.IsDefault))
 	if err := source.LocalFile.SetConfig(source.Cfg); err != nil {
 		return fmt.Errorf("LocalFile.SetConfig: %v", err)
 	} else if err = source.LocalFile.Save(); err != nil {
 		return fmt.Errorf("LocalFile.Save: %v", err)
 	}
-
 	localLoginSources.UpdateLoginSource(source)
+
```
Unknwon: Should we also call `ResetNonDefaultLoginSources` here?

models/login_source.go
```diff
@@ -394,7 +410,8 @@ func (s *LocalLoginSources) UpdateLoginSource(source *LoginSource) {
 	for i := range s.sources {
 		if s.sources[i].ID == source.ID {
 			*s.sources[i] = *source
-			break
+		} else if source.IsDefault {
+			s.sources[i].IsDefault = false
```
Unknwon: This line actually has no effect theoretically because no DB updates in this method after this line.

models/login_source.go
```diff
@@ -291,22 +297,33 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
 	return source, nil
 }
 
+// ResetNonDefaultLoginSources clean other default source flag
+func ResetNonDefaultLoginSources(id int64) error {
+	_, err := x.NotIn("id", []int64{id}).Cols("is_default").Update(&LoginSource{IsDefault: false})
```
Unknwon: Actually, why not use `x.Where("id != ?", id)`, isn't that more intuitive?

models/login_source.go
```diff
@@ -291,22 +297,33 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
...
 	source.LocalFile.SetGeneral("name", source.Name)
 	source.LocalFile.SetGeneral("is_activated", com.ToStr(source.IsActived))
+	source.LocalFile.SetGeneral("is_default", com.ToStr(source.IsDefault))
 	if err := source.LocalFile.SetConfig(source.Cfg); err != nil {
 		return fmt.Errorf("LocalFile.SetConfig: %v", err)
 	} else if err = source.LocalFile.Save(); err != nil {
 		return fmt.Errorf("LocalFile.Save: %v", err)
 	}
-
 	localLoginSources.UpdateLoginSource(source)
```
Unknwon: Not sure why the error wasn't checked here, can you help add error-checking?

models/login_source.go
```diff
@@ -291,22 +297,53 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
 	return source, nil
 }
 
+// ResetNonDefaultLoginSources clean other default source flag
+func ResetNonDefaultLoginSources(source *LoginSource) error {
+	// update changes to DB
+	if _, err := x.NotIn("id", []int64{source.ID}).Cols("is_default").Update(&LoginSource{IsDefault: false}); err != nil {
```
Unknwon: Why not use `x.Where("id != ?", id)`, isn't that more intuitive?

models/login_source.go
```diff
@@ -291,22 +297,53 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
...
 
 	source.LocalFile.SetGeneral("name", source.Name)
 	source.LocalFile.SetGeneral("is_activated", com.ToStr(source.IsActived))
+	source.LocalFile.SetGeneral("is_default", com.ToStr(source.IsDefault))
 	if err := source.LocalFile.SetConfig(source.Cfg); err != nil {
 		return fmt.Errorf("LocalFile.SetConfig: %v", err)
 	} else if err = source.LocalFile.Save(); err != nil {
 		return fmt.Errorf("LocalFile.Save: %v", err)
 	}
+	ResetNonDefaultLoginSources(source)
```
Unknwon: Need to add error-checking here, or just `return ResetNonDefaultLoginSources(source)`

models/login_source.go
```diff
@@ -291,22 +297,53 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
...
+// ResetNonDefaultLoginSources clean other default source flag
+func ResetNonDefaultLoginSources(source *LoginSource) error {
+	// update changes to DB
+	if _, err := x.NotIn("id", []int64{source.ID}).Cols("is_default").Update(&LoginSource{IsDefault: false}); err != nil {
+		return err
+	}
+	// write changes to local authentications
+	for i := range localLoginSources.sources {
+		localSource := &LoginSource{}
+		*localSource = *localLoginSources.sources[i]
```
Unknwon: This for loop only performs read-only operations, so I think we don't need this line here, just directly use `localLoginSources.sources[i]` should be fine.

models/login_source.go
```diff
@@ -291,22 +297,53 @@ func GetLoginSourceByID(id int64) (*LoginSource, error) {
...
+		if localSource.LocalFile != nil {
+			if localSource.ID != source.ID {
```
Unknwon: Combine two ifs into `localSource.LocalFile != nil && localSource.ID != source.ID`

Unknwon:
Thanks for the work! Looks like in a good shape now.

luhaixun:
Thanks!

PR #5306, opened 2018-06-22
Add new Authentication Source: GitHub, including GitHub Enterprise.

luhaixun:
This pull request will support GitHub Authentication both for github.com and any customized GitHub Enterprise.

The related issues are:
https://github.com/gogs/gogs/issues/2927
https://github.com/gogs/gogs/issues/3033

Issue #3033, opened 2016-04-28
Support for Oauth in API


raphink:
Authentication tokens cannot be passed using standard OAuth2.

The following code returns a 401 HTTP error, but passes by adding `?token=<mytoken>` to the URL:

```
package main

import ( 
  "fmt"
  "golang.org/x/oauth2"
)

func main() { 
  ts := oauth2.StaticTokenSource(
    &oauth2.Token{AccessToken: "<mytoken>"},
  ) 
  tc := oauth2.NewClient(oauth2.NoContext, ts)

  a, _ := tc.Get("http://localhost:3000/api/v1/repos/<user>/<repo>/issues")
  fmt.Printf("%v", a)
}
```

raphink:
If this was supported, then https://github.com/google/go-github would work with gogs by just overriding the `BaseURL` property of a client.

bkcsoft:
Gogs had OAuth-support a while back but was removed since no-one wanted to maintain it. 

And no, google/go-github wouldn't work since Gogs has a slightly different json-schema :disappointed:

JohnTheodore:
@bkcsoft is there a link to the old commit for oauth2 that was working? there might be some someone interested in looking at or maintaining oauth2. I want to point them to the right resources.

stevegt:
Huh.  I need oauth so I can get out of gitlab.  Like @JohnTheodore, I'd be willing to help make it happen and keep it running.  But that "don't send pull request" label is coming across as pretty harsh.  @Unknwon, what criteria would need to be satisfied for you to accept pull requests for this feature?

stevegt:
Anyone looking for this feature may benefit from reading these two issues:
https://github.com/gogits/gogs/issues/1304
https://github.com/go-gitea/gitea/issues/26

JohnTheodore:
@stevegt thanks for the comment! go-gitea has login/authn oauth2, and commit status apis just merged. I think once you can generated scoped oauth2 tokens that can read only and update commit status, go-gitea would integrate with drone in a similar way that drone integrates with github or bitbucket. exciting!

Issue #2927, opened 2016-04-05
Github Authentication


r0kk3rz:
Github Authentication will allow people who already have a GitHub Account to create issues, edit wiki, and maybe even fork and submit pull requests on a public Gogs instance.

One of the great things about Github is the community it has, and the ease of which to contribute to existing projects once you're past the account hurdle. This feature will allow Gogs to act as a kind of distributed Github where you can get the benefits of hosting your own server without the drawbacks of losing access to the great community and low barriers to entry that hosting your project on Github gets you.

joubertredrat:
+ 1, good feature

ewoutp:
+1

Unknwon:
This is claimed to be fixed by merging #5340, please help test on `develop` branch.

PR #5340, opened 2018-07-15
 Add new Authentication Source: GitHub, including GitHub Enterprise

luhaixun:
This is a reopen of pull request #5306 as closed by mistake. The PR will support GitHub Authentication both for github.com and any customized GitHub Enterprise.

The related issues are:
#2927
~~#3033~~


cupen:
Looks good for me.

cupen:
Error log: 
2018/07/25 16:38:02 [ERROR] [...g/context/context.go:175 ServerError()] UserLogin: e-mail has been used [email: me@xxmail.com]

cupen:
How about use github oauth2.0 ?   You know, put user's password in the memory of your computer wasn't very safely.

luhaixun:
@cupen for the first email issue, that's because Gogs has already has that email linked to another account, you should change to another one and try again.

As of safety concerns, although the credential token will be encrypted before storing it in the database, I feel they share the same security risk. 

OAuth2.0 is a good idea, but it is not friendly if the same user uses different browse or clean the browser cache before login, he has to authorize it every time.

cupen:
@luhaixun Thanks for your feedback.  It looks like we have two different viewpoint.
1. security issue 
   I'm not just meaning "password store". We know, if not https, the username/password is transmitted in clear text, from user's machine to your(or him or mine) gogs machine. It's a long-range could be attacked. If the username/password is't just used by gogs, e.g.: github, google? double-kill, triple kill.  So it's not a little problem, though it's not your code's mistake. Maybe we can refactor a little code of gogs account system?

2. oauth2.0 is not friendly ? 
    Yes or no. But I tend to think the vast majority of usercases is:  one browser <--> one service. 
    e.g. I use gmail in chrome,  weixin chat in firefox, not use gmail or weixin chat both in chrome and firefox.

luhaixun:
@cupen For [1] security concern, I agree with you that it's a good chance for Gogs think more about this. 💯 

As of [2], yes we have a different view indeed because my Chrome profile is set to clear all cache every time after I close the browser, which is hard set by our Dev IT. :(

yogicodes:
It works for me!

Unknwon:
I'm merging this into a feature branch because of some required fixes/changes.

Unknwon:
>security issue
I'm not just meaning "password store". We know, if not https, the username/password is transmitted in clear text, from user's machine to your(or him or mine) gogs machine. It's a long-range could be attacked if the username/password is the third-part services account of yours. e.g.: github, google? double-kill, triple-kill. So it's not a little problem, though it's not your code's mistake. Maybe we can refactor a little code of gogs account system for adapte oauth2.0 ?

Just tested and proved that use personal access token as password with user profile permission actually works. So no password transmission at all.

Unknwon:
Merged into `develop` branch.

CodingPurple:
Getting this error

```
Unknown error: RangeError
```

Unknwon:
@CodingPurple thanks your feedback! But your feedback is as vague as said nothing. Can you provide detailed steps?

Unknwon:
@codeskyblue Hi, I cannot reproduce this error. I suspected you are using a modified version of Gogs and changed validation rule to make a "Range" binding rule to required.

luhaixun:
@codeskyblue could you help run below query in your gogs database for a quick check?
select id,type from login_source;

Issue #5342, opened 2018-07-16
The problem of branch protection


geekcomb:
- Gogs version (or commit ref): 0.11.53.0603
- Git version: 1.7.1
- Operating system: centos 6.7
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] MSSQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log gist (usually found in `log/gogs.log`):

## Description
a branch protection take a team whitelist, after save config,a new member joined the team,when this new member commit code，the operation is not successful, will prompting rejected.

## Reason
When the branch protection whitelist is saved, the whitelist of the user is calculated by the team members and saved to the protect_branch_whitelist table. This problem occurs When a team adds new member, it will not automatically add to the whitelist, so it happened.

##Modify
Modify the IsUserInProtectBranchWhitelist method of repo_branch.go file and solve the problem by dynamically judging user permissions.

PR #5343, opened 2018-07-16
The problem of branch protection#5342

geekcomb:
The problem Linked [#5342](https://github.com/gogs/gogs/pull/5342)


Unknwon:
Thanks for the PR!

However, I think correct solution is to update whitelist when change team member.

PR #5365, opened 2018-08-06
Fix open redirect vulnerability #5364

### LINKED ISSUE #5364: Insecure function `isValidRedirect` leads to open redirect vulnerability
- Gogs version (or commit ref): <= 0.11.53.0603
  - [ ] No
  - [ ] Not relevant

## Description
The function `isValidRedirect` in `gogs/routes/user/auth.go` is used in login action to validate if url is on the same site.

```
// isValidRedirect returns false if the URL does not redirect to same site.
// False: //url, http://url
// True: /url
func isValidRedirect(url string) bool {
	return len(url) >= 2 && url[0] == '/' && url[1] != '/'
}
```
If the `Location` header startswith `/\`, it will be transformed to `//` by browsers.

Check PoC [here](https://try.gogs.io/user/login?redirect_to=%2F%5Cexample.com).
PoC gif:
Imgur

A positive fix might looks like:
```
// isValidRedirect returns false if the URL does not redirect to same site.
// False: //url, http://url, /\url
// True: /url
func isValidRedirect(url string) bool {
	return len(url) >= 2 && url[0] == '/' && url[1] != '/' && url[1] != '\\'
}
```

Discoverer: bluecatli from Tencent's Xuanwu Lab

chromium1337:
Fix #5364

### DIFF
--- a/routes/user/auth.go
+++ b/routes/user/auth.go
@@ -73,10 +73,10 @@ func AutoLogin(c *context.Context) (bool, error) {
 }
 
 // isValidRedirect returns false if the URL does not redirect to same site.
-// False: //url, http://url
+// False: //url, http://url, /\url
 // True: /url
 func isValidRedirect(url string) bool {
-	return len(url) >= 2 && url[0] == '/' && url[1] != '/'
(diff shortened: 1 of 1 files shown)


Unknwon:
Thank you!

Issue #5379, opened 2018-08-16
port gogs error


tfzxyinhao:
- Gogs version (or commit ref):  0.11.53.0603
- Git version: git version 2.18.0
- Operating system: centos 7
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] MSSQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [x] No
  - [ ] Not relevant
- Log gist (usually found in `log/gogs.log`):

2018/08/15 22:13:01 [ INFO] xiaolu 0.11.53.0603
2018/08/15 22:13:01 [ INFO] Cache Service Enabled
2018/08/15 22:13:01 [ INFO] Session Service Enabled
2018/08/15 22:13:01 [ INFO] Mail Service Enabled
2018/08/15 22:13:01 [ INFO] Register Mail Service Enabled
2018/08/15 22:13:01 [ INFO] Notify Mail Service Enabled
2018/08/15 22:13:01 [FATAL] [...gs/routes/install.go:66 GlobalInit()] Fail to initialize ORM engine: sync database struct error: Error 1071: Specified key was too long; max key length is 1000 bytes

## Description
yesterday, i want to port gogs to new server and the steps is follow
1. at old server execute ./gogs backup
2. scp gogs and backup-xxx-xx.zip to new server
3.  at new server execute ./gogs restore --from=backup-xxx-xxx.zip
4. error '[FATAL] Fail to import database: fail to sync table 'User': Error 1071: Specified key was too long; max key length is 1000 bytes'

old server and new server both are centos 7, the biggest different is memory

Unknwon:
Duplicated to https://github.com/gogs/gogs/issues/179.

PR #5380, opened 2018-08-16
fix unique key too long when use myisam

tfzxyinhao:
#5379 
fix column define error


Unknwon:
Is your MySQL Version >= 5.7?

tfzxyinhao:
<5.7

Unknwon:
Well... you need at least 5.7 for charset utf8mb4

PR #5381, opened 2018-08-16
Add NoNewPrivileges=true to systemd unit file

CRCinAU:
Also add comment about unsupported options in different systemd versions.


Unknwon:
Thank you!

Issue #4996, opened 2018-01-16
Docker:  html/template:  "home" is undefined


dmlambea:
- Gogs version (or commit ref): 0.11.34.1122  (image downloaded from Docker Hub)
- Git version:
- Operating system: CoreOS Stable 1576.5.0
- Database (use `[x]`):
  - [X] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [X] No
  - [ ] Not relevant
- Log gist (usually found in `log/gogs.log`):

## Description
First run of the container image (stable, unmodified from Docker Hub) over an existing installation of Gogs (/data volume already containing more than 16000 files) makes some URLs unusable.

Error shown in browser is:  `html/template:  "home" is undefined`.  When logged in, the dashboard works but some other URLs with templates like org/create or admin/dashboard fail with the same error.

Opening a shell into the running container and killing the gogs PID makes Gogs to restart fully functional. But restarting the container from scratch reproduces the problem.

I've noticed that removing the line `chown -R git:git /data /app/gogs ~git/` from /app/gogs/docker/s6/gogs/setup makes the container fully functional in all cases. The Dockerfile used to create the functional container is:

> FROM gogs/gogs:0.11.34
> RUN sed -i 's/^chown/# chown/g' /app/gogs/docker/s6/gogs/setup

...

weisd:
got the same issues

wurst-case-scenario:
same for me

seven-tan:
same for me

seven-tan:
html/template:repo/diff/page:6:14: no such template "repo/commits_table"

argent-smith:
Have the same issue. Also, the first container log lines are:

```
05/04/2018 16:58:01usermod: no changes
05/04/2018 16:58:01init:socat  | Linked container DB will be binded to localhost on port 3306
05/04/2018 16:58:01init:socat  | Can't bind linked container GOGS_MYSQL_1 to localhost, port 3306 already in use
```

amit-handa:
any updates on this ? I am facing same issue 0.11.49.xx gogs docker image.

Unknwon:
Only happens if host is CentOS? I couldn't reproduce on Mac.

amit-handa:
yes, could be. I ran it on rhel 7, it worked. on centos 7, it didnt, so far. strangely.

[...]

PR #5390, opened 2018-08-26
Don't try to chown the git repos on startup (refs #4996).

rossigee:
As per comments in issue, the recursive 'chown' command being removed in this patch causes problems with templates when run as a container, as well as delays service startup unnecessarily for users with large/slow repos.


Unknwon:
Thank you!

PR #5393, opened 2018-08-28
Hide dashboard org creation when not allowed

jakobandersen:
The admin -> DISABLE_REGULAR_ORG_CREATION setting hides organization creation in the header, but not in the dashboard. The current button gives a 404 page.

This PR simply hides the button, using the same conditional found in the header template.


Unknwon:
Thank you!

PR #5411, opened 2018-09-10
fix aarch64 dockerfile make docker image error

leonunix:
old aarch64 dockerfile didn't work a long ago . this patch is fix this problem
fix aarch64 dockerfile make docker image error


Unknwon:
Thank you!

PR #5459, opened 2018-10-11
local in German is 'Lokal' not 'Lokalisierung' (localization)

dtrv:
A simple correction of a translation mistake into German to avoid confusion during authentication.

'Lokalisierung' means localization not 'local' - that's just 'Lokal'.


Unknwon:
Thanks for the PR!

However, I cannot merge according to https://github.com/gogs/gogs/wiki/Contributing-Code#those-we-do-not-accept.

PR #5481, opened 2018-10-27
Dockerfile.aarch64hub: install shadow package

masenf:
The shadow package contains the `usermod(8)` utility which is used by `start.sh` and `finalize.sh` to configure the git user.

The other Dockerfiles **do** install this package, for some reason it appears to have been missed in the aarch64hub variant. Further, I'm not sure that Dockerfile.aarch64hub is being used, but while trying to use it for building an aarch64 image I discovered issues with ssh login and the git user having the wrong uid/gid. Ultimately was able to debug this down to missing the usermod/groupmod utilities -- simple fix.


Unknwon:
Thank you!

Issue #5008, opened 2018-01-21
.Owner.Website with rel="me" ?


mro:
so that e.g. https://indieweb.org/ sees the back-link. Test e.g. with https://indiewebify.me/

https://github.com/gogits/gogs/blob/6f2347fc71f17b5703a9b1f383a2d3451f88b741/templates/user/profile.tmpl#L34

Or maybe `rel="me nofollow"` like github does.

What do you think?

Unknwon:
This is claimed to be fixed by merging #5485, please help test on `develop` branch or https://try.gogs.io.

federicoculloca:
Works for the given testcase.

See https://indiewebify.me/validate-rel-me/?url=https%3A%2F%2Ftry.gogs.io%2Funknwon - it found the link but then gogs.io doesn't (obviously) link back.

mro:
awesome, tested with https://try.gogs.io/issue-5008

PR #5485, opened 2018-10-29
Added rel="me" to website link on user profile

federicoculloca:
I added "me" to the value of the "rel" attribute on links pointing to a user's website, so this can be used for authentication in an IndieWeb context.

This targets issue #5008.

### DIFF
--- a/templates/user/profile.tmpl
+++ b/templates/user/profile.tmpl
@@ -31,7 +31,7 @@
 							{{if .Owner.Website}}
 								<li>
 									<i class="octicon octicon-link"></i>
-									<a target="_blank" rel="noopener noreferrer" href="{{.Owner.Website}}">{{.Owner.Website}}</a>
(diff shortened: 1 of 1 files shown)


templates/user/profile.tmpl
```diff
@@ -31,7 +31,7 @@
 							{{if .Owner.Website}}
 								<li>
 									<i class="octicon octicon-link"></i>
-									<a target="_blank" rel="noopener noreferrer" href="{{.Owner.Website}}">{{.Owner.Website}}</a>
+									<a target="_blank" rel="noopener noreferrer me" href="{{.Owner.Website}}">{{.Owner.Website}}</a>
```
Unknwon: Should we add "nofollow" as well as suggested in https://github.com/gogs/gogs/issues/5008?

federicoculloca: Sorry for the delay. Done :)

mro:
Thanks a lot, @federicoculloca !

PR #5498, opened 2018-11-05
repo/http: update CORS headers to allow custom user-agent

wmhilton:
At some point after #4970 was merged, isomorphic-git started sending a custom User-Agent in its HTTP requests to deal with some services (\*cough\* gists \*cough\*) which filtered git traffic by User-Agent. Sadly, this broke cloning from Gogs using isomorphic-git in the browser (https://github.com/isomorphic-git/isomorphic-git/issues/555). This PR fixes it by telling browsers it is OK for CORS requests to send a custom User-Agent header.


Unknwon:
Thank you!

PR #5504, opened 2018-11-07
Fix #5473 Trigger push webhook on new branch

### LINKED ISSUE #5473: Webhook not triggering when new branch pushed to mirrored repository
- Git version: 
- Operating system: 
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [ ] SQLite
  - [ ] No
  - [x] Not relevant

## Description

Webhook is not triggering after sync when new branch is pushed to mirrored repository.
I believe that this piece of code is responsible: https://github.com/gogs/gogs/blob/a1098384c09bc0d569ec82d19c26415e00cc364b/models/mirror.go#L406-L412

Expected behaviour would be to trigger webhooks on all changes. However, current implementation triggers webhook on new branches only on second push.

...

danilodemorais:
This commit fixes issue #5473 and makes a new branch behave like a
      push event and trigger the appropriate webhook.

### DIFF
--- a/models/mirror.go
+++ b/models/mirror.go
@@ -5,6 +5,7 @@
 package models
 
 import (
+	"container/list"
 	"fmt"
 	"net/url"
 	"strings"
@@ -211,7 +212,6 @@ func parseRemoteUpdateOutput(output string) []*mirrorSyncResult {
 		}
 
 		refName := lines[i][idx+3:]
-
 		switch {
 		case strings.HasPrefix(lines[i], " * "): // New reference
 			results = append(results, &mirrorSyncResult{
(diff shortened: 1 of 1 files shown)


models/mirror.go
```diff
@@ -403,14 +403,6 @@ func SyncMirrors() {
 				continue
 			}
 
-			// Create reference
-			if result.oldCommitID == GIT_SHORT_EMPTY_SHA {
-				if err = MirrorSyncCreateAction(m.Repo, result.refName); err != nil {
-					log.Error(2, "MirrorSyncCreateAction [repo_id: %d]: %v", m.RepoID, err)
-				}
-				continue
```
Unknwon: I think all we need is just to remove this single line of "continue", everything else should works automatically.

models/mirror.go
```diff
@@ -419,21 +411,51 @@ func SyncMirrors() {
...
+				if err != nil {
+					log.Error(2, "GetFullCommitID [%d]: %v", m.RepoID, err)
+					continue
+				}
+				if commit, err := gitRepo.GetCommit(refUniqueCommitID); err != nil {
+					log.Error(2, "GetCommit [repo_id: %d, commit_id: %s]: %v", m.RepoID, refUniqueCommitID, err)
+					continue
+				} else {
+					commits = list.New()
+					commits.PushFront(commit)
```
Unknwon: This only collects the latest commit, which is not always true, a branch can be created with m multiple commits. (see my another comment)

models/mirror.go
danilodemorais: If the continue is removed then the code will fail later when trying to retrieve the full commitID for GIT_SHORT_EMPTY_SHA which does not exist.

models/mirror.go
danilodemorais: I'm working on it

Issue #5502, opened 2018-11-06
Large Issue comment exceeds dashboard section


Unicorn0x1:
- Gogs version (or commit ref): 0.11.66
- Git version: 2.19.1
- Operating system: Ubuntu
- Database (use `[x]`):
  - [x] PostgreSQL
  - [x] MySQL
  - [ ] MSSQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [x] Yes (provide example URL)
  - [ ] No
  - [ ] Not relevant
- Log gist (usually found in `log/gogs.log`):

## Description

Large Issue comment(the word without space) exceeds dashboard section

![screenshot](https://user-images.githubusercontent.com/43472149/48070285-82121a00-e1fd-11e8-96d6-b2a9ae500c41.png)

Unicorn0x1:
What about `word-break: break-all;`

Unknwon:
> What about `word-break: break-all;`

Sounds good!

Unknwon:
This is claimed to be fixed by merging #5507, please help test on `develop` branch or https://try.gogs.io.

Unicorn0x1:
It's working fine!

PR #5507, opened 2018-11-08
Fix issue 5502

ajruckman:
Fixes issue #5502. I used `overflow-wrap: break-word` rather than `word-break: break-all` because it will break long words without breaking regular sentences.


Unicorn0x1:
Cool! +1

Unknwon:
Thank you!

PR #5508, opened 2018-11-08
Fix #5473 Trigger push webhook on new branch

### LINKED ISSUE #5473: Webhook not triggering when new branch pushed to mirrored repository

danilodemorais:
This PR is the continuation of the PR #5504. The reason I've opened another one is because I had branched from master and not develop, sorry for that.

This commit fixes issue #5473 and makes a new branch behave like a push event and trigger the appropriate webhook.

The solution here partially mimics the behaviour of Github when a new ref is created. Github adds to the payload of the `push` webhook the empty SHA "000000000..." for the previous commit and the latest commit for the new ref as the new commit, e.g.:
```
  "before": "0000000000000000000000000000000000000000",
  "after": "61164b65aa91cd4982e3ee8eaf108d3ebb265fd1",
```
The above was implemented in this PR. However Github also sends:
  -  any commits that were made between the newest commit of the new ref and the last commit of its ancestor;
  -  or an empty array of commits if there was no commits in the new ref.

I don't have time at the moment to implement the second part above (a possible solution is in https://stackoverflow.com/questions/3161204/find-the-parent-branch-of-a-git-branch), however I found in another part of the code, in the file `models/update.go` line 118, that Gogs simply returns the latest 10 commits for a `newRef` when it finds one in its own repositories.
Therefore, I implemented the same behaviour for the mirror when a `newRef` is found. With or without new commits, the webhook will return the latest 10 commits for the new ref. I think this is a good compromise at the moment because it doesn't break normal CICD pipelines (that should run when a new branch is pushed) and makes Gogs' behaviour one step closer to Github's.

### DIFF
--- a/models/mirror.go
+++ b/models/mirror.go
@@ -5,6 +5,7 @@
 package models
 
 import (
+	"container/list"
 	"fmt"
 	"net/url"
 	"strings"
@@ -211,7 +212,6 @@ func parseRemoteUpdateOutput(output string) []*mirrorSyncResult {
 		}
 
 		refName := lines[i][idx+3:]
-
 		switch {
 		case strings.HasPrefix(lines[i], " * "): // New reference
 			results = append(results, &mirrorSyncResult{
@@ -403,14 +403,6 @@ func SyncMirrors() {
 				continue
 			}
 
-			// Create reference
-			if result.oldCommitID == GIT_SHORT_EMPTY_SHA {
-				if err = MirrorSyncCreateAction(m.Repo, result.refName); err != nil {
-					log.Error(2, "MirrorSyncCreateAction [repo_id: %d]: %v", m.RepoID, err)
-				}
-				continue
-			}
-
 			// Delete reference
 			if result.newCommitID == GIT_SHORT_EMPTY_SHA {
 				if err = MirrorSyncDeleteAction(m.Repo, result.refName); err != nil {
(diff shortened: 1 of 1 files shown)


models/mirror.go
```diff
@@ -403,14 +403,6 @@ func SyncMirrors() {
 				continue
 			}
 
-			// Create reference
-			if result.oldCommitID == GIT_SHORT_EMPTY_SHA {
-				if err = MirrorSyncCreateAction(m.Repo, result.refName); err != nil {
-					log.Error(2, "MirrorSyncCreateAction [repo_id: %d]: %v", m.RepoID, err)
-				}
-				continue
```
danilodemorais: We can't just remove the continue, the code is going to break later if oldCommitID doesn't in the repo.

danilodemorais:
Ping @Unknwon

Unknwon:
Tested, thank you!

PR #5511, opened 2018-11-12
update Dockerfile

ehlxr:
1、Compress the size of the gogs binary
2、Add time zone ENV


Unknwon:
Thanks for the PR!

What is the purpose of compress size here? The image is already small enough.

PR #5521, opened 2018-11-17
Implement issue title and content prefilling

### LINKED ISSUE #5302: Allow prefilling the title and body of new issues using query parameters
- Git version: not relevant
- Operating system: Debian
- Gogs version (or commit ref): `0.11.29.0727 `
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] MSSQL
  - [ ] SQLite
  - [x] Yes (provide example URL): https://try.gogs.io/Radiergummi/test-directory-slashes/issues/new?title=foo&body=bar
  - [ ] No
  - [ ] Not relevant
not relevant 

## Description
On Github, you can set the issue title and content using query parameters, like so:

https://github.com/gogs/gogs/issues/new?title=foo&body=bar

If you open that URL, you'll land at the new issue page with the issue title set to *foo* and the body to *bar*. This is especially helpful for issue reporting links that I'd like to implement in my applications development mode (Exception messages will contain a link to the new issue page, prefilled with all relevant debugging details).  
Would it be possible to integrate this into Gogs? *Maybe* one could even go further and allow to set labels, milestones and assignments this way.

Charliekenney23:
Resolves #5302.

Just added some code to the `GET /:user/:repo/issues/new` route to add `title` and `content` query string parameters to the context, so that they are prefilled in the from template when provided. The feature request mentioned using `body`, but `content` was already used in the template file so I went with that.

One thing to note is, with the current setup in [templates/repo/issue/comment_tab.tmpl](https://github.com/gogs/gogs/blob/master/templates/repo/issue/comment_tab.tmpl#L8), a repository's issue template takes precedence over the provided `content` query string parameter when templating the issue body, contrary to how it works on Github. @Unknwon Interested to see what your take on this is.

### DIFF
--- a/routes/repo/issue.go
+++ b/routes/repo/issue.go
@@ -345,6 +345,8 @@ func NewIssue(c *context.Context) {
 	c.Data["PageIsIssueList"] = true
 	c.Data["RequireHighlightJS"] = true
 	c.Data["RequireSimpleMDE"] = true
+	c.Data["title"] = c.Query("title")
+	c.Data["content"] = c.Query("content")
 	setTemplateIfExists(c, ISSUE_TEMPLATE_KEY, IssueTemplateCandidates)
 	renderAttachmentSettings(c)
 


Unknwon:
Thanks for the PR!

I think current is fine until someone explicitly asks for other options.

PR #5534, opened 2018-11-29
Include org in new mirror url

sj26:
The organisation dashboard page includes a _Mirror_ tab which has a _New Mirror_ (+) button. At the moment clicking this button goes to the _New Migration_ page with _This repository will be a mirror_ checked, but doesn't select the previously viewed organisation. This change makes it work like the _New Repository_ button from the same dashboard page which does select the previously viewed organisation.

screen shot 2018-11-29 at 9 04 48 pm


Unknwon:
Thank you!

PR #5544, opened 2018-12-06
Update user.go

mvrahden:
## Problem
failed to receive OS-Username on some docker OS before

## Description
updates the method to retrieve the USER-name.

## What changed?
This version initially triggers the native user-object and if this fails queries the OS-Env. Instead of relying on a solution purely build on OS-Environment variables.

### DIFF
--- a/pkg/user/user.go
+++ b/pkg/user/user.go
@@ -9,7 +9,17 @@ import (
 )
 
 func CurrentUsername() string {
-	curUserName := os.Getenv("USER")
+	user, err := user.Current()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	curUserName := user.Username
+	if len(curUserName) > 0 {
+		return curUserName
+	}
+
+	curUserName = os.Getenv("USER")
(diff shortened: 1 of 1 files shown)


Unknwon:
`user.Current()` uses CGO, and has problem with cross-compilation.

Unknwon:
What we found was we can build, but when try to start in target platform (e.g. build on macOS and target is linux amd64), it says something like "not implemented" (if I remember correctly).

mvrahden:
Ok, I understand! Thank you very much for the answer :) 💯

FlyingOnion:
Sorry for adding comment after so long. I have the same problem, and my solution is to add a environment `USER=git` when I launch a container. I know `git` may be the default user of gogs for the time being, but are there any better solutions if custom users are added for gogs?

unknwon:
> I have the same problem, and my solution is to add a environment `USER=git` when I launch a container.

@FlyingOnion: Not sure what you're asking exactly... but this looks correct to me :D

PR #5568, opened 2018-12-24
Support plaintext alt for HTML emails

guysmoilov:
This makes the sent messages less likely to be thrown away by spam filters, even if you DO want to be able to send HTML messages in general.

I know there's no opened issue or request for this, but I did it for myself and thought it would be nice to give you the chance to also add this very small feature that could potentially help people later on.

### DIFF
--- a/pkg/mailer/mailer.go
+++ b/pkg/mailer/mailer.go
@@ -39,7 +39,7 @@ func NewMessageFrom(to []string, from, subject, htmlBody string) *Message {
 
 	contentType := "text/html"
 	body := htmlBody
-	if setting.MailService.UsePlainText {
+	if setting.MailService.UsePlainText || setting.MailService.AddPlainTextAlt {
 		plainBody, err := html2text.FromString(htmlBody)
 		if err != nil {
 			log.Error(2, "html2text.FromString: %v", err)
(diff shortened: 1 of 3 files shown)


Unknwon:
Thanks for the PR!

1. The title of this PR looks confusing to me after reading the code, can you rephrase it?
2. Please an a sample option after this line:

https://github.com/gogs/gogs/blob/1f11c1f71a314a85ab09b75f10a527a351020144/conf/app.ini#L251-L252

Unknwon:
Basically, to answer what does this PR actually do?

pkg/mailer/mailer.go
```diff
@@ -39,16 +39,21 @@ func NewMessageFrom(to []string, from, subject, htmlBody string) *Message {
...
 			log.Error(2, "html2text.FromString: %v", err)
 		} else {
 			contentType = "text/plain"
 			body = plainBody
+			switchedToPlaintext = true
 		}
 	}
 	msg.SetBody(contentType, body)
+	if switchedToPlaintext && setting.MailService.AddPlainTextAlt && !setting.MailService.UsePlainText {
+		msg.AddAlternative("text/html", htmlBody)
```
Unknwon: From what I read on this line, seems like it is "Support HTML alt for plaintext emails"?

guysmoilov: That's a technical detail of how to setup the MIME message, but semantically it's the other way around. Mail clients will show the HTML as first priority, and the plaintext is just a backup.

guysmoilov: According to https://godoc.org/gopkg.in/gomail.v2#Message.AddAlternative

guysmoilov: I'll add a code comment

Unknwon:
LGTM, thanks again!

PR #5584, opened 2019-01-06
README: add LunaNode cloud deployment URL

uakfdotb:
I developed a simple Gogs launcher to allow our users to easily deploy Gogs on our cloud platform at https://github.com/LunaNode/launchgogs

It sets up Gogs via the Docker package and adds an HTTPS-enabled nginx reverse proxy. We're displaying the launcher internally but I noticed the Deploy to Cloud section and it'd be great if it can be added there as well!


Unknwon:
Thank you!

PR #5612, opened 2019-02-04
Require a logged in admin for macaron toolbox access

guysmoilov:
Possible fix for: https://github.com/gogs/gogs/issues/5578


guysmoilov:
@Unknwon I updated the PR to exclude the healthcheck, and it seems to work fine, allows admin only to use /debug/pprof while still allowing /healthcheck.
Is there another problem?

Unknwon:
Thank for the update!

Although it should work (supposed you've tested on you end), but current implementation feels somewhat a quick hack. I need sometime to evaluate a solution.

PR #5620, opened 2019-02-15
Removes ticks around the path to the ini file. Fixes #5619

### LINKED ISSUE #5619: Incompatible authorized_keys on windows using PowerShell as SSH default shell
**Describe the bug**
I have installed GOGS as a service on Windows 2012 Server R2 using OpenSSH as an external SSH server.
SSHD is configured to use PowerShell as shell.
After registering the Public SSH Key, I get an error checking out any repository  via SSH with:

```
$ git pull --rebase
Gogs: Internal error
Fail to get repository owner 'username': Error 1045: Access denied for user 'root'@'localhost' (using password: NO)
fatal: Could not read from remote repository.
```
Connecting directly with SSH shows me the error:
```
PTY allocation request failed on channel 0
2019/02/14 16:13:15 [ WARN] Custom config ''C:/gogs/custom/conf/app.ini'' not found, ignore this if you're running first time
Hi there, You've successfully authenticated, but Gogs does not provide shell access.
If this is unexpected, please log in with password and setup Gogs under another user.
Connection to xxx.xxx.xxx.xxx closed.
```

I tracked it down to the entries inside authorized_keys which are not compatible on windows.

This entry is generated and does *not* work:
`command="C:/gogs/gogs.exe serv key-3 --config='C:/gogs/custom/conf/app.ini'", ...`

**Fix**: Removing the single tick quotes get it to work:
`command="C:/gogs/gogs.exe serv key-3 --config=C:/gogs/custom/conf/app.ini", ...`

After this, the repository could be checked out.

**To Reproduce**
- Install on Windows 2012 R2 with external openssh
- Try checking out with ssh

**Expected behavior**
Instead of the error above, the repository should be checked out.

**Additional context**

- Gogs version (or commit ref): 0.11.86.0130
- Git version: 2.20.1.windows.1
- SSH version: OpenSSH_for_Windows_7.9p1, LibreSSL 2.6.5 Shell: **PowerShell**
- Operating system: WIndows 2012 R2
- Database:
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [x] SQLite
  - [ ] Yes: provide example URL
  - [x] No:  This is SSH specific and therefore cannot reproduced on that site
  There are no relevant log entries created.

Kretikus:
This pull requests fixes, the problem described in #5619.

On Windows the ticks prevent gogs to find the INI file.

I do not know how it behaves having spaces in the file name.

### DIFF
--- a/models/ssh_key.go
+++ b/models/ssh_key.go
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	_TPL_PUBLICK_KEY = `command="%s serv key-%d --config='%s'",no-port-forwarding,no-X11-forwarding,no-agent-forwarding,no-pty %s` + "\n"
+	_TPL_PUBLICK_KEY = `command="%s serv key-%d --config=%s",no-port-forwarding,no-X11-forwarding,no-agent-forwarding,no-pty %s` + "\n"
 )
 
 var sshOpLocker sync.Mutex


Unknwon:
Thanks for the PR! But this would definitely breaks non-Windows systems (if removing ticks would work in Windows) if the file path contains spaces.

Kretikus:
Hi!
The OpenSSH Server is configured to use PpwerShell as default shell...  It might be the underlying issue. I will try to verify that.

Kretikus:
Hi!
I have confirmed that the problem was PowerShell. I suggest changing the documentation to mention SSH needs to be configured to use bash as shell

Unknwon:
Docs updated as suggested: https://gogs.io/docs/installation

PR #5628, opened 2019-02-20
Add assignee to issue emails

### LINKED ISSUE #4220: When a new issue assigned or any comments coming, the assignee should also be emailed.
When a new issue assigned or any comments coming, the assignee should also be emailed.

before https://github.com/gogits/gogs/blob/master/models/issue_mail.go#L123 , add following code:
```
//keep the Assignee as the same role as the watchers
	if issue.Assignee != nil {
		if com.IsSliceContainsStr(names, issue.Assignee.Name) == false {
			tos = append(tos, issue.Assignee.Email)
			names = append(names, issue.Assignee.Name)
		}
	}
```

any mistake or omission?

hochhaus:
Fixes #4220


Unknwon:
Thank you!

PR #5658, opened 2019-03-23
make "ENABLE_RAW_FILE_RENDER_MODE = true" workable!

xhawk18:
It will still returns response "text/plain; charset=utf-8" even after ENABLE_RAW_FILE_RENDER_MODE = true was set in app.ini when access raw file. 
With ?render=1 append in the raw url get the same response.

So I make a patch here to send correct mime-type, and remove "render=1".

Tested and built by Termux golang.

### DIFF
--- a/routes/repo/download.go
+++ b/routes/repo/download.go
@@ -7,13 +7,169 @@ package repo
 import (
 	"io"
 	"path"
-
+	"strings"
 	"github.com/gogs/git-module"
 
 	"github.com/gogs/gogs/pkg/context"
 	"github.com/gogs/gogs/pkg/setting"
 	"github.com/gogs/gogs/pkg/tool"
 )
+
+var mimetypes map[string]string = map[string]string{
+	".epub": "application/epub+zip",
+	".fif": "application/fractals",
+	".spl": "application/futuresplash",
+	".hta": "application/hta",
+	".hqx": "application/mac-binhex40",
+	".vsi": "application/ms-vsi",
+	".accdb": "application/msaccess",
+	".accda": "application/msaccess.addin",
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for the PR!

>It will still returns response "text/plain; charset=utf-8" even after ENABLE_RAW_FILE_RENDER_MODE = true was set in app.ini when access raw file.
With ?render=1 append in the raw url get the same response.

You need to prove this.

>So I make a patch here to send correct mime-type, and remove "render=1".

`?render=1` exists for the reason that you do not want unexpected rendering of possible malicious content.

xhawk18:
> You need to prove this.

?render=1 或 ?render=0 或其他值，均不起作用，故移除之。

证明办法，配置文件加 ENABLE_RAW_FILE_RENDER_MODE = true，访问一个html文件的raw file，返回均为text。

Unknwon:
Thanks for the follow up but that is not a prove. That's another statement of the problem.

xhawk18:
这是一个bug，一个事实，你自己跑一下就知道了，证明个啥。
不提了，爱改不改。。

我的pull request收回。请不要合并本人劳动成果。

xhawk18:
我为维护组这种对待bug的态度感到羞愧！

Unknwon:
https://try.gogs.io/unknwon/hello/raw/master/html/index.html?render=1

image

xhawk18:
fianlly i would like clarify why my test failed to show html files.

I have a mistake that fill "ENABLE_RAW_FILE_RENDER_MODE = true“ in file "app/conf.ini",
while the correct configure file path is "custom/conf/app.ini".

Thanks the author of gogs to do the great job!

PR #5671, opened 2019-04-08
templates/base/head: Add meta viewport for iPad, etc.

Xe:
https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag

Without this patch:

5798D602-CDD4-4564-A2AD-C81E62CC4F1C

With this patch:

4205E420-2527-4ED7-8FC1-4E56D45D46A3

### DIFF
--- a/templates/base/head.tmpl
+++ b/templates/base/head.tmpl
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head data-suburl="{{AppSubURL}}">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks for the PR!

But what is the real difference other than the scale?

Xe:
That's the difference. It makes the site not giant. It hurts my eyes to have the site so big.

Unknwon:
I see. I think you should use head template inject: https://discuss.gogs.io/t/how-to-inject-custom-head-and-footer/943

Xe:
I'm dissatisfied with this result.

PR #5689, opened 2019-04-19
Include the Sha in webhook create payloads.

Beanow:
This goes along with https://github.com/gogs/go-gogs-client/issues/103 and https://github.com/gogs/go-gogs-client/pull/104

For Drone CI integrations, having the commit Sha in a create (tag) webhook event is being relied on for docker builds. See drone-plugins/drone-docker#229

This adds it to the payload.

Git tea does make this available, visible here https://github.com/go-gitea/go-sdk/blob/2858b80da5f75fa2eea3fcf3205a5e6fd16da0ec/gitea/hook.go#L196

### DIFF
--- a/models/action.go
+++ b/models/action.go
@@ -604,6 +604,7 @@ func CommitRepoAction(opts CommitRepoActionOptions) error {
 		if err = PrepareWebhooks(repo, HOOK_EVENT_CREATE, &api.CreatePayload{
 			Ref:           refName,
 			RefType:       "tag",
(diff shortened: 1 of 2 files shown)


Unknwon:
Thank you!

theWayToAI:
It seems like that this issue is still not in the master and not in the latest tag?

unknwon:
> It seems like that this issue is still not in the master and not in the latest tag?

I don't think it is not. Why?

PR #5724, opened 2019-06-14
docker: prevent Gogs from running a recursive chown on each boot

spieker:
The `chown` command in the setup file of Gogs is supposed to be executed just once. As soon as the app.ini file exists, this step should be skipped. However, the test was checking if a directory with the name app.ini exists instead of a file. Therefore, the `chown` command was executed on each boot.


Unknwon:
Thank you!

PR #5750, opened 2019-07-16
pkg/context: Render live notice banner from file

achilleas-k:
As discussed already on the forum (https://discuss.gogs.io/t/contributing-new-features-from-fork/2645), this PR adds a live notice banner to all GOGS pages when a notice file is found in the custom configuration directory.

In summary:
- Contexter checks if there is a file called `notice` under the `GOGS_CUSTOM` directory and loads it.
- The first line is treated as a header/title and everything else as the message body.
- Message body is rendered as HTML (tags allowed).
- File size is limited to 1024 bytes.
- File mime type must be text.
- Notice is rendered in `head.tmpl` for all pages.

I'm not sure whether it would be best for certain aspects of this to be configurable.  Perhaps the location and name of the file could be an optional setting (`$GOGS_CUSTOM/notice` by default), as well as the max file size.

Let me know

The screenshots below show how the following text file is rendered:
```
Upcoming scheduled downtime

<p>This <b>Friday (19 July, 2019)</b> there will be a scheduled service
interruption for maintenance.  The service will be unavailable starting
<b>17:00 UTC</b> and we expect the downtime to last approximately <b>2
hours</b>.</p>

<p>Please refer to the <a href=announcement.html>announcements page</a> for
more details. </p>

```

front-page
login-page
repository-view

### DIFF
--- a/pkg/context/notice.go
+++ b/pkg/context/notice.go
@@ -20,7 +20,6 @@ import (
 // on all pages.
 func readServerNotice() map[string]interface{} {
 	fpath := path.Join(setting.CustomPath, "notice")
-
 	if !com.IsExist(fpath) {
 		return nil
 	}
--- a/pkg/context/notice.go
+++ b/pkg/context/notice.go
@@ -7,34 +7,34 @@ package context
 import (
 	"os"
 	"path"
-	"strings"
 
 	"github.com/Unknwon/com"
 	log "gopkg.in/clog.v1"
 
+	"github.com/gogs/gogs/pkg/markup"
 	"github.com/gogs/gogs/pkg/setting"
 	"github.com/gogs/gogs/pkg/tool"
 )
 
 // readServerNotice checks if a notice file exists and loads the message to display
 // on all pages.
-func readServerNotice() map[string]interface{} {
-	fpath := path.Join(setting.CustomPath, "notice")
+func readServerNotice() string {
+	fpath := path.Join(setting.CustomPath, "notice.md")
 	if !com.IsExist(fpath) {
-		return nil
+		return ""
 	}
 
 	f, err := os.Open(fpath)
 	if err != nil {
 		log.Error(2, "Failed to open notice file %s: %v", fpath, err)
-		return nil
+		return ""
 	}
 	defer f.Close()
 
 	fi, err := f.Stat()
 	if err != nil {
 		log.Error(2, "Failed to stat notice file %s: %v", fpath, err)
-		return nil
+		return ""
 	}
 
 	// Limit size to prevent very large messages from breaking pages
@@ -42,25 +42,23 @@ func readServerNotice() map[string]interface{} {
 
 	if fi.Size() > maxSize { // Refuse to print very long messages
 		log.Error(2, "Notice file %s size too large [%d > %d]: refusing to render", fpath, fi.Size(), maxSize)
-		return nil
+		return ""
 	}
 
 	buf := make([]byte, maxSize)
 	n, err := f.Read(buf)
 	if err != nil {
 		log.Error(2, "Failed to read notice file: %v", err)
-		return nil
+		return ""
 	}
 	buf = buf[:n]
 
 	if !tool.IsTextFile(buf) {
 		log.Error(2, "Notice file %s does not appear to be a text file: aborting", fpath)
-		return nil
+		return ""
 	}
 
-	noticetext := strings.SplitN(string(buf), "\n", 2)
-	return map[string]interface{}{
-		"Title":   noticetext[0],
-		"Message": noticetext[1],
-	}
+	// noticetext := strings.SplitN(string(buf), "\n", 2)
+
+	return string(markup.RawMarkdown(buf, ""))
 }
--- a/pkg/context/notice.go
+++ b/pkg/context/notice.go
@@ -1,0 +1,63 @@
+package context
+
+import (
+	"os"
+	"path"
+	"strings"
+
+	"github.com/Unknwon/com"
+	"github.com/gogs/gogs/pkg/setting"
+	"github.com/gogs/gogs/pkg/tool"
+	log "gopkg.in/clog.v1"
+)
+
+// readServerNotice checks if a notice file exists and loads the message to display
+// on all pages.
+func readServerNotice() map[string]interface{} {
+	fileloc := path.Join(setting.CustomPath, "notice")
+
+	// Limit size to prevent very large messages from breaking pages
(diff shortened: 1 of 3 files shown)


unknwon:
Thanks! Will review ASAP.

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
+package context
```
unknwon:
```suggestion
// Copyright 2019 The Gogs Authors. All rights reserved.
// Use of this source code is governed by a MIT-style
// license that can be found in the LICENSE file.

package context
```

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"os"
+	"path"
+	"strings"
+
+	"github.com/Unknwon/com"
+	"github.com/gogs/gogs/pkg/setting"
+	"github.com/gogs/gogs/pkg/tool"
+	log "gopkg.in/clog.v1"
```
unknwon: Import path should be grouped as follows:

```go
	"github.com/Unknwon/com"
	log "gopkg.in/clog.v1"
	
	"github.com/gogs/gogs/pkg/setting"
	"github.com/gogs/gogs/pkg/tool"
```

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// readServerNotice checks if a notice file exists and loads the message to display
+// on all pages.
+func readServerNotice() map[string]interface{} {
+	fileloc := path.Join(setting.CustomPath, "notice")
```
unknwon: Should we consider add a subdirectory under custom path i.e. `notices`, and currently only use the file `notices/main.html`?

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Limit size to prevent very large messages from breaking pages
+	var maxlen int64 = 1024
```
unknwon:
```suggestion
	var maxSize int64 = 1024
```

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if !com.IsExist(fileloc) {
```
unknwon: It would be better if we check existence of the file before define variable `maxSize`, because `maxSize` is not used until we `.Stat()` it.

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil
+	}
+
+	log.Trace("Found notice file")
```
unknwon: This could be too verbose because it prints on every page request, what about move this detection line at start of Gogs? Or simply remove it?

pkg/context/notice.go
unknwon:
```suggestion
	fpath := path.Join(setting.CustomPath, "notice")
```

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
...
+	fp, err := os.Open(fileloc)
```
unknwon:
```suggestion
	f, err := os.Open(fileloc)
```

pkg/context/notice.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		log.Error(2, "Failed to open notice file %s: %v", fileloc, err)
+		return nil
+	}
+	defer fp.Close()
+
+	finfo, err := fp.Stat()
```
unknwon:
```suggestion
	fi, err := fp.Stat()
```

pkg/context/notice.go
achilleas-k: I suppose that would be good.  Keeps the possibility open of adding more types of notices, or notices for specific pages.  Would using an html file be good?  Currently, the body text is rendered as html, but the header line is not, so it's not completely html content.

Just felt I should clarify that point, but I can change it to your suggestion if you want.

unknwon: >Currently, the body text is rendered as html, but the header line is not, so it's not completely html content.

You're right, actually I was thinking adding things like `title: My sweet title`, so we could possibly add more attributes in the future. And separate title and content by two line breaks (as right now, as how Git commit message works). What do you think?

pkg/context/notice.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/gogs/gogs/pkg/setting"
+	"github.com/gogs/gogs/pkg/tool"
+)
+
+// readServerNotice checks if a notice file exists and loads the message to display
+// on all pages.
+func readServerNotice() map[string]interface{} {
+	fpath := path.Join(setting.CustomPath, "notice")
+
```
unknwon:
```suggestion

```

pkg/context/notice.go
achilleas-k: The commit message convention is exactly what I was thinking about too.  The code simply breaks off the first line and takes the rest as body text, but when I write my own notice files I always leave an empty line, so I agree it would be cleaner that way.

Another idea might be to render it as markdown (and force the file to be called `notice.md` to make it clear), so a short notice with a title could be:
```markdown
# Server downtime

We will be taking the server **offline** on Sunday.
```

unknwon: Markdown idea is actually brilliant! Would you mind update PR to address this?

Helpful link: https://github.com/gogs/gogs/blob/798636c95b207fc97a785ce545d7af7197361447/pkg/markup/markdown.go#L120

achilleas-k:
Updated the PR to have it render the notice as markdown.  Attached screenshot shows how the following text is rendered:
```markdown
## Upcoming scheduled downtime

-----

This **Friday (25 July, 2019)** there will be a scheduled service
interruption for maintenance.  The service will be unavailable starting
**17:00 UTC** and we expect the downtime to last approximately **2 hours**.

Please refer to the [announcements page](announcement.html) for more details.
```

image

unknwon:
Thanks again! 

I'm merging this into a tmp branch since I'm going to make some changes based on this PR.

Will post update on this thread once finished.

unknwon:
Just one small change to the PR:

https://github.com/gogs/gogs/commit/d775fe793605757e6f51057da9913ad5a11f6336#diff-9daf9f93bd40cc3b193d86984af1ac65R22

The location of the notice banner file.

achilleas-k:
Thanks for merging.  Changes look good, much cleaner than my version (making it a Context method).  One issue:
https://github.com/gogs/gogs/commit/d775fe793605757e6f51057da9913ad5a11f6336#r34474619

unknwon:
Pushed :D

PR #5760, opened 2019-07-25
Fix and update packager.io build files

half-duplex:
This PR fixes packager.io builds based on this repo, removes Debian 7 and Ubuntu 12.04 which are out of support, and adds Debian 10 and Ubuntu 18.04.
Builds are available at https://packager.io/gh/half-duplex/gogs/refs/develop - I don't have machines other than Debian 10 available to test, but I haven't had any problems there.

The existing deb repo, built from https://github.com/pkgr/gogs/tree/pkgr, is very out-of-date, see https://github.com/gogs/docs/pull/232. Is there any interest in just enabling packager.io for this repo directly? It's a couple clicks by a maintainer at https://packager.io/apps/new, and maybe changing branches to `master,develop` to offer pre-release builds.

Possible concerns:
- Are there any other officially supported heroku-related things that the vendor.json change may affect?
- packager.io seems to build the latest commit but label it as the latest tag - is this behavior appropriate on master, or does the specific tag commit need to be put in a dedicated repo/branch for it to grab?
- I'm not sure what the previously-added bzr and mercurial build dependencies are for, if cruft removal is desired I can spin up some VMs to test

### DIFF
--- a/.pkgr.yml
+++ b/.pkgr.yml
@@ -1,15 +1,13 @@
 targets:
-  debian-7: &debian
+  debian-8: &debian
     build_dependencies:
       - libpam0g-dev
     dependencies:
       - libpam0g
       - git
-  debian-8:
-    <<: *debian
   debian-9:
     <<: *debian
-  ubuntu-12.04:
+  debian-10:
     <<: *debian
   ubuntu-14.04:
     <<: *debian
@@ -18,6 +16,8 @@ targets:
     build_dependencies:
       - bzr
       - mercurial
+  ubuntu-18.04:
+    <<: *debian
   centos-6: &el
     build_dependencies:
       - pam-devel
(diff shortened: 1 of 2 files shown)


unknwon:
Thanks for the PR!

Though I am not 100% sure what will happen, but changes look fine to me.

half-duplex:
:+1::crossed_fingers:
Do you have an interest in enabling pkgr directly for this repo (gogs/gogs) so it's always up to date? Otherwise I can set mine up as an unofficial one that's actually tended to.

unknwon:
Thanks for the suggestion!

>Do you have an interest in enabling pkgr directly for this repo (gogs/gogs) so it's always up to date? 

This sounds a better idea, just created: https://packager.io/gh/gogs/gogs anything else we need to do?

half-duplex:
As long as anything on master can be assumed stable, just hit the "Build Now" button on that page and you're done.
If you want to offer builds for the develop branch too, that can be enabled in settings.
Thank you!

unknwon:
@half-duplex thanks a lot! I've enabled `develop` branch as well, seems like it's building!

unknwon:
Oops, it failed with compile error that could be not reproduced locally...

half-duplex:
Oop, the `master` ones won't work until these patches make it there, but then it should. `develop` should work as-is.
The old setup was trying to `govendor update`, which updates a dependency to an incompatible version. The heroku.sync=false disabled that, leaving the vendored version in place.

unknwon:
Gotcha, rebuilding now!

PR #5773, opened 2019-08-06
Add rsync to dockerDevelop

pabloalcantara:
Add rsync to docker.
Rsync is nearly a necessity to optimize backup inside Openshift container.
Using TAR (default without Rsync), the process of external backup constantly stops at the middle.


unknwon:
Thank you!

PR #5774, opened 2019-08-06
Update TRANSLATORS

mathiasrw:
Added my name as translator as I helped with the Danish version (via Crowdin)

---
This change is [Reviewable](https://reviewable.io/reviews/gogs/gogs/5774)


unknwon:
Thank you very much!

Issue #5768, opened 2019-08-02
`mysql.sql` incompatible with MySql 8


Shiki-H:
**Describe the bug**
Setting up database for gogs using `mysql.sql` will give the following error in MySql 8:

```
ERROR 1193 (HY000) at line 1: Unknown system variable 'innodb_file_format'
```

**To Reproduce**
1. Go to `gogs/scripts`
2. Run `mysql -u [username] -p [password] < mysql.sql`
3. See error

**Additional context**
`innodb_file_format` and `innodb_large_prefix` have been deprecated in MySql 8: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-0.html

If you are using MySql 8, you can simply remove the corresponding lines to make it work.

- Gogs version (or commit ref): 0.11.86
- Git version: 2.17.1
- Operating system: Ubuntu 18.04
- Database (replace `[ ]` with `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] MSSQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes: provide example URL
  - [x] No: this is related to local set up 
- Log gist (usually found in `log/gogs.log`):

unknwon:
This is claimed to be fixed by merging #5786, please help test on `develop` branch.

PR #5786, opened 2019-08-18
fixed mysql compatibility issue

Shiki-H:
Fix to #5768 
I have tested it on MySQL 5.7, MySQL 8 and MariaDB 10.3


unknwon:
Thank you!

PR #5788, opened 2019-08-21
Dont send email to non active

Supme:


unknwon:
Thanks for the PR!

Can you confirm with these changes, it works as intended (i.e. not sending notification emails to non-active users)?

Supme:
Yes I confirm. I have tested this.

Issue #5794, opened 2019-09-01
Init scripts for DDWRT / Entware is missing


mb-0:
**Describe the bug**
The arm architecture allow Gogs to run on smaller systems, such as routers with custom firmware. OpenWRT, DDWRT and NAS firmware commonly rely on Entware for extensibility, but rc / init is very different from full-fledged linux distributions to keep things small & fast for the weaker hardware.

There is no init scripts available for users attempting to run gogs on Entware supported systems.

**To Reproduce**
1. Install gogs as per guide
2. Start in manually to confirm all is looking good.
3. Attempt to launch gogs as a daemon / background service.
4. Observe

**Expected behavior**
Have an init script for Entware supported systems.

**Additional context**
- Gogs version (or commit ref): all
- Git version:  all
- Operating system: Linux test 4.4.187 #655 SMP PREEMPT Tue Aug 6 12:03:57 +04 2019 armv7l DD-WRT
- Database:
  - [x] MySQL
- Can you reproduce the bug at https://try.gogs.io:
  - [x] No: bug is service startup related
- Log gist (usually found in `log/gogs.log`): not relevant

- Gist for setting up gogs on a ddwrt router: https://gist.github.com/mb-0/606840ace38a056109f526dd0e5c6762

mb-0:
PR that may solve this: https://github.com/gogs/gogs/pull/5795

unknwon:
Closed by https://github.com/gogs/gogs/pull/5795.

PR #5795, opened 2019-09-01
ddwrt / entware init script

mb-0:
Init script for DDWRT / Entware.
Present init script examples are not suitable for embedded systems, however, arm is indeed supported by gogs.

This PR is targeting issue #5794.


unknwon:
Thank you!

Issue #5782, opened 2019-08-15
Sending email to non-active users


Supme:
1. Go to any repository 
2. Click on 'Watch '
3. Set this account is not active and prohibited to login
4. But notifications sending to his email

**Expected behavior**

**Screenshots**

**Additional context**

- Gogs version (or commit ref):  0.11.86.0130
- Git version: 2.1.4
- Operating system: linux
- Database (replace `[ ]` with `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [ ] MSSQL
  - [x] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes: provide example URL
  - [x] No: i`m not admin
- Log gist (usually found in `log/gogs.log`):

Supme:
And removing people from organization does not remove them from repository watchers.

Supme:
Or/and make button "unwatch" for admin in repository watchers page.

unknwon:
Thanks for the feedback!

I think the solution is as easy as check if user is active before add to the recipients.

unknwon:
This is claimed to be fixed by merging https://github.com/gogs/gogs/pull/5788, please help test on `develop` branch.

Supme:
I don't know if I want to remove users from observers when remove them from organization, or only make button "unwatch" for admin.

PR #5803, opened 2019-09-17
Allow admin to remove observers from the repository (#5782)

Supme:


templates/repo/user_cards.tmpl
```diff
@@ -19,6 +19,15 @@
 						<span class="octicon octicon-clock"></span> {{$.i18n.Tr "user.join_on"}} {{DateFmtShort .Created}}
 					{{end}}
 				</div>
+
+        {{if $.IsAdmin}}
+          <form class="display inline" action="{{$.RepoLink}}/action/unwatch?user_name={{.Name}}&redirect_to={{$.Link}}" method="POST">
```
unknwon: I think it would be better to use user's ID.

And since this is POST request, use a hidden input `<input type="hidden" name="user_id" value="xxx">` is more ideal.

routes/repo/repo.go
```diff
@@ -235,7 +235,17 @@ func Action(c *context.Context) {
 	case "watch":
 		err = models.WatchRepo(c.User.ID, c.Repo.Repository.ID, true)
 	case "unwatch":
-		err = models.WatchRepo(c.User.ID, c.Repo.Repository.ID, false)
+		if userName := c.Query("user_name"); userName != "" {
+			if c.User.IsAdmin {
+				var user *models.User
+				if user, err = models.GetUserByName(userName); err != nil {
```
unknwon: Make it into two lines with `user, err := ` fits better here.

Supme: this err used later

templates/repo/user_cards.tmpl
Supme: I quite agree.

routes/repo/repo.go
```diff
@@ -235,7 +236,18 @@ func Action(c *context.Context) {
 	case "watch":
 		err = models.WatchRepo(c.User.ID, c.Repo.Repository.ID, true)
 	case "unwatch":
-		err = models.WatchRepo(c.User.ID, c.Repo.Repository.ID, false)
+		if userID := c.Query("user_id"); userID != "" {
```
unknwon: Sorry I forgot to mention in the last round of review, there is a better method:

https://github.com/gogs/gogs/blob/ef2e29db725c573d0d9bef37bcee56022aeea3c9/vendor/gopkg.in/macaron.v1/context.go#L240-L243

templates/repo/user_cards.tmpl
```diff
@@ -19,6 +19,16 @@
...
+
+        {{if $.IsAdmin}}
+          <form class="display inline" action="{{$.RepoLink}}/action/unwatch?redirect_to={{$.Link}}" method="POST">
+ 				    {{$.CSRFTokenHTML}}
+ 				    <input type="hidden" name="user_id" value="{{.ID}}">
+						<div class="ui labeled button" tabindex="0">
+						  <button class="ui red small button">{{$.i18n.Tr "repo.unwatch"}}</button>
+						</div>
+					</form>
+   			{{end}}
```
unknwon: The indentation here seems mixed, I see spaces, tabs, and are not aligned properly.

routes/repo/repo.go
Supme: thanks

PR #5804, opened 2019-09-18
ldap: log two non-errors on Trace level

zwets:
Minor fix that reduces two lines incorrectly logged as Error to Trace level.


unknwon:
Thank you!

PR #5806, opened 2019-09-19
Improvements for ddwrt init script

mb-0:
No particular issue called out here: general reliability improvements to support weaker routers.

- Removed reliance on sudo, daemonize deals with forking as user.
- Added wait for SQL as it may take more time to start up on a router: make sure you configure this before using.
- Proper Start/Stop/Reset is now implemented.


unknwon:
Thank you!

PR #5812, opened 2019-10-02
api/v1: fix status response in checkUserFollowing

sauyon:
It appears that the current behavior is to always return not found on a call to `checkUserFollowing`. This was introduced in f1e0ebfe937213be2af4a3180fae9f43949de00e, when `c.Status(204)` was incorrectly replaced with `c.NotFound()`. I have replaced this with `c.NoContent()`.

Found with [LGTM.com](https://lgtm.com/projects/g/gogs/gogs/alerts/?mode=list&lang=go), on which I am a developer.


unknwon:
Good catch, thank you!

unknwon:
LGTM.com looks very nice!

PR #5820, opened 2019-10-13
token: disallow multiple tokens with same name

frodeaa:
Fail with 422 Unprocessable Entity if the token name
already exist

ref: https://github.com/gogs/gogs/issues/5587

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -374,6 +374,7 @@ delete_token = Delete
 access_token_deletion = Personal Access Token Deletion
 access_token_deletion_desc = Delete this personal access token will remove all related accesses of application. Do you want to continue?
 delete_token_success = Personal access token has been removed successfully! Don't forget to update your application as well.
+token_name_exists = Token with same name already exists.
 
 orgs.none = You are not a member of any organizations.
 orgs.leave_title = Leave organization
--- a/models/error.go
+++ b/models/error.go
@@ -253,6 +253,19 @@ func (err ErrAccessTokenNotExist) Error() string {
 	return fmt.Sprintf("access token does not exist [sha: %s]", err.SHA)
 }
 
+type ErrAccessTokenNameAlreadyExist struct {
+	Name string
+}
+
+func IsErrAccessTokenNameAlreadyExist(err error) bool {
+	_, ok := err.(ErrAccessTokenNameAlreadyExist)
+	return ok
+}
+
+func (err ErrAccessTokenNameAlreadyExist) Error() string {
+	return fmt.Sprintf("access token already exist [name: %s]", err.Name)
+}
+
 type ErrAccessTokenEmpty struct {
 }
 
--- a/models/error.go
+++ b/models/error.go
@@ -253,19 +253,6 @@ func (err ErrAccessTokenNotExist) Error() string {
 	return fmt.Sprintf("access token does not exist [sha: %s]", err.SHA)
 }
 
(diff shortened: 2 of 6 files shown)


models/error.go
```diff
@@ -253,6 +253,19 @@ func (err ErrAccessTokenNotExist) Error() string {
 	return fmt.Sprintf("access token does not exist [sha: %s]", err.SHA)
 }
 
+type ErrAccessTokenNameAlreadyExist struct {
```
unknwon: Could you put this new error type in the [errors](https://github.com/gogs/gogs/tree/master/models/errors) subdirectory? We should create a new file named `token.go` to include it.

models/token.go
```diff
@@ -69,6 +85,12 @@ func GetAccessTokenBySHA(sha string) (*AccessToken, error) {
 	return t, nil
 }
 
+// ListAccessTokensByName returns a list of access tokens belongs to given user with name.
+func ListAccessTokensByName(uid int64, name string) ([]*AccessToken, error) {
```
unknwon: What is the use case of this function?

routes/user/setting.go
```diff
@@ -607,7 +607,12 @@ func SettingsApplicationsPost(c *context.Context, f form.NewAccessToken) {
 		Name: f.Name,
 	}
 	if err := models.NewAccessToken(t); err != nil {
-		c.ServerError("NewAccessToken", err)
+		if models.IsErrAccessTokenNameAlreadyExist(err) {
+			c.Flash.Error("NewAccessToken: " + err.Error())
```
unknwon: We should add an i18n entry in the locale file `locale_en-US.ini`, e.g. `settings.token_name_exists`, and use `c.Tr(...)` here. (see an example on line 619)

models/token.go
frodeaa: probably none as it will in the future only return 0 or 1

routes/user/setting.go
frodeaa: hm, do we need to run an extra command to get the i18n entry?

the error message is `token_name_exists` when running gogs with `make web`

conf/locale/locale_en-US.ini
```diff
@@ -729,6 +729,7 @@ settings.site = Official Site
 settings.update_settings = Update Settings
 settings.change_reponame_prompt = This change will affect how links relate to the repository.
 settings.advanced_settings = Advanced Settings
+settings.token_name_exists = Token with same name already exists.
```
unknwon: Hmm sorry I know it is somewhat confusing, the line should actually be added after/around here:

https://github.com/gogs/gogs/blob/99f7b6c1f37e8c6e78009c5793652e402df6b9a4/conf/locale/locale_en-US.ini#L370

FYI, `settings.` indicate the section name, then followed by the key name. The place where you added is actually `repo.settings.`, which is not what we want.

models/token.go
```diff
@@ -47,10 +48,26 @@ func (t *AccessToken) AfterSet(colName string, _ xorm.Cell) {
 	}
 }
 
+func isAccessTokenNameExist(uid int64, name string) (bool, error) {
+	tokens := make([]*AccessToken, 0, 5)
```
unknwon: I think we can make this function as simple as this (`.Get()` method):

https://github.com/gogs/gogs/blob/99f7b6c1f37e8c6e78009c5793652e402df6b9a4/models/user.go#L483

Which in this case, since it only has one line, we probably don't need this function to exist anymore.

unknwon:
Thank you!

PR #5835, opened 2019-10-24
Enable Go modules

unknwon:
Also rename import path as "gogs.io/gogs".


topillar:
Just finished build exp/go-go-modules for Windows、Aarch64 and Aarch64(with SQLite3 and cert) with success, then I found this merge.
go mod----- Cool tool!

unknwon:
@topillar haha! I'm working on update docs right now.

topillar:
Nice app , thank you very much!

PR #5843, opened 2019-10-28
repo: delete local copies on owner name change

guysmoilov:
And protect local copy deletion with repoWorkingPool.

Solves https://github.com/gogs/gogs/issues/5823

### DIFF
--- a/internal/db/user.go
+++ b/internal/db/user.go
@@ -675,7 +675,7 @@ func ChangeUserName(u *User, newUserName string) (err error) {
 		RemoveAllWithNotice("Delete repository wiki local copy", repo.LocalWikiPath())
 		return nil
 	}); err != nil {
-		return fmt.Errorf("Delete repository wiki local copy: %v", err)
+		return fmt.Errorf("delete repository and wiki local copy: %v", err)
 	}
 
 	// Rename or create user base directory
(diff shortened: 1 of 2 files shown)


internal/db/user.go
```diff
@@ -670,6 +670,8 @@ func ChangeUserName(u *User, newUserName string) (err error) {
 	// Delete all local copies of repository wiki that user owns.
```
unknwon:
```suggestion
	// Delete all local copies of repositories and wikis the user owns.
```

internal/db/user.go
```diff
@@ -670,6 +670,8 @@ func ChangeUserName(u *User, newUserName string) (err error) {
...
 	if err = x.Where("owner_id=?", u.ID).Iterate(new(Repository), func(idx int, bean interface{}) error {
 		repo := bean.(*Repository)
+		deleteRepoLocalCopy(repo)
+		// TODO: By the same reasoning, shouldn't we also sync access to the local wiki path?
 		RemoveAllWithNotice("Delete repository wiki local copy", repo.LocalWikiPath())
 		return nil
 	}); err != nil {
```
unknwon: I can't comment on line 678...

```go
return fmt.Errorf("delete repository and wiki local copy: %v", err)
```

guysmoilov: Thanks, changed it

unknwon:
Thank you!

PR #5850, opened 2019-11-06
tls: update default CurvePreferences and CipherSuites

kaseiwang:
* Enable X25519 curve and reorder curve list to improve key exchange performance
* Enable ECDSA ciphers for EC certs
* Enable CHACHA20_POLY1305 ciphers
* Disable RSA key exchange algorithm which don't provide PFS
* Disable non-AEAD ciphers


unknwon:
Thank you!

PR #5856, opened 2019-11-07
updated the highlights.js plugin

johannes69ki:
This is my request for a new feature.

I just ran into the issue that we want to put SAS code files (some analytics software suite https://www.sas.com/) into our gogs but that there was no working synthax hightlighting for these file types. I figured out that you are using highlights.js as a plugin for this and further figured out that the current version of highlights.js actually supports SAS.

I therefore herewith suggest to update the plugin to the current version.

While locally testing my changes I noticed that the automatic language deduction for SAS files did not always work and therefore I added some additional code that explicitly sets the language type for highlights.js based on the file exentsion of the source file.

I think that this is a nice to have for everyone, and thus the pull request.


unknwon:
Nice, thank you!

PR #5857, opened 2019-11-08
public: makes CodeMirror mode by filename lookups case-insensitive

johannes69ki:
Fixes a minor bug.

When using the CodeMirror editor to edit and commit a source file, the synthax highlighting did not work, for filenames like e.g. Test.R where the extension is written in uppercase. It is common for the R programming language to use an uppercase R... Modes in CodeMirror are however always stored in lowercase. Thus the mode/language could not be matched by the CodeMirror.findModeByExtension function. This pull request fixes this behavior.

I made another pull request to CodeMirror itself, to fix the root cause: https://github.com/codemirror/CodeMirror/pull/6057


unknwon:
Thank you!

PR #5859, opened 2019-11-11
Change PATCH to POST

m-pavel:
Changes required in gogs project to fix https://github.com/gogs/go-gogs-client/issues/111


unknwon:
Thanks, please see [my comment](https://github.com/gogs/go-gogs-client/issues/111#issuecomment-553235004).

PR #5868, opened 2019-11-19
issue template typo

innerpeacez:
image
image

What happened? Interesting.    :)


unknwon:
Hi, thanks for the feedback!

In fact, GitHub does not care the numbers, you can use all `1.` and it will just work.

PR #5888, opened 2019-12-21
repo: use editorconfig v2

greut:
The editorconfig library go bumped to v2 with some improvements and bugfixes. The goal will be to tackle the `end_of_line` issue https://github.com/gogs/gogs/issues/3550 if this is still a thing required.


internal/context/repo.go
```diff
@@ -9,9 +9,9 @@ import (
 	"io/ioutil"
 	"strings"
 
-	"gopkg.in/editorconfig/editorconfig-core-go.v1"
 	"gopkg.in/macaron.v1"
 
+	"github.com/editorconfig/editorconfig-core-go/v2"
```
unknwon: This import grouping should be changed.

internal/template/template.go
```diff
@@ -116,8 +116,8 @@ func NewFuncMap() []template.FuncMap {
 		},
 		"TabSizeClass": func(ec *editorconfig.Editorconfig, filename string) string {
 			if ec != nil {
-				def := ec.GetDefinitionForFilename(filename)
-				if def.TabWidth > 0 {
+				def, err := ec.GetDefinitionForFilename(filename)
+				if err != nil && def.TabWidth > 0 {
```
unknwon: `err == nil`?

greut: Good catch, that muscle memory.

PR #5890, opened 2019-12-28
Fix private repositories are hidden in the organization's view bug

### LINKED ISSUE #5869: Private repositories are hidden in the organization's view
Within the organization's page I was expecting to find our private repositories, but they weren't there; only the public ones. I happened to find them only throught the generic "explore" menù.

The organizazion's page should list only the public repositories when the user isn'n logged in or if he/she doesn't have the permissions. Otherwise both the public and private ones should be listed.

Gogs version 0.11.96.1107

alimy:
This patch Fixes #5869.


unknwon:
Thank you!

PR #5896, opened 2020-01-04
Show file name in window/tab title

GeorchW:
I've just encountered the problem that when browsing multiple files in different tabs, the tab title does not show the file name, which makes browsing a repository a happy game of memory for the entire family. This PR takes the fun away and should just display the file name in the title.

### DIFF
--- a/internal/route/repo/view.go
+++ b/internal/route/repo/view.go
@@ -132,6 +132,7 @@ func renderFile(c *context.Context, entry *git.TreeEntry, treeLink, rawLink stri
 		return
 	}
 
+	c.Data["Title"] = blob.Name() + " - " + c.Data["Title"]
 	c.Data["FileSize"] = blob.Size()
 	c.Data["FileName"] = blob.Name()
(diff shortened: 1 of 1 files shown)


unknwon:
Thanks for the PR! 

But the CI is failing with some type error.

internal/route/repo/view.go
```diff
@@ -132,6 +132,7 @@ func renderFile(c *context.Context, entry *git.TreeEntry, treeLink, rawLink stri
 		return
 	}
 
+	c.Data["Title"] = blob.Name() + " - " + c.Data["Title"]
```
roberChen: see [macaron.Context.Data](https://github.com/go-macaron/macaron/blob/b2cd309dc711ad238e63785012080a37ccd6314c/context.go#L97), the type is ```map[string]interface{}```, it should be change to ```string```.

unknwon (changes requested):
See previous comment.

PR #5900, opened 2020-01-10
user/setting: check for IsErrEmailAlreadyUsed when updating user

### LINKED ISSUE #5899: Error 500 when changing email address to existing one
**Describe the bug**
When changing email address in `/user/settings`, if the new email already exists for another user, an error 500 page is displayed instead of the proper error message.

**Gogs version or commit**
76b87b1bbd9dcbb1adf0192b1168905565fea774

**Git version**

```
$ git version
git version 2.22.2
```

**Operating system**
Alpine in Docker (on Arch Linux)

**Database**
sqlite

**To Reproduce**
1. Go to https://try.gogs.io/user/settings
2. Change Email address to one that's already taken.
3. Click "Update Profile".
4. See error 500 page.

**Can you reproduce the bug at https://try.gogs.io?**
Yes.  See repro instructions above.

**Expected behavior**
The message "Email address has already been used." should appear above the form in a red box.

**Actual behavior**
500 internal server error page.

**Additional context**
The email conflict error is incorrectly checked during the username change (`db.ChangeUserName()`) instead of when the whole user information is updated (`db.UpdateUser()`).
See https://github.com/gogs/gogs/blob/76b87b1bbd9dcbb1adf0192b1168905565fea774/internal/route/user/setting.go#L78-L79

achilleas-k:
Check for email collisions when updating the entire user information, not when the username is being changed.

Changing a user's email address through `/user/settings` to an email address that's already taken shows an internal server error instead of the appropriate error message.

The `ErrEmailAlreadyUsed` error type was incorrectly checked for during the username update in the same function, when it is impossible for that error to occur.  This PR performs the check after the `db.UpdateUser()` call, which can return the "email already used" error, and renders the error message on the form accordingly, instead of failing with an internal server error.

Fixes #5899

### DIFF
--- a/internal/route/user/setting.go
+++ b/internal/route/user/setting.go
@@ -75,8 +75,6 @@ func SettingsPost(c *context.Context, f form.UpdateProfile) {
 				switch {
 				case db.IsErrUserAlreadyExist(err):
 					msg = c.Tr("form.username_been_taken")
-				case db.IsErrEmailAlreadyUsed(err):
-					msg = c.Tr("form.email_been_used")
 				case db.IsErrNameReserved(err):
 					msg = c.Tr("form.name_reserved")
 				case db.IsErrNamePatternNotAllowed(err):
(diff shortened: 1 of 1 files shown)


unknwon:
Thank you!

PR #5920, opened 2020-02-12
Use go-bindata to embed `public` and `templates` files into binary

alimy:
1. use go-bindata pack up public and templates files into gogs binary then no need public/templates file in gogs work directory.
2. after this patch will make gogs update more convenience.
3. after make relealse the package will just contain files below:
```
% tree -L  1 gogs
gogs
├── LICENSE
├── README.md
├── README_ZH.md
├── gogs
└── scripts

```


alimy:
Installation
To install the newest go-bindata, use the following:
```
% go get -u github.com/go-bindata/go-bindata/...
```

unknwon:
Thanks for the PR!

According to https://github.com/gogs/gogs/blob/master/conf/README.md, you should use https://github.com/kevinburke/go-bindata. Can you verify the result is the same? Or re-generate if necessary?

alimy:
sorry, I don't notice the go-bindata need use github.com/kevinburke/go-bindata.  Compare github.com/go-bindata/go-bindata, the github.com/kevinburke/go-bindata not support generate create http.FileSystem instance code by ``` go-bindata -fs ...```  that [PR](https://github.com/go-bindata/go-bindata/pull/18) supported by github.com/go-bindata/go-bindata.

This patch depend go-bindata generate a function to returan a ```http.FileSystem``` in internal/assets/public so have 3 solutions fix this patch:
1.  use github.com/go-bindata/go-bindata to generate static files.
2. commit  a PR to github.com/kevinburke/go-bindata to support generate create http.FileSystem instance code like [PR](https://github.com/go-bindata/go-bindata/pull/18).I will do this later.
3. add a custom function in native to create a http.FileSystem instance from generated static public files code.

I think github.com/go-bindata/go-bindata is good  same as github.com/kevinburke/go-bindata so how about use github.com/go-bindata/go-bindata as our generate tool.

unknwon:
> I think github.com/go-bindata/go-bindata is good same as github.com/kevinburke/go-bindata so how about use github.com/go-bindata/go-bindata as our generate tool.

Gogs used to use `github.com/go-bindata/go-bindata` but has changed to use `github.com/kevinburke/go-bindata` because that's the one installed by `brew install go-bindata`.

> 3\. add a custom function in native to create a http.FileSystem instance from generated static public files code.

I think this is my preferred solution. 

Besides, (I haven't looked at code in detail), it would be nice to have:
1. In dev mode, it reads directly from the file system.
2. In non-dev mode, it reads from what's been embedded.

alimy:
Had optimize patch:
1. use github.com/kevinburke/go-bindata as generate tool.
2. add a custom function to instance a http.FileSystem from generated static public assets code.
3. add ```ENABLE_ASSETS = true``` item in configure file to support either enable use embedded public and templates assets (not use ```RUN_MODE``` item because sometimes we just test enable embedded public assets feature in dev mode and other scenes).

Makefile
```diff
@@ -1,9 +1,11 @@
 LDFLAGS += -X "gogs.io/gogs/internal/setting.BuildTime=$(shell date -u '+%Y-%m-%d %I:%M:%S %Z')"
 LDFLAGS += -X "gogs.io/gogs/internal/setting.BuildGitHash=$(shell git rev-parse HEAD)"
 
-DATA_FILES := $(shell find conf | sed 's/ /\\ /g')
+ASSETS_FILES := $(shell find conf | sed 's/ /\\ /g')
+TEMPLATES_FILES := $(shell find templates | sed 's/ /\\ /g')
+PUBLIC_FILES := $(shell find public | sed 's/ /\\ /g')
 LESS_FILES := $(wildcard public/less/gogs.less public/less/_*.less)
-GENERATED  := internal/bindata/bindata.go public/css/gogs.css
+GENERATED  := internal/assets/assets_gen.go internal/assets/templates/templates_gen.go internal/assets/public/public_gen.go public/css/gogs.css
```
unknwon: We should put `CONF_FILES` in its own directory as well.

internal/assets/public/public.go
```diff
@@ -0,0 +1,109 @@
...
+		&bindataFileInfo{
+			name:    name,
+			size:    0,
+			mode:    os.FileMode(2147484068), // equal os.FileMode(0644)|os.ModeDir
+			modTime: time.Time{},
+		}}
+}
+
+// AssetFile return a http.FileSystem instance that data backend by asset
+func AssetFile() http.FileSystem {
```
unknwon: I think this is more clear:
```suggestion
func FileSystem() http.FileSystem {
```

internal/assets/public/public.go
```diff
@@ -0,0 +1,109 @@
+package public
```
unknwon: This file seems having more complicated logic than I would expect, do you think there is a way to reuse some stdlib functions/methods to simplify it? Or did you copy it from go-bindata/go-bindata?

internal/cmd/web.go
```diff
@@ -61,8 +62,7 @@ and it takes care of all the other things for you`,
 
 // checkVersion checks if binary matches the version of templates files.
 func checkVersion() {
```
unknwon: Since we're now embedding all these files, I think it's time to get rid of this function entirely!

internal/db/repo.go
```diff
@@ -23,15 +23,16 @@ import (
 	"github.com/nfnt/resize"
 	"github.com/unknwon/cae/zip"
 	"github.com/unknwon/com"
+
```
unknwon:
```suggestion
```

internal/mailer/mail.go
```diff
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"html/template"
 
+	"gogs.io/gogs/internal/assets/templates"
```
unknwon: It is misplaced, should be moved to line 16.

unknwon:
Thanks again for the contribution!

PS: If you think you don't have bandwidth work on this in the coming week, LMK! I'll push changes to this PR. I want to get this merged :)

alimy:
Thanks your suggestions. I had update the patch and  welcome optimize the code :)

internal/assets/public/public.go
alimy: yes, it's is copy from go-bindata/go-bindata that base [PR](https://github.com/go-bindata/go-bindata/pull/18) and make a lite optimize just do it fun. Maybe have some other way optimize it take a time later :)

conf/app.ini
```diff
@@ -8,6 +8,8 @@ APP_NAME = Gogs
 RUN_USER = git
 ; Either "dev", "prod" or "test"
 RUN_MODE = dev
+; Either load public and templates assets from disk
```
unknwon:
```suggestion
; Enable to load assets (i.e. "conf", "templates", "public") from disk instead of embedded bindata
```

unknwon:
OK, I think I'm happy with the current state, merging!

ozzi-:
Is there any possibility to edit the templates without recompiling?
Background, we used a custom home template adding a logo and different text.

unknwon:
@ozzi- A restart of Gogs is all you need if you are editing files under `custom/` directory.

Issue #5934, opened 2020-02-23
API for listing Git repository tree


bharatnc:
**Describe the feature**

I'm running into a use case where I would like to fetch all the files in a given repo for a given `ref` (say master)  and then fetch their raw content. 

This endpoint: `/api/v1/repos/:user/:repo-name/raw/:refs/:filename`  is very much useful if I actually know which file I need to fetch raw.

It would be useful to have api endpoint which can fetch a list of all files for a given ref for a given repo for a given user. 

**Describe the solution you'd like**

When given:  `/repos/:user/:repo_name/tree` fetch the list of all the files in that repo. Additionally params for recursive functionality can also work with `?recursive=1 `.  For more context, Gitlab provides a repository tree API: https://docs.gitlab.com/ee/api/repositories.html.

**Additional context**

I perused the gogs source code for API and haven't found any tree endpoint exposed: https://github.com/gogs/gogs/blob/master/internal/route/api/v1/repo/repo.go. 

 I haven't actually looked at the core backend to understand if all that's needed is to expose the backend logic via proper api path. 

Would like to help implement it if we don't already have this feature  and is do-able.

PS: Pardon if I might be double reporting an existing/similar issue, but I tried my best to isearch through existing issues to ensure that this hasn't been asked before and the closet I can find are issues like: #1061 and #1253.

Thank you!

unknwon:
Thanks for the feedback!

For the API design, I would like to align with GitHub API v3, i.e. https://developer.github.com/v3/git/trees/. It is definitely doable with existing codebase (though we don't need to truncate results in the initial implementation). 

Roughly, you would:
1. Get the root tree via: https://github.com/gogs/git-module/blob/f29bce1e3a666c02175d080892be185405dd3af4/repo_tree.go#L20
2. List entries in the root tree: https://github.com/gogs/git-module/blob/f29bce1e3a666c02175d080892be185405dd3af4/tree.go#L137
3. For each tree entry in the root tree, do https://github.com/gogs/git-module/blob/f29bce1e3a666c02175d080892be185405dd3af4/tree.go#L108

NOTE: The methods I liked are on `git-module@v0.8.3`, NOT `git-module@v1.0.0-beta` (this is a refactored one, and haven't migrated Gogs to adopt it yet, it has huge amount of work and testing ahead).

bharatnc:
@unknwon thanks for the pointers, I'm going to attempt to take a look.

jianboy:
[...]

PR #5937, opened 2020-02-24
api: support getting repository Git tree (#5934)

bharatnc:
This PR adds the tree  api endpoint to gogs api:
`GET/repos/:owner/:repo/git/trees/:tree_sha`

This new api endpoint that is being added is in conformance to
the GitHub REST API v3 specification. Documentation can be found
here: developer.github.com/v3/git/trees/#get-a-tree

For a given user, repo and sha value, this api (currently) returns
a single tree using the SHA1 value for that tree.

- Recursive implementation is yet to be implemented.
- Creating a Tree using POST is yet to be implemented.

Example curl:

```

l curl -H "Authorization: token REDACTED" http://localhost:3000/api/v1/repos/root/testrepo/git/trees/c59441ded1549b149def0d4c54594d31a7f3718f -X GET  | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   940  100   940    0     0  12034      0 --:--:-- --:--:-- --:--:-- 12051

[
  {
    "sha": "c59441ded1549b149def0d4c54594d31a7f3718f",
    "tree": [
      {
        "mode": "120000",
        "path": "/home/bharatnc/gogs-repositories/root/testrepo.git",
        "sha": "472ac2361b65136b393d652de25341e2ea44f299",
        "size": 1077,
        "type": "blob",
        "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/472ac2361b65136b393d652de25341e2ea44f299"
      },
      {
        "mode": "120000",
        "path": "/home/bharatnc/gogs-repositories/root/testrepo.git",
        "sha": "70fcb456d436f08462602f26df6fb7e167e7a916",
        "size": 12,
        "type": "blob",
        "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/70fcb456d436f08462602f26df6fb7e167e7a916"
      },
      {
        "mode": "120000",
        "path": "/home/bharatnc/gogs-repositories/root/testrepo.git",
        "sha": "092c58d4b63df5779a4d020b1fdbb762421bbb4f",
        "size": 380,
        "type": "blob",
        "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/092c58d4b63df5779a4d020b1fdbb762421bbb4f"
      }
    ],
    "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/c59441ded1549b149def0d4c54594d31a7f3718f"
  }
]

```

### DIFF
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -39,7 +39,7 @@ func GetRepoGitTree(c *context.APIContext) {
 
 	templateURL := fmt.Sprintf("%s/repos/%s/%s/git/trees", c.BaseURL, c.Params(":username"), c.Params(":reponame"))
 
-	if entries == nil {
+	if len(entries) == 0 {
 		c.JSONSuccess(&repoGitTree{
 			Sha: c.Params(":sha"),
 			URL: fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -6,7 +6,6 @@ import (
 	"github.com/gogs/git-module"
 
 	"gogs.io/gogs/internal/context"
-	"gogs.io/gogs/internal/db"
 )
 
 type repoGitTree struct {
@@ -25,7 +24,6 @@ type repoGitTreeEntry struct {
 }
 
 func GetRepoGitTree(c *context.APIContext) {
-	repoPath := db.RepoPath(c.Params(":username"), c.Params(":reponame"))
 	gitTree, err := c.Repo.GitRepo.GetTree(c.Params(":sha"))
 	if err != nil {
 		c.NotFoundOrServerError("GetRepoGitTree", git.IsErrNotExist, err)
@@ -63,7 +61,7 @@ func GetRepoGitTree(c *context.APIContext) {
 			mode = ""
 		}
 		children = append(children, &repoGitTreeEntry{
-			Path: repoPath,
+			Path: entry.Name(),
 			Mode: mode,
 			Type: string(entry.Type),
 			Size: entry.Size(),
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -47,10 +47,9 @@ func GetRepoGitTree(c *context.APIContext) {
 		return
 	}
 
-	var children []*repoGitTreeEntry
-	var mode string
-
+	children := make([]*repoGitTreeEntry, 0, len(entries))
 	for _, entry := range entries {
+		var mode string
 		switch entry.Type {
 		case git.ObjectCommit:
 			mode = "160000"
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -9,13 +9,13 @@ import (
 	"gogs.io/gogs/internal/db"
 )
 
-type repoGitTreeResponse struct {
-	Sha  string         `json:"sha"`
-	URL  string         `json:"url"`
-	Tree []*repoGitTree `json:"tree"`
+type repoGitTree struct {
+	Sha  string              `json:"sha"`
+	URL  string              `json:"url"`
+	Tree []*repoGitTreeEntry `json:"tree"`
 }
 
-type repoGitTree struct {
+type repoGitTreeEntry struct {
 	Path string `json:"path"`
 	Mode string `json:"mode"`
 	Type string `json:"type"`
@@ -36,17 +36,18 @@ func GetRepoGitTree(c *context.APIContext) {
 		c.ServerError("GetRepoTree", err)
 		return
 	}
-	var children []*repoGitTree
+	var children []*repoGitTreeEntry
 	var mode string
 	templateURL := `%s/repos/%s/%s/git/trees`
 	templateURL = fmt.Sprintf(templateURL, c.BaseURL, c.Params(":username"), c.Params(":reponame"))
 	if entries == nil {
-		res := &repoGitTreeResponse{
+		c.JSONSuccess(&repoGitTree{
 			Sha: c.Params(":sha"),
 			URL: fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
-		}
-		c.JSONSuccess(res)
+		})
+		return
 	}
+
 	for _, entry := range entries {
 		switch entry.Type {
 		case git.ObjectCommit:
@@ -60,7 +61,7 @@ func GetRepoGitTree(c *context.APIContext) {
 		default:
 			mode = ""
 		}
-		children = append(children, &repoGitTree{
+		children = append(children, &repoGitTreeEntry{
 			Path: repoPath,
 			Mode: mode,
 			Type: string(entry.Type),
@@ -69,10 +70,9 @@ func GetRepoGitTree(c *context.APIContext) {
 			URL:  fmt.Sprintf(templateURL+"/%s", entry.ID.String()),
 		})
 	}
-	results := &repoGitTreeResponse{
+	c.JSONSuccess(&repoGitTree{
 		Sha:  c.Params(":sha"),
 		URL:  fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
 		Tree: children,
-	}
-	c.JSONSuccess(results)
+	})
 }
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -28,18 +28,17 @@ func GetRepoGitTree(c *context.APIContext) {
 	repoPath := db.RepoPath(c.Params(":username"), c.Params(":reponame"))
 	gitTree, err := c.Repo.GitRepo.GetTree(c.Params(":sha"))
 	if err != nil {
-		c.NotFoundOrServerError("GetRepoTree", git.IsErrNotExist, err)
+		c.NotFoundOrServerError("GetRepoGitTree", git.IsErrNotExist, err)
 		return
 	}
 	entries, err := gitTree.ListEntries()
 	if err != nil {
-		c.ServerError("GetRepoTree", err)
+		c.ServerError("GetRepoGitTree", err)
 		return
 	}
-	var children []*repoGitTreeEntry
-	var mode string
-	templateURL := `%s/repos/%s/%s/git/trees`
-	templateURL = fmt.Sprintf(templateURL, c.BaseURL, c.Params(":username"), c.Params(":reponame"))
+
+	templateURL := fmt.Sprintf("%s/repos/%s/%s/git/trees", c.BaseURL, c.Params(":username"), c.Params(":reponame"))
+
 	if entries == nil {
 		c.JSONSuccess(&repoGitTree{
 			Sha: c.Params(":sha"),
@@ -47,6 +46,9 @@ func GetRepoGitTree(c *context.APIContext) {
 		})
 		return
 	}
+
+	var children []*repoGitTreeEntry
+	var mode string
 
 	for _, entry := range entries {
 		switch entry.Type {
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -18,7 +18,7 @@ type repoGitTreeEntry struct {
 	Path string `json:"path"`
 	Mode string `json:"mode"`
 	Type string `json:"type"`
-	Size int64  `json:"size,omitempty"`
+	Size int64  `json:"size"`
 	Sha  string `json:"sha"`
 	URL  string `json:"url"`
 }
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -74,6 +74,5 @@ func GetRepoGitTree(c *context.APIContext) {
 		URL:  fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
 		Tree: children,
 	}
-
 	c.JSONSuccess(results)
 }
--- a/internal/route/api/v1/repo/tree.go
+++ b/internal/route/api/v1/repo/tree.go
@@ -1,0 +1,79 @@
+package repo
+
+import (
+	"fmt"
+
+	"github.com/gogs/git-module"
+
+	"gogs.io/gogs/internal/context"
+	"gogs.io/gogs/internal/db"
+)
+
+type repoGitTreeResponse struct {
+	Sha  string         `json:"sha"`
+	URL  string         `json:"url"`
+	Tree []*repoGitTree `json:"tree"`
+}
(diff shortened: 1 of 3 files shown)


bharatnc:
@unknwon , I made some progress with respect to the trees api endpoint for gogs based on your inputs. Would appreciate feedback. Thank you!

internal/route/api/v1/repo/file.go
```diff
@@ -45,6 +48,72 @@ func GetArchive(c *context.APIContext) {
...
 
+func GetRepoTree(c *context.APIContext) {
+	repoPath := db.RepoPath(c.Params(":username"), c.Params(":reponame"))
+	gitRepo, err := git.OpenRepository(repoPath)
+
+	if err != nil {
+		c.ServerError("OpenRepository", err)
+		return
+	}
+	c.Repo.GitRepo = gitRepo
```
unknwon: `c.Repo.GitRepo` is already set in `context.RepoRef()` call, so you can just use it.

internal/route/api/v1/repo/file.go
```diff
@@ -45,6 +48,72 @@ func GetArchive(c *context.APIContext) {
...
+		c.NotFoundOrServerError("GetRepoTree", git.IsErrNotExist, err)
+		return
+	}
+	entries, err := gitTree.ListEntries()
+	if err != nil {
+		c.ServerError("GetRepoTree", err)
+		return
+	}
+	if entries == nil {
+		c.NotFound()
```
unknwon: Is it really not found? Or it's just an empty tree/directory.

internal/route/api/v1/repo/file.go
bharatnc: Thanks, fixed it.

internal/route/api/v1/repo/file.go
bharatnc: Yea, you're right. It's shouldn't be not found, it should be just an empty tree. I am going to preserve the top level structure but return an empty tree. 

Which will ook like:
```bash
[
  {
    "url":"http://localhost:3000/api/v1/repos/root/testrepo/git/trees/c59441ded1549b149def0d4c54594d31a7f3718f"
    "sha": "c59441ded1549b149def0d4c54594d31a7f3718f",
    "tree": []
}
```

bharatnc:
@unknwon,

> I don't see you add the new import of golang.org/x/tools and githubo.com/modern-go/* anywhere in the diffs, why go.mod wants to include those?
They should be gone now. Not sure what happened there.

I don't see any changes to "public/", "templates/" directories and any "*.less" files, where does these changes come form?
> Yep, interesting, I have discarded changes to those files and updated the PR. In fact when I again ran `make web`, I could see that new files are generated - though there are no changes to the UI :( 

```bash➜  gogs git:(bnc/gogs-5934-git-tree-api) ✗ make web
rm -f internal/assets/public/public_gen.go
go generate internal/assets/public/public.go
gofmt -s -w internal/assets/public/public_gen.go
```

> Which IDE you're using? The "internal/route/api/v1/repo/file.go" looks like is not gofmted, and it would more ideal to put in its own file, namely "tree.go".

i use GoLand by Jetbrains. But that shouldn't matter, I usually run `gofmt -w <file>` to ensure formatting. Only in this file it did not make any difference. I have made a few changes and formatted the file again so that it looks normal now.

Thank you for the feedback. I have also moved this new Tree method into it's own file called  `tree.go` as per your suggestion. Please let me know if you would like me to do some other changes. Thank you!

unknwon:
> > I don't see you add the new import of golang.org/x/tools and githubo.com/modern-go/* anywhere in the diffs, why go.mod wants to include those?
> > They should be gone now. Not sure what happened there.
> 
> I don't see any changes to "public/", "templates/" directories and any "*.less" files, where does these changes come form?

I think it was my mistake that I didn't properly update these files on `master`, which I fixed yesterday, so no longer showing them on your PR.

unknwon:
> i use GoLand by Jetbrains. But that shouldn't matter, I usually run `gofmt -w <file>` to ensure formatting. Only in this file it did not make any difference. I have made a few changes and formatted the file again so that it looks normal now.

GoLand sadly do not enable `gofmt` by default, here's how you could tell it to format:
image

bharatnc:
I have also synced my fork with the master and have properly formatted the changes.

bharatnc:
> > i use GoLand by Jetbrains. But that shouldn't matter, I usually run `gofmt -w <file>` to ensure formatting. Only in this file it did not make any difference. I have made a few changes and formatted the file again so that it looks normal now.
> 
> GoLand sadly do not enable `gofmt` by default, here's how you could tell it to format:
> image

Oh yea, I usually manually do `Tools ->  Go Tools -> Go fmt file`

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+
+	"github.com/gogs/git-module"
+
+	"gogs.io/gogs/internal/context"
+	"gogs.io/gogs/internal/db"
+)
+
+func GetRepoTree(c *context.APIContext) {
```
unknwon:
```suggestion
func GetRepoGitTree(c *context.APIContext) {
```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		c.NotFoundOrServerError("GetRepoTree", git.IsErrNotExist, err)
+		return
+	}
+	entries, err := gitTree.ListEntries()
+	if err != nil {
+		c.ServerError("GetRepoTree", err)
+		return
+	}
+	var results []map[string]interface{}
```
unknwon: Use `[]map[string]interface{}` in this case is not ideal, should define concrete types:

```go
type repoGitTreeResponse struct {
	Sha  string `json:"sha"`
	URL  string `json:"url"`
	Tree []*repoGitTree `json:"tree"`
}

type repoGitTree struct{
	Path string `json:"path"`
	Mode string `json:"mode"`
	Type string `json:"type"`
	Size int    `json:"size,omitempty"`
	Sha  string `json:"sha"`
	URL  string `json:"url"`
}
```

_Generated using https://mholt.github.io/json-to-go/_

unknwon:
PS: you don’t need to force push, we’re going to squash and merge anyway.

internal/route/api/v1/repo/tree.go
bharatnc: Sure, changed.

internal/route/api/v1/repo/tree.go
bharatnc: I agree, these should be now implemented, verified that these still return the same response:
```bash
{
  "sha": "c59441ded1549b149def0d4c54594d31a7f3718f",
  "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/c59441ded1549b149def0d4c54594d31a7f3718f",
  "tree": [
    {
      "path": "/home/bharatnc/gogs-repositories/root/testrepo.git",
      "mode": "120000",
      "type": "blob",
      "size": 1077,
      "sha": "472ac2361b65136b393d652de25341e2ea44f299",
      "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/472ac2361b65136b393d652de25341e2ea44f299"
    },
    {
      "path": "/home/bharatnc/gogs-repositories/root/testrepo.git",
      "mode": "120000",
      "type": "blob",
      "size": 12,
      "sha": "70fcb456d436f08462602f26df6fb7e167e7a916",
      "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/70fcb456d436f08462602f26df6fb7e167e7a916"
    },
    {
      "path": "/home/bharatnc/gogs-repositories/root/testrepo.git",
      "mode": "120000",
      "type": "blob",
      "size": 380,
      "sha": "092c58d4b63df5779a4d020b1fdbb762421bbb4f",
      "url": "http://localhost:3000/api/v1/repos/root/testrepo/git/trees/092c58d4b63df5779a4d020b1fdbb762421bbb4f"
    }
  ]
}

```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,79 @@
...
+	Size int64  `json:"size,omitempty"`
+	Sha  string `json:"sha"`
+	URL  string `json:"url"`
+}
+
+func GetRepoGitTree(c *context.APIContext) {
+	repoPath := db.RepoPath(c.Params(":username"), c.Params(":reponame"))
+	gitTree, err := c.Repo.GitRepo.GetTree(c.Params(":sha"))
+	if err != nil {
+		c.NotFoundOrServerError("GetRepoTree", git.IsErrNotExist, err)
```
unknwon:
```suggestion
		c.NotFoundOrServerError("GetRepoGitTree", git.IsErrNotExist, err)
```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,79 @@
...
+		return
+	}
+	entries, err := gitTree.ListEntries()
+	if err != nil {
+		c.ServerError("GetRepoTree", err)
```
unknwon:
```suggestion
		c.ServerError("GetRepoGitTree", err)
```

bharatnc:
> PS: you don’t need/have to force push, we’re going to squash and merge anyway.

Noted - will maintain additional commits hereafter. I'm just used to maintaining a single commit (force pushing) on a daily basis :) .

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,79 @@
...
+		return
+	}
+	var children []*repoGitTree
+	var mode string
+	templateURL := `%s/repos/%s/%s/git/trees`
+	templateURL = fmt.Sprintf(templateURL, c.BaseURL, c.Params(":username"), c.Params(":reponame"))
```
unknwon: Why not just 

```go
templateURL := fmt.Sprintf("%s/repos/%s/%s/git/trees", c.BaseURL, c.Params(":username"), c.Params(":reponame"))
```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,79 @@
...
+	if entries == nil {
+		res := &repoGitTreeResponse{
+			Sha: c.Params(":sha"),
+			URL: fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
+		}
+		c.JSONSuccess(res)
```
unknwon: You need a `return` after this.

unknwon: We could go to the extreme:
```go
c.JSONSuccess(&repoGitTreeResponse{
	Sha: c.Params(":sha"),
	URL: fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
})
return
```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,78 @@
...
+		c.NotFoundOrServerError("GetRepoTree", git.IsErrNotExist, err)
+		return
+	}
+	entries, err := gitTree.ListEntries()
+	if err != nil {
+		c.ServerError("GetRepoTree", err)
+		return
+	}
+	var children []*repoGitTree
+	var mode string
```
unknwon: We don't use these two variables until after the `if entries == nil {`, move them down to the their use cases is better.

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,78 @@
...
+	templateURL := `%s/repos/%s/%s/git/trees`
+	templateURL = fmt.Sprintf(templateURL, c.BaseURL, c.Params(":username"), c.Params(":reponame"))
+	if entries == nil {
+		res := &repoGitTreeResponse{
+			Sha: c.Params(":sha"),
+			URL: fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
+		}
+		c.JSONSuccess(res)
+	}
+	for _, entry := range entries {
```
unknwon: An empty line before the for loop is better for visually separating the logic pieces.

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"fmt"
+
+	"github.com/gogs/git-module"
+
+	"gogs.io/gogs/internal/context"
+	"gogs.io/gogs/internal/db"
+)
+
+type repoGitTreeResponse struct {
```
unknwon: Sorry I just came up a better name for it 😝 
```suggestion
type repoGitTree struct {
```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,78 @@
...
+	Sha  string         `json:"sha"`
+	URL  string         `json:"url"`
+	Tree []*repoGitTree `json:"tree"`
+}
+
+type repoGitTree struct {
```
unknwon:
```suggestion
type repoGitTreeEntry struct {
```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,78 @@
...
+			Sha:  entry.ID.String(),
+			URL:  fmt.Sprintf(templateURL+"/%s", entry.ID.String()),
+		})
+	}
+	results := &repoGitTreeResponse{
+		Sha:  c.Params(":sha"),
+		URL:  fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
+		Tree: children,
+	}
+	c.JSONSuccess(results)
```
unknwon: Again, we can go to the extreme:

```go
c.JSONSuccess( &repoGitTreeResponse{
		Sha:  c.Params(":sha"),
		URL:  fmt.Sprintf(templateURL+"/%s", c.Params(":sha")),
		Tree: children,
	})
```

internal/route/api/v1/repo/tree.go
```diff
@@ -0,0 +1,79 @@
...
+			mode = "040000"
+		case git.ObjectBlob:
+			mode = "120000"
+		case git.ObjectTag:
+			mode = "100644"
+		default:
+			mode = ""
+		}
+		children = append(children, &repoGitTreeEntry{
+			Path: repoPath,
```
unknwon: Since we don't support recursive yet, I believe the `Path` is simply `entry.Name()`.

Exposing server path is actually a security problem :)

unknwon:
FYI, I directly edited the file for some minor changes. Please see my comment above, that should be the last thing we need to address!

bharatnc:
> FYI, I directly edited the file for some minor changes. Please see my comment above, that should be the last thing we need to address!

@unknwon  thanks and I have addressed the last bit of the change that's required. I really appreciate the reviews!

unknwon:
Congrats, merging!

bharatnc:
@unknwon I can work on adding the docs for this new endpoint.

unknwon:
Great :)

PR #5951, opened 2020-02-28
public: fix jquery3.x syntax error

g-u-a-n:
`$.post().success -> $.post().done`

The jqXHR.success(), jqXHR.error(), and jqXHR.complete() callback methods are removed as of jQuery 3.0. Need to use jqXHR.done(), jqXHR.fail(), and jqXHR.always() instead.


unknwon:
Thank you!

Issue #5949, opened 2020-02-28
Support API  `GET /repos/:owner/:repo/contents/:path`


bharatnc:
**Describe the feature**

The gogs content api `GET /repos/:username/:reponame/raw/:ref/:path` returns the raw file contents: https://github.com/gogs/docs-api/blob/master/Repositories/Contents.md.

For my use case, I need  a `Github` style api: https://developer.github.com/v3/repos/contents/#get-contents.

This would be extremely useful because as far as I have explored the gogs api, there isn't a direct way to get all the information about a file / directory by just providing the path to it: 

**Describe the solution you'd like**

Example: `GET /repos/:owner/:repo/readme`. Will return the following response:

```bash
{
  "type": "file",
  "encoding": "base64",
  "size": 5362,
  "name": "README.md",
  "path": "README.md",
  "content": "encoded content ...",
  "sha": "3d21ec53a331a6f037a91c368710b99387d012c1",
  "url": "https://api.github.com/repos/octokit/octokit.rb/contents/README.md",
  "git_url": "https://api.github.com/repos/octokit/octokit.rb/git/blobs/3d21ec53a331a6f037a91c368710b99387d012c1",
  "html_url": "https://github.com/octokit/octokit.rb/blob/master/README.md",
  "download_url": "https://raw.githubusercontent.com/octokit/octokit.rb/master/README.md",
  "_links": {
    "git": "https://api.github.com/repos/octokit/octokit.rb/git/blobs/3d21ec53a331a6f037a91c368710b99387d012c1",
    "self": "https://api.github.com/repos/octokit/octokit.rb/contents/README.md",
    "html": "https://github.com/octokit/octokit.rb/blob/master/README.md"
  }
}
```
(Reference: https://developer.github.com/v3/repos/contents/#get-contents)

There are two ways that we can about to implement this:
1. Modify the existing `/repos/:username/:reponame/raw/:ref/:path` so that it returns a json response along with the content in base64 encoded format.

2. Add a new api endpoint `GET /repos/:owner/:repo/readme` that returns the information about the file along with the contents of the file. Optionally, we can just omit the contents of the file and just return all the other information.

**Describe alternatives you've considered**

I have quite exhaustively used various methods to go about this. Given a repo like:

```
gogs-demo(root)___
                 |
                 dir1  (file1, file3)
                 |
                 README
```

[...]

PR #5963, opened 2020-03-03
api: `GET /repos/:owner/:repo/contents/:path`

bharatnc:
This PR adds support to #5949:
`GET /repos/:owner/:repo/contents/:path`

Curl:

```bash
curl -H "Authorization: token REDACTED"
http://localhost:3000/api/v1/repos/root/testrepo/contents//master/README.md
-X GET  | jq .
```

Curl  Response:

```bash
{
  "type": "blob",
  "size": 12,
  "name": "README.md",
  "path": "README.md",
  "sha": "70fcb456d436f08462602f26df6fb7e167e7a916",
  "url": "http://localhost:3000/api/v1/repos/root/testrepo/contents/README.md",
  "git_url": "http://localhost:3000/api/v1/repos/root/testrepo/trees/70fcb456d436f08462602f26df6fb7e167e7a916",
  "html_url": "http://localhost:3000/api/v1/repos/root/testrepo/tree/70fcb456d436f08462602f26df6fb7e167e7a916",
  "download_url": "http://localhost:3000/api/v1/root/testrepo/raw/README.md",
  "_links": {
    "git": "http://localhost:3000/api/v1/repos/root/testrepo/trees/70fcb456d436f08462602f26df6fb7e167e7a916",
    "self": "http://localhost:3000/api/v1/repos/root/testrepo/contents/README.md",
    "html": "http://localhost:3000/api/v1/repos/root/testrepo/tree/70fcb456d436f08462602f26df6fb7e167e7a916"
  },
  "content": "IyB0ZXN0cmVwbwoK"
}
```


internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+	"net/url"
+
+	"github.com/gogs/git-module"
+	"gogs.io/gogs/internal/context"
+	"gogs.io/gogs/internal/db/errors"
+)
+
+type GitRepoContent struct {
+	Type            string `json:"type"`
+	Target          string `json:"target, omitempty"`
```
golangcibot: structtag: struct field tag `json:"target, omitempty"` not compatible with reflect.StructTag.Get: suspicious space in struct tag value (from `govet`)

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+	Name            string `json:"name"`
+	Path            string `json:"path"`
+	Sha             string `json:"sha"`
+	URL             string `json:"url"`
+	GitURL          string `json:"git_url"`
+	HTMLURL         string `json:"html_url"`
+	DownloadURL     string `json:"download_url"`
+	Links           Links  `json:"_links"`
+	Content         string `json:"content,omitempty"`
+	SubmoduleGitURL string `json:"target,omitempty"`
```
golangcibot: structtag: struct field SubmoduleGitURL repeats json tag "target" also at path.go:15 (from `govet`)

bharatnc:
@unknwon I've implemented the path functionality as per the GitHub API v3. Would appreciate reviews. I have verified that the functionality is working. Thank you!

unknwon:
Thanks, will review it tonight.

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
+package repo
```
unknwon: I think the file should be named `contents.go`

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+
+import (
+	"encoding/base64"
+	"fmt"
+	"net/url"
+
+	"github.com/gogs/git-module"
+	"gogs.io/gogs/internal/context"
+	"gogs.io/gogs/internal/db/errors"
```
unknwon:
```suggestion
	"github.com/gogs/git-module"

	"gogs.io/gogs/internal/context"
	"gogs.io/gogs/internal/db/errors"
```

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+	Path            string `json:"path"`
+	Sha             string `json:"sha"`
+	URL             string `json:"url"`
+	GitURL          string `json:"git_url"`
+	HTMLURL         string `json:"html_url"`
+	DownloadURL     string `json:"download_url"`
+	Links           Links  `json:"_links"`
+	Content         string `json:"content,omitempty"`
+	SubmoduleGitURL string `json:"submodule_git_url,omitempty"`
+	Encoding        string `json:"encoding,omitempty"`
```
unknwon:
```suggestion
type repoContents struct {
	Type            string `json:"type"`
	Target          string `json:"target,omitempty"`
	SubmoduleGitURL string `json:"submodule_git_url,omitempty"`
	Encoding        string `json:"encoding,omitempty"`
	Size            int64  `json:"size"`
	Name            string `json:"name"`
	Path            string `json:"path"`
	Content         string `json:"content,omitempty"`
	Sha             string `json:"sha"`
	URL             string `json:"url"`
	GitURL          string `json:"git_url"`
	HTMLURL         string `json:"html_url"`
	DownloadURL     string `json:"download_url"`
	Links           Links  `json:"_links"`
```

unknwon: It would be ideal to match the same order as GitHub API v3.

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+	// :baseurl/repos/:username/:project/git/trees/:sha
+	templateGitURLLink = "%s/repos/%s/%s/trees/%s"
+	// :baseurl/repos/:username/:project/tree/:sha
+	templateHTMLLLink = "%s/repos/%s/%s/tree/%s"
+)
+
+func GetPathContents(c *context.APIContext) {
+	treeEntry, err := c.Repo.Commit.GetTreeEntryByPath(c.Repo.TreePath)
+	if err != nil {
+		c.NotFoundOrServerError("GetTreeEntryByPath", git.IsErrNotExist, err)
```
unknwon:
```suggestion
		c.NotFoundOrServerError("GetTreeEntryByPath", git.IsErrNotExist, err)
		return
```

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+	templateDownloadURL = "%s/%s/%s/raw/%s"
+	// :base-url/repos/:username/:project/contents/:path
+	templateSelfLink = "%s/repos/%s/%s/contents/%s"
+	// :baseurl/repos/:username/:project/git/trees/:sha
+	templateGitURLLink = "%s/repos/%s/%s/trees/%s"
+	// :baseurl/repos/:username/:project/tree/:sha
+	templateHTMLLLink = "%s/repos/%s/%s/tree/%s"
+)
+
+func GetPathContents(c *context.APIContext) {
```
unknwon:
```suggestion
func GetContents(c *context.APIContext) {
```

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+	treeEntry, err := c.Repo.Commit.GetTreeEntryByPath(c.Repo.TreePath)
+	if err != nil {
+		c.NotFoundOrServerError("GetTreeEntryByPath", git.IsErrNotExist, err)
+	}
+
+	gitURL := fmt.Sprintf(templateGitURLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), treeEntry.ID.String())
+	htmlURL := fmt.Sprintf(templateHTMLLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), treeEntry.ID.String())
+	selfurl := fmt.Sprintf(templateSelfLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), c.Repo.TreePath)
```
unknwon:
```suggestion
	selfURL := fmt.Sprintf(templateSelfLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), c.Repo.TreePath)
```

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+const (
+	// :base-url/:username/:project/raw/:refs/:path
+	templateDownloadURL = "%s/%s/%s/raw/%s"
+	// :base-url/repos/:username/:project/contents/:path
+	templateSelfLink = "%s/repos/%s/%s/contents/%s"
+	// :baseurl/repos/:username/:project/git/trees/:sha
+	templateGitURLLink = "%s/repos/%s/%s/trees/%s"
+	// :baseurl/repos/:username/:project/tree/:sha
+	templateHTMLLLink = "%s/repos/%s/%s/tree/%s"
+)
```
unknwon: It feels less ideal to have package-level constants. I would suggest to hard-code this string literals inside the function body. And refactor later when appropriate.

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+		GitURL:  gitURL,
+		HTMLURL: htmlURL,
+	}
+	// A tree entry can only be one of the following types:
+	// 1. Tree (Directory)
+	// 2. SubModule
+	// 3. SymLink
+	// 4. Blob
+	if treeEntry.IsDir() {
+		dirTree, err := c.Repo.GitRepo.GetTree(treeEntry.ID.String())
```
unknwon: I didn't check, does  `treeEntry` has `ListEntries` method? So we don't need to get the tree again.

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,198 @@
...
+
+		if err != nil {
+			c.ServerError("GetBlobContent", err)
+			return
+		}
+
+		gr.Content = b
+		c.JSONSuccess(gr)
+		return
+	}
```
unknwon: We can check these three cases (with shorter code logic) first, then deal with the `treeEntry.IsDir()` case to save one level of indentation, since that chuck of code is large.

unknwon:
Thanks for the PR! Submitted first round of review.

internal/route/api/v1/repo/path.go
bharatnc: Yes, `contents.go` makes more sense. I've renamed.

internal/route/api/v1/repo/path.go
```diff
@@ -0,0 +1,200 @@
+package repo
+
```
bharatnc: Super weird. I had renamed path.go to contents.go and then started accepting all the UI suggestions and began committing. Looks like all the commits went to path.go file instead of contents.go :(

internal/route/api/v1/repo/path.go
```diff
@@ -11,21 +11,21 @@ import (
 	"gogs.io/gogs/internal/db/errors"
 )
 
-type GitRepoContent struct {
+type repoContents struct {
```
bharatnc: It feels like this should be singluar `repoContent`. Later on in the code we can then return  `[]*repoCpntent` in which case `repoContents` feels a little odd to me.

internal/route/api/v1/repo/path.go
bharatnc: Sure, moved them into the funcs and have left a TODO.

internal/route/api/v1/repo/path.go
bharatnc: Sure makes sense to me. This should be fixed now.

internal/route/api/v1/repo/path.go
bharatnc: I just checked again, it  looks like it doesn't: Here https://github.com/gogs/git-module/blob/master/tree_entry.go or here: https://github.com/gogs/git-module/blob/f29bce1e3a666c02175d080892be185405dd3af4/tree_entry.go. 

It looks like only `Tree` has that method and that too only in this specific commit: https://github.com/gogs/git-module/blob/f29bce1e3a666c02175d080892be185405dd3af4/tree.go#L137

bharatnc:
> Thanks for the PR! Submitted first round of review.

@unknwon  thanks for the review too! I have addressed all of the comments I think. Let me know if you have more suggestions. Thanks!

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+
+		if err != nil {
+			c.ServerError("GetBlobContent", err)
+			return
+		}
+
+		gr.Content = b
+		c.JSONSuccess(gr)
+		return
+	} else { // treeEntry is a directory
```
unknwon: This else case (and its indentation) is actually not needed. We can just do:
```suggestion
	...
	}
	
	// treeEntry is a directory
```

Then we can merge body of `AppendDirTreeEntries` function back to this function.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+	} else if treeEntry.IsLink() {
+		gr.Type = "symlink"
+		gr.Target = c.Repo.TreePath
+		c.JSONSuccess(gr)
+		return
+
+	} else if gr.Type == "blob" { // tree entry is a blob
+		gr.Type = "blob"
+		b, err := getBase64EncodedBlob(c)
+
```
unknwon:
```suggestion
```

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+		return
+
+	} else if treeEntry.IsLink() {
+		gr.Type = "symlink"
+		gr.Target = c.Repo.TreePath
+		c.JSONSuccess(gr)
+		return
+
+	} else if gr.Type == "blob" { // tree entry is a blob
+		gr.Type = "blob"
```
unknwon: I think GitHub call it "file"?

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+			c.NotFoundOrServerError("AppendDirTreeEntries", git.IsErrNotExist, err)
+			return
+
+		}
+		c.JSONSuccess(results)
+		return
+	}
+}
+
+func getBase64EncodedBlob(c *context.APIContext) (string, error) {
```
unknwon: I think having this function independently is not worth it as of this PR.

Should merge back to the `GetContents` function.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+	if c.Repo.Repository.IsBare {
+		return "", errors.New("RepositoryIsBare")
+	}
```
unknwon: This check is not needed, the request would fail even before reaching to this point.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+
+	blob, err := c.Repo.Commit.GetBlobByPath(c.Repo.TreePath)
+	if err != nil {
+		return "", errors.New("ErrorGetBlobByPath")
+	}
+	buf := make([]byte, 1024)
```
unknwon: Why do we want to limit the length? Shouldn't we read everything at once (we need it anyway) then encode in base64?

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+
+	for _, entry := range entries {
+
+		gitURL := fmt.Sprintf(templateGitURLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), entry.ID.String())
+		htmlURL := fmt.Sprintf(templateHTMLLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), entry.ID.String())
+		selfURL := fmt.Sprintf(templateSelfLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), c.Repo.TreePath)
+
+		var contentType string
+		if entry.IsDir() {
+			contentType = "tree"
```
unknwon:
```suggestion
			contentType = "dir"
```

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+		} else if entry.IsSubModule() {
+			contentType = "submodule"
+		} else if entry.IsLink() {
+			contentType = "symlink"
+		} else {
+			contentType = "blob"
```
unknwon:
```suggestion
			contentType = "file"
```

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+	// :baseurl/repos/:username/:project/git/trees/:sha
+	templateGitURLLink := "%s/repos/%s/%s/trees/%s"
+	// :baseurl/repos/:username/:project/tree/:sha
+	templateHTMLLLink := "%s/repos/%s/%s/tree/%s"
+
+	gitURL := fmt.Sprintf(templateGitURLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), treeEntry.ID.String())
+	htmlURL := fmt.Sprintf(templateHTMLLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), treeEntry.ID.String())
+	selfURL := fmt.Sprintf(templateSelfLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), c.Repo.TreePath)
+
+	gr := &repoContents{
```
unknwon:
```suggestion
	contents := &repoContents{
```

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		return "", err
+	}
+	if n >= 0 {
+		buf = buf[:n]
+	}
+	return base64.StdEncoding.EncodeToString(buf), nil
+}
+
+func AppendDirTreeEntries(entries git.Entries, c *context.APIContext) ([]*repoContents, error) {
```
unknwon: Should remove this function and merge the body as mentioned above.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,209 @@
...
+		Path:        c.Repo.TreePath,
+		Sha:         treeEntry.ID.String(),
+		Links: Links{
+			Git:  gitURL,
+			Self: selfURL,
+			HTML: htmlURL,
+		},
+		URL:     selfURL,
+		GitURL:  gitURL,
+		HTMLURL: htmlURL,
```
unknwon: Should respect the order of fields to how they're defined in the struct:
```suggestion
		Size:        treeEntry.Size(),
		Name:        treeEntry.Name(),
		Path:        c.Repo.TreePath,
		Sha:         treeEntry.ID.String(),
		URL:     selfURL,
		GitURL:  gitURL,
		HTMLURL: htmlURL,
		DownloadURL: fmt.Sprintf(templateDownloadURL, c.BaseURL, c.Params(":username"), c.Params(":reponame"), c.Repo.TreePath),
		Links: Links{
			Git:  gitURL,
			Self: selfURL,
			HTML: htmlURL,
		},
```

unknwon: Same for other places.

unknwon:
Thanks for addressing the feedback!

I expect to few more rounds of reviews, be prepared :)

internal/route/api/v1/repo/contents.go
bharatnc: I initially wanted to have it as the same function. But the function grew really big and since smaller funcs usually make things a tad bit clearer, I just extracted it into it's own func. 

I'm tempted to ask if we can have this as a separate func, but sure, I can follow your suggestion if you want me to. Please do let me know.

unknwon: TBH this function is abstracted very well... (at least it should only accept a blob object, which is what it is asking for, not the whole context) Let's put it back first.

bharatnc: Sure, will do that.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,205 @@
...
+
+	results, err := AppendDirTreeEntries(entries, c)
+
+	if err != nil {
+		c.NotFoundOrServerError("AppendDirTreeEntries", git.IsErrNotExist, err)
+		return
+
+	}
+	c.JSONSuccess(results)
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

internal/route/api/v1/repo/contents.go
bharatnc: sure, moved this into the main func.

internal/route/api/v1/repo/contents.go
bharatnc: Fixed.

internal/route/api/v1/repo/contents.go
bharatnc: fixed, thanks!

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,176 @@
...
+			DownloadURL: fmt.Sprintf(templateDownloadURL, c.BaseURL, c.Params(":username"), c.Params(":reponame"), c.Repo.TreePath),
+			Links: Links{
+				Git:  gitURL,
+				Self: selfURL,
+				HTML: htmlURL,
+			},
+		})
+	}
+	c.JSONSuccess(results)
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

internal/route/api/v1/repo/contents.go
bharatnc: merged the func back, it has removed those duplicate vars that I was using more than once in a couple of places. Thanks!

internal/route/api/v1/repo/contents.go
bharatnc: yes, changed.

internal/route/api/v1/repo/contents.go
bharatnc: removed.

internal/route/api/v1/repo/contents.go
bharatnc: removed.

bharatnc:
> Thanks for addressing the feedback!
> 
> I expect to have few more rounds of reviews, be prepared :)

of course! Thanks for the prompt feedback!

bharatnc:
> > Thanks for addressing the feedback!
> > I expect to have few more rounds of reviews, be prepared :)
> 
> of course! Thanks for the prompt feedback!

I plan to test out once more - given all the refactoring that was done - need to make sure that the api works.

bharatnc:
> > > Thanks for addressing the feedback!
> > > I expect to have few more rounds of reviews, be prepared :)
> > 
> > 
> > of course! Thanks for the prompt feedback!
> 
> I plan to test out once more - given all the refactoring that was done - need to make sure that the api works.

Just verified that everything works the sameway after refactoring. Thanks!

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+	"encoding/base64"
+	"fmt"
+	"net/url"
+
+	"github.com/gogs/git-module"
+
+	"gogs.io/gogs/internal/context"
+)
+
+type repoContents struct {
```
unknwon: Sorry you're right, I didn't notice when requesting a directory, we should return `[]*repoContent`.
```suggestion
type repoContent struct {
```

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+	// :baseurl/repos/:username/:project/git/trees/:sha
+	templateGitURLLink := "%s/repos/%s/%s/trees/%s"
+	// :baseurl/repos/:username/:project/tree/:sha
+	templateHTMLLLink := "%s/repos/%s/%s/tree/%s"
+
+	gitURL := fmt.Sprintf(templateGitURLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), treeEntry.ID.String())
+	htmlURL := fmt.Sprintf(templateHTMLLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), treeEntry.ID.String())
+	selfURL := fmt.Sprintf(templateSelfLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), c.Repo.TreePath)
+
+	contents := &repoContents{
```
unknwon:
```suggestion
	content := &repoContent{
```

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+	// :base-url/:username/:project/raw/:refs/:path
+	templateDownloadURL := "%s/%s/%s/raw/%s"
+	// :base-url/repos/:username/:project/contents/:path
+	templateSelfLink := "%s/repos/%s/%s/contents/%s"
+	// :baseurl/repos/:username/:project/git/trees/:sha
+	templateGitURLLink := "%s/repos/%s/%s/trees/%s"
+	// :baseurl/repos/:username/:project/tree/:sha
+	templateHTMLLLink := "%s/repos/%s/%s/tree/%s"
+
+	gitURL := fmt.Sprintf(templateGitURLLink, c.BaseURL, c.Params(":username"), c.Params(":reponame"), treeEntry.ID.String())
```
unknwon: We are getting `c.Params(":username")` and `c.Params(":reponame")` so many times, why not create two variables: `username` and `reponame`?

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+	if treeEntry.IsSubModule() {
+		contents.Type = "submodule"
+		parsedURL, err := url.Parse(c.BaseURL)
+		if err != nil {
+			c.ServerError("ErrorURLParse", err)
+			return
+		}
+		host := parsedURL.Host
+		submoduleURL := fmt.Sprintf("git://%s/%s/%s.git", host, c.Params(":name"), c.Params(":reponame"))
+		contents.SubmoduleGitURL = submoduleURL
```
unknwon:
```suggestion
		contents.SubmoduleGitURL = fmt.Sprintf("git://%s/%s/%s.git", host, c.Params(":name"), c.Params(":reponame"))
```

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+		c.JSONSuccess(contents)
+		return
+
+	} else if treeEntry.Type == "blob" {
+		blob, err := c.Repo.Commit.GetBlobByPath(c.Repo.TreePath)
+		if err != nil {
+			c.ServerError("ErrorGetBlobByPath", err)
+			return
+		}
+		buf := make([]byte, 1024)
```
unknwon: Why do we want to limit to 1024 bytes? Does GitHub have such limit?

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+		return
+	}
+
+	entries, err := dirTree.ListEntries()
+	if err != nil {
+		c.NotFoundOrServerError("ListDirTreeEntries", git.IsErrNotExist, err)
+		return
+	}
+
+	var results = make([]*repoContents, 0, len(entries))
```
unknwon:
```suggestion
	contents := make([]*repoContent, 0, len(entries))
```

unknwon: This line should move after the `if == 0` block.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+	if len(entries) == 0 {
+		c.JSONSuccess(&repoGitTree{})
```
unknwon: Does GitHub API response an empty Git tree for empty directory? What do we get from a GitHub API in this case?

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,177 @@
...
+		return
+	}
+
+	for _, entry := range entries {
+
```
unknwon:
```suggestion
```

internal/route/api/v1/repo/contents.go
bharatnc: Sorry, missed out on your earlier comment. So, when I was figuring out the best way to do this, I referred to how the other api/eps in gogs actually work. I observed that `/internal/route/repo/download.go` (serveData) and also funcs in `internal/route/repo/editor.go` etc use this pattern. So I was just following that. Should I change it / keep it?

internal/route/api/v1/repo/contents.go
bharatnc: Good question, let me check that.

internal/route/api/v1/repo/contents.go
bharatnc: nice! I have creates two variables and used them everywhere else.

internal/route/api/v1/repo/contents.go
bharatnc: Fixed, thanks!

internal/route/api/v1/repo/contents.go
unknwon: Those are for the purpose of detecting the content, but here we want to response the full file content 😅

internal/route/api/v1/repo/contents.go
bharatnc: So @unknwon , I tried creating a empty directory, but looks like you cannot commit an empty directory to git. Also tried committing a file along and later deleting that file to make the directory empty - also not possible.

 But with an empty repo, when I tested I got the following response:
```bash
curl -H "Authorization: token REDACTED" https://api.github.com/repos/bharatnc/test-repo/contents | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   132  100   132    0     0    261      0 --:--:-- --:--:-- --:--:--   261
{
  "message": "This repository is empty.",
  "documentation_url": "https://developer.github.com/v3/repos/contents/#get-contents"
}
```
I think we're almost guaranteed not to get an empty dir, but if we do, we should just respond with a server error I think. Does that sound right to you?

unknwon: Thanks for confirming this, yep, SGTM. Or simpler, just an empty array in JSON `[]`, because it is empty.

bharatnc: Sure just returning an empty array `c.JSONSuccess([]string{})`.

internal/route/api/v1/repo/contents.go
bharatnc: Oh I see I didn't really go in depth about those functions, but good to know -- thanks! I have made the change to read all and encode. I have also tested response, before and after the change, works fine.

bharatnc:
> Almost there
:+1:  - addressed all of the suggestions/changes that were required mostly.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,172 @@
...
+			return
+		}
+		host := parsedURL.Host
+		contents.SubmoduleGitURL = fmt.Sprintf("git://%s/%s/%s.git", host, username, reponame)
+		c.JSONSuccess(contents)
+		return
+
+	} else if treeEntry.IsLink() {
+		contents.Type = "symlink"
+		contents.Target = c.Repo.TreePath
```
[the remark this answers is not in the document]
unknwon: Same in dirTree.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,172 @@
...
+	// A tree entry can only be one of the following types:
+	// 1. Tree (Directory)
+	// 2. SubModule
+	// 3. SymLink
+	// 4. Blob
+	if treeEntry.IsSubModule() {
+		contents.Type = "submodule"
+		parsedURL, err := url.Parse(c.BaseURL)
+		if err != nil {
+			c.ServerError("ErrorURLParse", err)
```
unknwon: As of now, the convention for the message string is the function/method name, so here is `Parse`. Same for other places.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,172 @@
...
+			return
+		}
+		host := parsedURL.Host
+		contents.SubmoduleGitURL = fmt.Sprintf("git://%s/%s/%s.git", host, username, reponame)
```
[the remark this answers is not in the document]
bharatnc: Correct, it doesn't work. I was wanting to leave a TODO based on our previous discussion regarding these links. Thanks!

internal/route/api/v1/repo/contents.go
bharatnc: Oh, I didn't notice it - thanks! Fixed all the messages.

internal/route/api/v1/repo/contents.go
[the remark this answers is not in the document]
bharatnc: Not setting the URL, I've left a TODO in both places. The api response will not contain that field.

internal/route/api/v1/repo/contents.go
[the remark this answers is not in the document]
bharatnc: More specifically, according to the Github API v3: `"target": "/path/to/symlink/target"`. Target is the `blob.Data` if the target is a file - response will be similar to a blob data.

If the target is a directory, then looks like, we would need to just set Target to the path of the dir that is requested and the current format of response seems ok to me.

I will work on it, thanks!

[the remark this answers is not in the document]
bharatnc: For an example in my test repo: https://github.com/bharatnc/test-repo dir2 is a symlink to dir1
```bash
curl -H "Authorization: token REDACTED" https://api.github.com/repos/bharatnc/test-repo/contents/dir1 | jq .

[
  {
    "name": "file1.txt",
    "path": "dir1/file1.txt",
    "sha": "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
    "size": 0,
    "url": "https://api.github.com/repos/bharatnc/test-repo/contents/dir1/file1.txt?ref=master",
    "html_url": "https://github.com/bharatnc/test-repo/blob/master/dir1/file1.txt",
    "git_url": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
    "download_url": "https://raw.githubusercontent.com/bharatnc/test-repo/master/dir1/file1.txt",
    "type": "file",
    "_links": {
      "self": "https://api.github.com/repos/bharatnc/test-repo/contents/dir1/file1.txt?ref=master",
      "git": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
      "html": "https://github.com/bharatnc/test-repo/blob/master/dir1/file1.txt"
    }
  }
]

curl -H "Authorization: token REDACTED" https://api.github.com/repos/bharatnc/test-repo/contents/dir2 | jq .

{
  "name": "dir2",
  "path": "dir2",
  "sha": "df490f837a85cbc0e94a6921286c2f97b8dde399",
  "size": 4,
  "url": "https://api.github.com/repos/bharatnc/test-repo/contents/dir2?ref=master",
  "html_url": "https://github.com/bharatnc/test-repo/blob/master/dir2",
  "git_url": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/df490f837a85cbc0e94a6921286c2f97b8dde399",
  "download_url": "https://raw.githubusercontent.com/bharatnc/test-repo/master/dir2",
  "type": "symlink",
  "target": "dir1",
  "_links": {
    "self": "https://api.github.com/repos/bharatnc/test-repo/contents/dir2?ref=master",
    "git": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/df490f837a85cbc0e94a6921286c2f97b8dde399",
    "html": "https://github.com/bharatnc/test-repo/blob/master/dir2"
  }
}
```
file3 is a symlink to file 1, but the response is as follows:

```
curl -H "Authorization: token REDACTED" https://api.github.com/repos/bharatnc/test-repo/contents/file3.txt | jq .
{
  "name": "file3.txt",
  "path": "file3.txt",
  "sha": "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
  "size": 0,
  "url": "https://api.github.com/repos/bharatnc/test-repo/contents/file3.txt?ref=master",
  "html_url": "https://github.com/bharatnc/test-repo/blob/master/file3.txt",
  "git_url": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
  "download_url": "https://raw.githubusercontent.com/bharatnc/test-repo/master/file3.txt",
  "type": "file",
  "content": "",
  "encoding": "base64",
  "_links": {
    "self": "https://api.github.com/repos/bharatnc/test-repo/contents/file3.txt?ref=master",
    "git": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
    "html": "https://github.com/bharatnc/test-repo/blob/master/file3.txt"
  }
}

curl -H "Authorization: token REDACTED" https://api.github.com/repos/bharatnc/test-repo/contents/file1.txt | jq .

{
  "name": "file1.txt",
  "path": "file1.txt",
  "sha": "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
  "size": 0,
  "url": "https://api.github.com/repos/bharatnc/test-repo/contents/file1.txt?ref=master",
  "html_url": "https://github.com/bharatnc/test-repo/blob/master/file1.txt",
  "git_url": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
  "download_url": "https://raw.githubusercontent.com/bharatnc/test-repo/master/file1.txt",
  "type": "file",
  "content": "",
  "encoding": "base64",
  "_links": {
    "self": "https://api.github.com/repos/bharatnc/test-repo/contents/file1.txt?ref=master",
    "git": "https://api.github.com/repos/bharatnc/test-repo/git/blobs/e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
    "html": "https://github.com/bharatnc/test-repo/blob/master/file1.txt"
  }
}
```

*Note:* The file content is missing as they are all empty.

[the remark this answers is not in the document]
bharatnc: @unknwon  so based on the above behavior of the Gitlab api, it makes me think,
1.  A `blob` even if it's a `symlink` is still treated like a `blob`. Meaning, you still display the file contents with the same response that's for a `blob`.
2. Only a `dir`, if it's of type `symlink` responds with json with `target` set to the actual file name and doesn't display actually display the file tree in the response.

So ideally, we only need this check if the entry is a `dir`:

```golang
else if treeEntry.IsLink() {
		contents.Type = "symlink"
		contents.Target = c.Repo.TreePath
		c.JSONSuccess(contents)
		return

	}
```
If the above check reveals that a `dir` is also a `symlink`, then instead of displaying the entires, we just respond with the json pointing to the **"path of parent"** as `target`.

One **question** that I wanted to ask you is if git-module currently has methods to get the parent path/name for a given symlink?

As far as I have looked, I didn't find any.  If it's there, we can conveniently set that and respond with a valid json. Thanks and also let me know what you think?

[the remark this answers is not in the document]
unknwon: > If the above check reveals that a `dir` is also a `symlink`, then instead of displaying the entires, we just respond with the json pointing to the **"path of parent"** as `target`.

What do you mean by "path of parent"? 

---

The example you gave for "dir2 is a symlink to dir1" is actually request "dir1" not "dir2".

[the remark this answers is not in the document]
bharatnc: > > If the above check reveals that a `dir` is also a `symlink`, then instead of displaying the entires, we just respond with the json pointing to the **"path of parent"** as `target`.
> 
> What do you mean by "path of parent"?
> 
> The example you gave for "dir2 is a symlink to dir1" is actually requesting "dir1" not "dir2".

True, sorry, I have corrected myself above, my question was -  if dir2 is symlink to dir1, and we request for dir2,  is it currently possible to know that dir1 is infact the original file  (that points to dir2) ?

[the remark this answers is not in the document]
unknwon: Symlink is basically a TXT file with special bits, so the content of a symlink file is where it points to.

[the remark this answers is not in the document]
bharatnc: Oh I see! Thank makes perfect sense :). Pardon me for the many redirections, I think that's exactly what you meant in your very first comment in this thread!

> It is not true, it is a symlink so the target is actually the file content (i.e. blob.Data())

Will change the Target to be the blob.Data()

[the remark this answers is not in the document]
bharatnc: Should be fixed.

internal/route/api/v1/repo/contents.go
```diff
@@ -0,0 +1,197 @@
...
+	// 4. Blob
+	if treeEntry.IsSubModule() {
+		// TODO: submoduleURL is not set as current git-module doesn't handle it properly
+		contents.Type = "submodule"
+		c.JSONSuccess(contents)
+		return
+
+	} else if treeEntry.IsLink() {
+		contents.Type = "symlink"
+		blob, err := c.Repo.Commit.GetBlobByPath(c.Repo.TreePath)
```
bharatnc: I'm not sure if the logic for getting blob by path should now be it's own func. I noticed that it's being now used in 4 different places.

unknwon: The API is much cleaner in git-module@v1.0.0 which could directly `[]byte` without a reader, I'll clean these up in migration regardless. Don't worry for now.

bharatnc: Sounds good, thanks!

unknwon:
Congrats!

bharatnc:
> Congrats!

@unknwon thank you very much!

PR #5980, opened 2020-03-10
api: overhaul `/repos/owner/repos/contents`

unknwon:
This is what aligns with https://developer.github.com/v3/repos/contents/#get-contents.

cc @bharatnc since you're using this API, docs updated as well: https://github.com/gogs/docs-api/blob/master/Repositories/Contents.md#get-contents

### DIFF
--- a/internal/context/api.go
+++ b/internal/context/api.go
@@ -39,7 +39,7 @@ func (c *APIContext) Error(status int, title string, obj interface{}) {
 	}
 
 	if status == http.StatusInternalServerError {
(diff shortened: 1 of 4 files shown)


bharatnc:
> This is what aligns with https://developer.github.com/v3/repos/contents/#get-contents.
> 
> cc @bharatnc since you're using this API, docs updated as well: https://github.com/gogs/docs-api/blob/master/Repositories/Contents.md#get-contents

Thank you!

Issue #5985, opened 2020-03-13
Group patterns "" and "/*"  with "/contents" for the contents API 


bharatnc:
**Describe the bug**
 TLDR:  `/contents/*` and `/contents` should both work (Github API v3 works this way)

The contents api should default to the default repository path if `:path` is not present. 
Example:
The following should work and point to the contents at the root of the repo
```bash
http://try.gogs.io/api/v1/repos/testuser/testrepo/contents?ref=master
```

The following should also work and should point to the actual path specified if exists and is valid.
```bash
http://try.gogs.io/api/v1/repos/testuser/testrepo/contents/foo/bar?ref=master
```
Github API v3 works in the above described way: 

```bash
 gogs git:(master) ✗ curl -H "Authorization: token REDACTED" https://api.github.com/repos/bharatnc/test-repo/contents | jq . 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4091  100  4091    0     0   9283      0 --:--:-- --:--:-- --:--:--  9276
[
  {
    "name": "dir1",
    "path": "dir1",
    "sha": "b4f457c219dbb3517be908d4e70f0ada2fd8b8f9",
    "size": 0,
    "url": "https://api.github.com/repos/bharatnc/test-repo/contents/dir1?ref=master",
    "html_url": "https://github.com/bharatnc/test-repo/tree/master/dir1",
 
..... 
OUTPUT TRUNCATED
```

**Gogs version or commit**

On master.

**Expected behavior**

Without providing the `:path` it should automatically point to the root of the repo.

**Actual behavior**
Currently just throws 404 path or not found error.

**Fix**

The fix is actually quite simple in this case. Just add group `"contents"` with `""` and `"/*"`, in `gogs.io/gogs/internal/route/api/v1/api.go`.

```
diff --git a/internal/route/api/v1/api.go b/internal/route/api/v1/api.go
index b54a6df..9839404 100644
--- a/internal/route/api/v1/api.go
+++ b/internal/route/api/v1/api.go
@@ -271,7 +271,10 @@ func RegisterRoutes(m *macaron.Macaron) {
                                }, reqRepoAdmin())
 
                                m.Get("/raw/*", context.RepoRef(), repo.GetRawFile)
-                               m.Get("/contents/*", repo.GetContents)
+                               m.Group("/contents", func() {
+                                       m.Get("", repo.GetContents)
[...]

PR #5986, opened 2020-03-13
api: group "/contents" with "" and "/*" #5985

bharatnc:
This PR groups the contents api endpoint with both `""` and `/*` as
valid URL patterns.

When `""` is the URL pattern, this means that no repo path has'nt been
provided, in which case the path would be the default repo path.

When  `"/*"` is the URL pattern, then this would return the contents at
the specified path if it's valid.

Github API v3 is conformant to this behavior and have verified it
locally.

### DIFF
--- a/internal/route/api/v1/api.go
+++ b/internal/route/api/v1/api.go
@@ -271,7 +271,10 @@ func RegisterRoutes(m *macaron.Macaron) {
 				}, reqRepoAdmin())
 
 				m.Get("/raw/*", context.RepoRef(), repo.GetRawFile)
-				m.Get("/contents/*", repo.GetContents)
+				m.Group("/contents", func() {
+					m.Get("", repo.GetContents)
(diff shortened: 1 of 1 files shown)


bharatnc:
@unknwon minor change that makes contents api aligned with Github API V3. I've tested the change locally and works good. Thanks!

unknwon:
Just to verify, when matching the `""` (no path) one, it returns all entries in the repository root directory?

bharatnc:
> Just to verify, when matching the `""` (no path) one, it returns all entries in the repository root directory?

@unknwon  yes that's correct.

unknwon:
Great, thanks!

Would you mind also update the API docs to add a sentence that now `""` is an option too?

bharatnc:
> Great, thanks!
> 
> Would you mind also update the API docs to add a sentence that now `""` is an option too?

Sure I'll update.

PR #6056, opened 2020-04-02
db: include the Team ID in the error message

ss23:
This means that when using the API to create a new team, the output contains the existing team ID, not just the name.
While there may be the thought that this reveals sensitive information, it is never the case that a user can create or update a team without permission to view the teams in the first place.

One of the reasons for this change is without an existing API to get a list of Teams and their IDs using the API, it would be impossible for an administrator to add themselves to an Owners team (or introspect the Teams at all), as they never know the Teams ID. This is one way to get that functionality without another API call, though that would also be useful.

### DIFF
--- a/internal/db/error.go
+++ b/internal/db/error.go
@@ -400,6 +400,7 @@ func (err ErrLoginSourceInUse) Error() string {
 //              \/     \/      \/
 
 type ErrTeamAlreadyExist struct {
+	ID    int64
 	OrgID int64
 	Name  string
 }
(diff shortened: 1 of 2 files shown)


unknwon:
Thanks for the PR!

Could you explain why do these changes help expose the team ID? I don't see any change on the API layer (i.e. only DB layer), which means the information could be exposed by accident not by design, and no guaranteed it will survive after refactoring.

ss23:
Using the API, there is no other way to have an administrator add a user to the existing Owners team. With this change, you can attempt to create a new Owners team, see the ID, and use that to add a user to it.

There's more work refactoring the API to fix missing API calls, but this is just a quick way to ensure we can continue to use Gogs internally while we send PRs for them.

unknwon:
Thanks for the reply but sorry as I've warned you this is working by accident not by design (i.e. could stop working at anytime). Merging because it is no harm to the system.

ss23:
Thank you, I understand :)

Issue #6067, opened 2020-04-05
Using root to create /data directory in docker causes permission error


eightpigs:
**Describe the bug**

> Repository root path is invalid: mkdir /data/git/gogs-repositories: permission denied

**Gogs version or commit**

[07818d5f](https://github.com/gogs/gogs/commit/07818d5fa5aef7dd7dca1d556f59c7a146a9b00c)

**Git version**

```
$ git version
git version  2.22.2
```

**Operating system**

Alpine Linux (Docker)

**Database**

MySQL

**To Reproduce**
1. Non-root users (docker group) create directories: ./gogs/data
2. The Volumes : `./gogs/data:/data` in docker-compose.yaml
3. Run: docker-compose up
4. Configure and install in web ui : Repository Root Path `/data/git/gogs-repositories`
5. Click "Install Gogs"

**Can you reproduce the bug at https://try.gogs.io?**

no: Installation error.

**Expected behavior**

Normal installation: install successfully.

**Actual behavior**

Installation error.

**Screenshots**
<img width="981" alt="gogs install error" src="https://user-images.githubusercontent.com/7269913/78468633-537ad480-774c-11ea-8a83-589076d3bcc3.png">

eightpigs:
<img width="742" alt="add ssh_key bug" src="https://user-images.githubusercontent.com/7269913/78498752-c7ce6b80-777e-11ea-9708-303ca54aa30d.png">

PR #6068, opened 2020-04-05
docker: fix the bug of using root account to create `/data/*` directory

eightpigs:
Fixes [#6067 ](https://github.com/gogs/gogs/issues/6067)

- Create /data/* directory with git user

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -39,11 +39,11 @@ create_volume_subfolder() {
 }
 
 setids() {
+    export USER=git
     PUID=${PUID:-1000}
     PGID=${PGID:-1000}
-    groupmod -o -g "$PGID" git
-    usermod -o -u "$PUID" git
-    export USER=git
+    groupmod -o -g "$PGID" $USER
+    usermod -o -u "$PUID" $USER
 }
 
 setids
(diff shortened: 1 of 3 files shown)


docker/start.sh
```diff
@@ -43,6 +43,7 @@ setids() {
     PGID=${PGID:-1000}
     groupmod -o -g "$PGID" git
     usermod -o -u "$PUID" git
+    export USER=git
```
unknwon, line 46: We can go one step further (not sure if it will work, looks like so):
```suggestion
    export USER=git
    groupmod -o -g "$PGID" $USER
    usermod -o -u "$PUID" $USER
```

unknwon:
Thanks for the PR! Left a comment.

eightpigs:
During my use, there are also permission errors in adding ssh key. I found that ~git/.ssh was created by root. 

So, I adjusted the *docker/s6/gogs/setup* file:
1. Create ~git/.ssh using git
2. Removed unnecessary permission settings

At present, everything works fine in the container.

unknwon:
Awesome, thanks!

unknwon:
Reverted because of https://github.com/gogs/gogs/commit/f3b05961aae82fa3e19c47f03c78974cf793fda3.

PR #6081, opened 2020-04-08
Force disable visibility for private repos with external wiki or issues

alpencrossi:
PR is related to issue https://github.com/gogs/gogs/issues/5961

Changes in [gogs.js](https://github.com/gogs/gogs/blob/master/public/js/gogs.js) and [options.tmpl](https://github.com/gogs/gogs/blob/master/templates/repo/settings/options.tmpl) are for immediate user feedback in the frontend on what will happen with the settings.


public/js/gogs.js
```diff
@@ -1621,11 +1621,13 @@ $(document).ready(function() {
       $($(this).data("target")).removeClass("disabled");
     } else {
       $($(this).data("target")).addClass("disabled");
+      $($(this).data("uncheck")).prop("checked", false);
     }
   });
   $(".enable-system-radio").change(function() {
     $($(this).data("enable")).removeClass("disabled");
     $($(this).data("disable")).addClass("disabled");
+    $($(this).data("uncheck")).prop("checked", false);
```
unknwon, line 1630: You're so talented to update this mess and make it work 😄 I hate JS. LOL

unknwon:
Thanks and congrats!

public/js/gogs.js
alpencrossi: Not my cup of tea either :laughing: ... anyway, thx

Issue #5739, opened 2019-07-03
README.md symlink should render content


adrianlshaw:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Creating a symlink called README.md in the repository root (that points to another markdown file) should render a preview on the repository web page. 

At the moment, it only displays the name of the symlink in the preview pane.
See screenshot below.

On Github a README symlink to a markdown file will render it on the page.

**Describe alternatives you've considered**
Can't think of any right now.

**Additional context**
![Screenshot from 2019-07-03 21-48-07](https://user-images.githubusercontent.com/762640/60624206-959d7100-9ddc-11e9-86a4-ba928a627c6f.png)

eightpigs:
Supplement, view the raw of the symbol file:

`gogs  | 2020/04/17 11:01:35 [ERROR] [.../internal/route/repo/download.go:54 SingleDownload()] get blob: the entry is not a blob`

![image](https://user-images.githubusercontent.com/7269913/79528176-e1838300-809b-11ea-8a87-c870602450c3.png)

PR #6113, opened 2020-04-17
Support for jumping to source file/dir via symlink.

eightpigs:
Fixed: [#5739 ](https://github.com/gogs/gogs/issues/5739)

### Features

1. The file's symlink can show or jump to source file.
2. The dir's symlink can jump to source dir.
3. The source file/dir for the symlink can be deleted.
4. When viewing a symbolic link, the source file name is displayed (Screenshots 1).
5. Prompt message is displayed when the symbolic link source file is deleted (Screenshots 2).

### Notes

**delete source file**

If the source file/dir of symlink was deleted, the symlink the file will be displayed as a plain text file.
At this point, you can make normal changes to the symbol file.

### Test Cases

1. Soft link file/dir.
2. Delete symlink source file/dir.
3. Jump to source dir by symlink, then new file & upload file.
4. Jump to source file by symlink, then edit or download(raw) it.

### Screenshots

**Feature: 4**
image

**Feature: 5**
image

### DIFF
--- a/internal/route/repo/view.go
+++ b/internal/route/repo/view.go
@@ -28,10 +28,11 @@ import (
 )
 
 const (
-	BARE     = "repo/bare"
-	HOME     = "repo/home"
-	WATCHERS = "repo/watchers"
-	FORKS    = "repo/forks"
+	BARE           = "repo/bare"
+	HOME           = "repo/home"
+	WATCHERS       = "repo/watchers"
+	FORKS          = "repo/forks"
+	SYMLINK_SYMBOL = " -> "
 )
 
 func renderDirectory(c *context.Context, treeLink string) {
@@ -59,6 +60,7 @@ func renderDirectory(c *context.Context, treeLink string) {
 	}
 
 	var readmeFile *git.Blob
+	var readmeFileName string
 	for _, entry := range entries {
 		if entry.IsTree() || !markup.IsReadmeFile(entry.Name()) {
 			continue
@@ -66,6 +68,17 @@ func renderDirectory(c *context.Context, treeLink string) {
 
 		// TODO(unknwon): collect all possible README files and show with priority.
 		readmeFile = entry.Blob()
+		if readmeFile != nil {
+			readmeFileName = readmeFile.Name()
+			sourceEntry, sourcePath, err := isSymlink(entry, c)
+			if err != nil {
+				readmeFileName += SYMLINK_SYMBOL + err.Error()
+			}
+			if sourceEntry != nil {
+				readmeFileName += SYMLINK_SYMBOL + sourcePath
+				readmeFile = sourceEntry.Blob()
+			}
+		}
 		break
 	}
 
@@ -82,7 +95,7 @@ func renderDirectory(c *context.Context, treeLink string) {
 
 		isTextFile := tool.IsTextFile(p)
 		c.Data["IsTextFile"] = isTextFile
-		c.Data["FileName"] = readmeFile.Name()
+		c.Data["FileName"] = readmeFileName
 		if isTextFile {
 			switch markup.Detect(readmeFile.Name()) {
 			case markup.MARKDOWN:
@@ -120,9 +133,8 @@ func renderDirectory(c *context.Context, treeLink string) {
 	}
 }
 
-func renderFile(c *context.Context, entry *git.TreeEntry, treeLink, rawLink string) {
+func renderFile(c *context.Context, entry *git.TreeEntry, fileName string, treeLink, rawLink string) {
 	c.Data["IsViewFile"] = true
-
 	blob := entry.Blob()
 	p, err := blob.Bytes()
 	if err != nil {
@@ -131,7 +143,7 @@ func renderFile(c *context.Context, entry *git.TreeEntry, treeLink, rawLink stri
 	}
 
 	c.Data["FileSize"] = blob.Size()
-	c.Data["FileName"] = blob.Name()
+	c.Data["FileName"] = fileName
 	c.Data["HighlightClass"] = highlight.FileNameToHighlightClass(blob.Name())
 	c.Data["RawFileLink"] = rawLink + "/" + c.Repo.TreePath
 
@@ -274,11 +286,29 @@ func Home(c *context.Context) {
 		return
 	}
 
+	sourceEntry, sourcePath, err := isSymlink(entry, c)
+	entryName := entry.Name()
+	if err != nil {
+		entryName += SYMLINK_SYMBOL + err.Error()
+	} else if sourceEntry != nil {
+		treeLink = branchLink + "/" + sourcePath
+		c.Repo.TreePath = sourcePath
+		entry = sourceEntry
+		entryName += SYMLINK_SYMBOL + sourcePath
+
+		// Required, if not overwritten:
+		// 	dir: It will cause an inability to upload and add files.
+		// 	file: It will not be able to be modified or downloaded(raw).
+		// Symlink only does links, the symbol file cannot be edited.
+		c.Data["TreePath"] = sourcePath
+	}
+
 	if entry.IsTree() {
 		renderDirectory(c, treeLink)
 	} else {
-		renderFile(c, entry, treeLink, rawLink)
+		renderFile(c, entry, entryName, treeLink, rawLink)
 	}
+
 	if c.Written() {
 		return
 	}
@@ -359,4 +389,21 @@ func Forks(c *context.Context) {
 	c.Data["Forks"] = forks
 
 	c.Success(FORKS)
+}
+
+// check and get source file/dir entry for symlink.
+func isSymlink(entry *git.TreeEntry, c *context.Context) (*git.TreeEntry, string, error) {
+	if entry.IsSymlink() {
+		p, err := entry.Blob().Bytes()
+		if err != nil {
+			return nil, "", err
+		}
+		sourcePath := string(p)
+		sourceEntry, err := c.Repo.Commit.TreeEntry("/" + sourcePath)
+		if err != nil {
+			return nil, sourcePath, err
+		}
+		return sourceEntry, sourcePath, nil
+	}
+	return nil, "", nil
 }


eightpigs:
After testing again, some bugs were found (symlink's raw/edit/delete), please wait for me to fix them.
And, I've changed my mind to support recursive acquisition of the final source file.

templates/repo/view_file.tmpl
```diff
@@ -1,15 +1,41 @@
-<div id="file-content" class="{{TabSizeClass .Editorconfig .FileName}}">
+<div id="file-content" class="{{TabSizeClass .Editorconfig .FileName}}">
```
unknwon, line 1:
```suggestion
<div id="file-content" class="{{TabSizeClass .Editorconfig .FileName}}">
```

templates/repo/view_file.tmpl
```diff
@@ -1,15 +1,41 @@
-<div id="file-content" class="{{TabSizeClass .Editorconfig .FileName}}">
+<div id="file-content" class="{{TabSizeClass .Editorconfig .FileName}}"> 
 	<h4 class="ui top attached header" id="{{if .ReadmeExist}}repo-readme{{else}}repo-read-file{{end}}">
 		{{if .ReadmeExist}}
 			<i class="octicon octicon-book"></i>
 			{{if .ReadmeInList}}
 				<strong>{{.FileName}}</strong>
+        {{if .IsSymlink}}
+          <span class="text normal">&nbsp;->&nbsp;<span>
+          {{ if .IsSymlinkSourceExists}}
```
unknwon, line 9: Please use tabs not spaces, I can see they are mixed in this file.

internal/route/repo/view.go
```diff
@@ -49,23 +56,39 @@ func renderDirectory(c *context.Context, treeLink string) {
 	entries.Sort()
 
 	c.Data["Files"], err = entries.CommitsInfo(c.Repo.Commit, git.CommitsInfoOptions{
-		Path:           c.Repo.TreePath,
+		Path:           treePath,
 		MaxConcurrency: conf.Repository.CommitsFetchConcurrency,
 		Timeout:        5 * time.Minute,
 	})
+
```
unknwon, line 63:
```suggestion
```

internal/route/repo/view.go
```diff
@@ -28,14 +28,21 @@ import (
...
 const (
-	BARE     = "repo/bare"
-	HOME     = "repo/home"
-	WATCHERS = "repo/watchers"
-	FORKS    = "repo/forks"
+	BARE           = "repo/bare"
+	HOME           = "repo/home"
+	WATCHERS       = "repo/watchers"
+	FORKS          = "repo/forks"
+	SYMLINK_SYMBOL = " -> "
```
unknwon, line 35: Seems never used?

internal/route/repo/view.go
```diff
@@ -49,23 +56,39 @@ func renderDirectory(c *context.Context, treeLink string) {
...
 	var readmeFile *git.Blob
+	var readmeFileName string
 	for _, entry := range entries {
 		if entry.IsTree() || !markup.IsReadmeFile(entry.Name()) {
 			continue
 		}
 
 		// TODO(unknwon): collect all possible README files and show with priority.
 		readmeFile = entry.Blob()
+		if readmeFile != nil {
```
unknwon, line 78: Do you know at when the `readmeFile` would be nil?

internal/route/repo/view.go
```diff
@@ -360,3 +421,25 @@ func Forks(c *context.Context) {
 
 	c.Success(FORKS)
 }
+
+// check and get source file/dir entry for symlink.
```
unknwon, line 425:
```suggestion
// isSymlink returns destination of given entry if it's a symlink.
```

internal/route/repo/view.go
```diff
@@ -360,3 +421,25 @@ func Forks(c *context.Context) {
...
+func isSymlink(entry *git.TreeEntry, c *context.Context) (ok bool, sourceEntry *git.TreeEntry, sourceTreePath string, errMsg string) {
```
unknwon, line 426: Any reason to use `errMsg string` instead of `err error`?

internal/route/repo/view.go
```diff
@@ -360,3 +421,25 @@ func Forks(c *context.Context) {
...
+	if ok = entry.IsSymlink(); ok {
+		p, err := entry.Blob().Bytes()
+		if err != nil {
+			errMsg = err.Error()
+			return
```
unknwon, line 431: Please return value explicitly with their names.

internal/route/repo/view.go
```diff
@@ -360,3 +421,25 @@ func Forks(c *context.Context) {
 
 	c.Success(FORKS)
 }
+
+// check and get source file/dir entry for symlink.
+func isSymlink(entry *git.TreeEntry, c *context.Context) (ok bool, sourceEntry *git.TreeEntry, sourceTreePath string, errMsg string) {
+	if ok = entry.IsSymlink(); ok {
```
unknwon, line 427: Please use defensive programming:
```go
	if !entry.IsSymlink() {
		return false, ...
	}
	
	...
```

eightpigs:
> 🤔 Why `internal/assets/templates/templates_gen.go` has so many more deletions than the addition? Can you double check nothing was deleted unintentionally?

I merged the master to the current and regenerated it.

unknwon:
> > 🤔 Why `internal/assets/templates/templates_gen.go` has so many more deletions than the addition? Can you double check nothing was deleted unintentionally?
> 
> I merged the master to the current and regenerated it.

Looking at https://patch-diff.githubusercontent.com/raw/gogs/gogs/pull/6113.diff, looks like many template files are deleted. Which go-bindata tool are you using? The one Gogs uses is github.com/kevinburke/go-bindata, which can be installed by `brew install go-bindata` on macOS.

eightpigs:
@unknwon I used `github.com/kevinburke/go-bindata` to regenerate templates_gen.go, please to review.

unknwon:
Thanks @eightpigs!

Which Go version are you using and have you ran `go get -u github.com/kevinburke/go-bindata` to make sure it is up-to-date? The diff in the `template_gen.go` seems like you're using an updated version of it. For example, `s/read/Read/` has been fixed in https://github.com/kevinburke/go-bindata/commit/1f34fcd85389344f29592b7efdefef286c76d7c0.

eightpigs:
Oh! I am too careless: I used curl to download the binaries, the edition is indeed old. Please review  again.

internal/route/repo/view.go
```diff
@@ -34,8 +34,14 @@ const (
 	FORKS    = "repo/forks"
 )
 
-func renderDirectory(c *context.Context, treeLink string) {
-	tree, err := c.Repo.Commit.Subtree(c.Repo.TreePath)
+func renderDirectory(c *context.Context, treeLink string, entry *git.TreeEntry, sourceEntry *git.TreeEntry, sourceTreePath string) {
```
unknwon, line 37:
```suggestion
func renderDirectory(c *context.Context, treeLink string, entry, sourceEntry *git.TreeEntry, sourceTreePath string) {
```

internal/route/repo/view.go
```diff
@@ -59,13 +65,26 @@ func renderDirectory(c *context.Context, treeLink string) {
 	}
 
 	var readmeFile *git.Blob
+	var readmeFileName string
```
unknwon, line 68: Can you explain why is this new variable needed?

internal/route/repo/view.go
```diff
@@ -120,20 +139,39 @@ func renderDirectory(c *context.Context, treeLink string) {
...
 	p, err := blob.Bytes()
 	if err != nil {
 		c.Error(err, "read blob")
 		return
 	}
 
-	c.Data["FileSize"] = blob.Size()
-	c.Data["FileName"] = blob.Name()
+	c.Data["FileSize"] = entry.Blob().Size()
+	c.Data["FileName"] = entry.Blob().Name()
```
unknwon, line 173: Why do we need to change these lines?

internal/route/repo/view.go
```diff
@@ -360,3 +417,23 @@ func Forks(c *context.Context) {
...
+func isSymlink(entry *git.TreeEntry, c *context.Context) (bool, *git.TreeEntry, string, error) {
+	if !entry.IsSymlink() {
+		return false, nil, "", nil
+	}
+	p, err := entry.Blob().Bytes()
+	if err != nil {
+		return true, nil, "", err
+	}
+	sourceTreePath := string(p)
+	sourceEntry, err := c.Repo.Commit.TreeEntry("/" + sourceTreePath)
```
unknwon, line 431: Is prepending "/" required? If so, what is the reason?

internal/route/repo/view.go
```diff
@@ -360,3 +417,23 @@ func Forks(c *context.Context) {
...
+	if err != nil {
+		return true, sourceEntry, sourceTreePath, err
+	}
+	if sourceEntry.IsSymlink() {
+		return isSymlink(sourceEntry, c)
```
unknwon, line 436: Nice :)

internal/route/repo/view.go
```diff
@@ -59,13 +65,26 @@ func renderDirectory(c *context.Context, treeLink string) {
...
 			continue
 		}
 
 		// TODO(unknwon): collect all possible README files and show with priority.
 		readmeFile = entry.Blob()
+		readmeFileName = readmeFile.Name()
+		if ok, sourceEntry, sourcePath, err := isSymlink(entry, c); ok {
+			c.Data["IsSymlinkSourceExists"] = true
+			if err != nil {
+				c.Data["SourceTreePath"] = err.Error()
```
unknwon, line 80: Error should have its own variable, not mixed with regular variable.

internal/route/repo/view.go
eightpigs: Because, if the current file is a symlink, the contents of the source file should be displayed, the blob is modified as the source file.

I think that symlink have no blobs and should be source files.

So:
- `entry.Blob()` is the current file
- `blob` could be a blob of the source file

internal/route/repo/view.go
eightpigs: I want to achieve the following result: **when the source file is not present, the reason for the non-existence can be displayed**.

Of course, this is indeed a bad way, I will try to fix it.


feature_05_the_source_file_not_exists

internal/route/repo/view.go
unknwon: I see. I think this is confusing, we should use `blob` and `sourceBlob`, as we do for `entry` and `sourceEntry`.

PR #6120, opened 2020-04-18
conf: fallback time layout to RFC 3339 when misconfigured

kousikmitra:
Scope:
* If the configuration file has wrong time format then use a default time format.


unknwon:
Thank you!

PR #6122, opened 2020-04-18
db: remove `db.User.LoginType` field

kousikmitra:
Changes:
* Remove db.User.LoginType field
* Use db.User.LoginSource to check if login type is plain login

Manual Test:

- [x] Register First User and Login

- [x] Add New User from Admin Panel and Login

- [x] Sign Up a New User via Register and Login


kousikmitra:
@unknwon Done with the testing and updated the same on PR description.

unknwon:
Thanks, merging!

PR #6170, opened 2020-05-21
lsif: don't upload indexes for forks

gbrik:
Sourcegraph's original published LSIF indexing CI guidelines had an error where the default setting on GitHub workflows would be for forks of the indexed repository to also build and upload LSIF indexes to Sourcegraph.com. The docs have been updated accordingly, and this PR contains the fix.


unknwon:
Thanks!

Issue #6209, opened 2020-06-20
Timestamp doesn’t updates on issue editing


Kosadchiy:
Timestamp db field "updated_unix" doesn’t updates on issue editing.

**Gogs version and commit**
```
Gogs version 0.11.91
```
2fe4d07

**Git version**

```
$ git version 2.24.3
```

**Operating system**
```
Debian GNU/Linux 7
or 
macOS 10.15.5
```

**Database**
```
MySQL 8.0.20
```

**To Reproduce**
1. Create an issue
2. Edit issue and change title
3. Save changes
4. "created_unix" is equal to "updated_unix"

**Can you reproduce the bug at https://try.gogs.io?**
Yes, the API returns JSON where created_at and updated_at are equal after issue changed:
```
[{
    "id": 5809,
    "number": 1,
    "user": {
        ...
    },
    "title": "Test task edited",
    "body": "",
    "labels": [],
    "milestone": null,
    "assignee": null,
    "state": "open",
    "comments": 0,
    "created_at": "2020-06-20T16:56:41Z",
    "updated_at": "2020-06-20T16:56:41Z",
    "pull_request": null
}]
```

**Expected behavior**
"updated_at" field changes after issue title update.

**Actual behavior**
Timestamp doesn’t updates on issue editing.

PR #6210, opened 2020-06-20
issue: fix timestamp updating

Kosadchiy:
Pull request with fix issue timestamp updating [#6209](https://github.com/gogs/gogs/issues/6209).
As you are using xorm events like BeforeUpdate () together with Cols (cols ...) method, it changes only fields wich you pass to Cols () method and BeforeUpdate () doesn’t work.
I think you also should check pull, release and repo updating. It may contains the same problem.

### DIFF
--- a/internal/db/issue.go
+++ b/internal/db/issue.go
@@ -413,7 +413,7 @@ func updateIssueCols(e Engine, issue *Issue) error {
 }
 
 // UpdateIssueCols only updates values of specific columns for given issue.
-func UpdateIssueCols(issue *Issue) error {
+func UpdateIssue(issue *Issue) error {
 	return updateIssueCols(x, issue)
 }
 
--- a/internal/db/issue.go
+++ b/internal/db/issue.go
@@ -407,14 +407,14 @@ func (issue *Issue) ReadBy(uid int64) error {
 	return UpdateIssueUserByRead(uid, issue.ID)
 }
 
-func updateIssueCols(e Engine, issue *Issue, cols ...string) error {
-	_, err := e.ID(issue.ID).Cols(cols...).Update(issue)
+func updateIssueCols(e Engine, issue *Issue) error {
+	_, err := e.ID(issue.ID).Update(issue)
 	return err
 }
 
 // UpdateIssueCols only updates values of specific columns for given issue.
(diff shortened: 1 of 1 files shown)


manfer:
I have not tested but I wonder if you can just solve this with:

```go
func updateIssueCols(e Engine, issue *Issue, cols ...string) error {
        cols = append(cols, "deadline_unix", "updated_unix")
	_, err := e.ID(issue.ID).Cols(cols...).Update(issue)
	return err
}
```

internal/db/issue.go
```diff
@@ -407,14 +407,14 @@ func (issue *Issue) ReadBy(uid int64) error {
 	return UpdateIssueUserByRead(uid, issue.ID)
 }
 
-func updateIssueCols(e Engine, issue *Issue, cols ...string) error {
-	_, err := e.ID(issue.ID).Cols(cols...).Update(issue)
+func updateIssueCols(e Engine, issue *Issue) error {
```
unknwon, line 410: I think the function should be renamed as well
```suggestion
func updateIssue(e Engine, issue *Issue) error {
```

internal/db/issue.go
```diff
@@ -407,14 +407,14 @@ func (issue *Issue) ReadBy(uid int64) error {
...
+	_, err := e.ID(issue.ID).Update(issue)
 	return err
 }
 
 // UpdateIssueCols only updates values of specific columns for given issue.
-func UpdateIssueCols(issue *Issue, cols ...string) error {
-	return updateIssueCols(x, issue, cols...)
+func UpdateIssueCols(issue *Issue) error {
```
unknwon, line 416:
```suggestion
func UpdateIssue(issue *Issue) error {
```

Kosadchiy:
> I have not tested but I wonder if you can just solve this with:
> 
> ```go
> func updateIssueCols(e Engine, issue *Issue, cols ...string) error {
>         cols = append(cols, "deadline_unix", "updated_unix")
> 	_, err := e.ID(issue.ID).Cols(cols...).Update(issue)
> 	return err
> }
> ```

I chosed @manfer decision. Obviosly it is better way. I have tested it works

unknwon (approved):
Thank you!

unknwon:
This patch has been cut with 0.12.1 release.

Issue #4620, opened 2017-07-11
No support of LaTeX rendering in IPython markdown.


kostyfisik:
Hi,

It look`s like no other opened IPython issues on LaTeX support, and there is a problem with a notebook  rendering (after it was implemented in #4070)

I have no local installation of gogs, I just tried it on try.gogs.io.  The problem file 

https://try.gogs.io/k.ladutenko/stud/src/master/fem-1d-linear-full-noJS.ipynb 

It does not render LaTeX equations in Markdown blocks (and LaTeX rendering looks to be fine in the output from SimPy module)

You can check https://github.com/kostyfisik/fem-intro/blob/master/fem-1d-linear.ipynb to see how it should be rendered.  The undrendered example looks like 

So, we can simply substitute $c_0$ with $V_0$

(it is located between Out[4] and In[4] blocks)

Best regards,
Kostya

gbrault:
Hello, do you know if any progress was made on this topic? 

** Support LaTeX rendering in IPython markdown**

gbrault:
I see that MathJax seems not loaded in the notebook viewer. It looks like adding something like
`<script type="text/javascript" src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>`
would do the trick.

However I don't know in which source template we have to inject that? Any guidance please? Thx

unknwon:
@gbrault I believe you can inject this line of code via [Custom Template](https://gogs.io/docs/features/custom_template#inject-code-into-existing-template).

gbrault:
I tried, using even more recent Mathjax script references see https://www.mathjax.org/#gettingstarted

but it doesn't work!

I used injection in the head.tmpl

If you write your own HTML (directly or via a template/theme engine), you can include MathJax by adding this snippet to your page:
<blockquote>
<script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
<script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
Here's a pre-populated example on jsbin you can re-use.
</blockquote>

unknwon:
I think the problem might be at the time of MathJax.js is trying to render, IPython markdown hasn't finished rendering yet.

jianboy:
Looking forward to release soon.

PR #6244, opened 2020-08-03
add latex formula rendering support

mgsky1:
# Goal
This pull request adds the support of rendering latex formulas in IPython notebook.

This improvement support  `$$..$$` and `$...$` expressions.

QQ20200803-234201@2x

# Description
Inspired by [issue #4620](https://github.com/gogs/gogs/issues/4620), I know [`mathjax`](https://www.mathjax.org/) would work. But it has some problems. First, the method described in that issue is not stable, sometime it works while sometime not. Second, that methos is not suitable for business environment, where cannot connect the Internet.

So, I include static JavaScript files of mathjx and use some tricks to improve stability. The basic steps are as follows:
- The mathjax reference is wrapped in `window.onload` statement. 
- And starting to render formulas after 1 second through `setTimeout` method.

# Browser Compatibility
I have tested in Chrome, FireFox, Edge(chromium), Safari. They work fine. But it does not work in IE.

# Note
- Generally, I changed **ONE** file in `gogs/templates/base/footer.tmpl`. Other files are [`offline mathjax files`](https://github.com/mathjax/MathJax-src/releases/tag/3.0.5).
- I use svg to prevent fonts problems.


mgsky1:
> @mgsky1 Thanks! The screenshot looks great 👍
> 
> From a quick scan, I have few questions:
> 
> 1. This PR adds a lot of JS files, are they all required?
> 2. I don't see where these JS files are imported? How does IPythonNotebook knows where to import them?
> 3. Diff of `gogs/templates/repo/view_file.tmpl` looks weird to me, the whole looks changed which I do not that is true. I suspect they have Windows style EOL not Unix. Can you double check this?

Thanks for your appreciation~

I modified and checked the code followed by your advice.

For the Question 1:

After testing, `tex-svg-full.js` is the **only one** JavaScript file required in this feature.

For the Question 2:

The JS file is imported in `view_file.tmpl`. After `Marked.js` finished rendering, MathJax will be injected by Jquery `html` methods.

Code block (in `templates/repo/view_file.tmpl` line 70-72):

```javaScript
// Before: Markdown rendering through Ajax...
// Mathjax Library injection
var latexInject = '<script id="MathJax-script" async src="{{AppSubURL}}/plugins/mathJax-es5-3.0.5/tex-svg-full.js"><\/script>';
$("#latexI").html(latexInject);
```
For the Question 3:

Yes, you are correct. I checked the meta data of `view_file.tmpl`. The EOL of original one is CRLF. I have transformed it to LF in the newest commit.

> **NOTE:**
>
> During my testing process, I found if you do not wrap the formulas in Markdown code block, `Marked.js` will escape `\\` to `\` which cause a problem that matrix and piecewise functions will be rendered in one line.
>
> I tried my best to found the solution. Finally, I think the best one is replacing `\\` to `\\\\` when editing IPythonNotebook. That would be fine both in gogs and IPythonNotebook.

unknwon:
Thanks for the followup!

> For the Question 1:
> 
> After testing, `tex-svg-full.js` is the **only one** JavaScript file required in this feature.

In that case, can you update the PR to only include that file only?

---

image

I'm using Refined GitHub and noticed mixed uses of spaces and tabs (marked in red), please use tabs in all places.

mgsky1:
> Thanks for the followup!
> 
> > For the Question 1:
> > After testing, `tex-svg-full.js` is the **only one** JavaScript file required in this feature.
> 
> In that case, can you update the PR to only include that file only?
> 
> image
> 
> I'm using Refined GitHub and noticed mixed uses of spaces and tabs (marked in red), please use tabs in all places.

Sorry for the indentation. I have fixed it.

mgsky1:
I don't know wether there are some problems in the last commit or I need to do some extra works, because I notice the tag is not updated.

unknwon:
> ...I notice the tag is not updated.

Which "tag" you're referring to?

mgsky1:
> > ...I notice the tag is not updated.
> 
> Which "tag" you're referring to?

The tag "needs followup". I notice that the status of some PRs have been changed to "reviewed".

unknwon:
> > > ...I notice the tag is not updated.
> > 
> > 
> > Which "tag" you're referring to?
> 
> The tag "needs followup". I notice that the status of some PRs have been changed to "reviewed".

Oh... I do that manually 😂 

> > For the Question 1:
> > After testing, `tex-svg-full.js` is the **only one** JavaScript file required in this feature.
> 
> In that case, can you update the PR to only include that file only?

I'm waiting on cleaning up files to be committed in this PR, since you mentioned there is only one file really needed.

PR #6246, opened 2020-08-07
osutil: use system API to get the username when env vars are empty

### LINKED ISSUE #6219: Backup: User configured to run Gogs is "git", but the current user is ""
**Describe the bug**

**Gogs version and commit**

```
/app/gogs/gogs -v
Gogs version 0.12.0+dev
```

**Operating system**

```
/app/gogs $ cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.11.6
```

**Database**

sqlite3

**To Reproduce**

When running `/app/gogs/gogs backup --verbose` I get the following error, even though the user I am running is `git`.

```
/app/gogs $ /app/gogs/gogs backup --verbose
2020/07/02 07:53:32 [FATAL] [gogs.io/gogs/gogs.go:40 main()] Failed to start application: init configuration: user configured to run Gogs is "git", but the current user is ""
/app/gogs $ whoami
git
/app/gogs $ id
uid=1000(git) gid=1000(git) groups=1000(git)
/app/gogs $
```

**Can you reproduce the bug at https://try.gogs.io?**

**Expected behavior**

The backup should be done

**Actual behavior**

Seems like `gogs` is not able to read the user to check against the configured user.

**Screenshots**

**Additional context**

bibo38:
Previously Gogs got the current user via environment variables.
As Golang supports this via the standard library, this function is
now used instead and environment variables are only used if it fails.

Fixes #6219

### DIFF
--- a/internal/osutil/osutil.go
+++ b/internal/osutil/osutil.go
@@ -6,6 +6,7 @@ package osutil
 
 import (
 	"os"
+	"os/user"
 )
 
 // IsFile returns true if given path exists as a file (i.e. not a directory).
@@ -33,12 +34,20 @@ func IsExist(path string) bool {
 	return err == nil || os.IsExist(err)
 }
 
-// CurrentUsername returns the current system user via environment variables.
+// CurrentUsername returns the current system user
 func CurrentUsername() string {
 	username := os.Getenv("USER")
 	if len(username) > 0 {
 		return username
 	}
 
-	return os.Getenv("USERNAME")
+	username = os.Getenv("USERNAME")
+	if len(username) > 0 {
+		return username
+	}
+
+	if user, err := user.Current(); err == nil {
+		username = user.Username
+	}
+	return username
 }
--- a/internal/osutil/osutil_test.go
+++ b/internal/osutil/osutil_test.go
@@ -5,6 +5,7 @@
 package osutil
 
 import (
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -82,4 +83,18 @@ func TestIsExist(t *testing.T) {
 func TestCurrentUsername(t *testing.T) {
 	// Make sure it does not blow up
 	CurrentUsername()
+}
+
+func TestCurrentUsernamePrefersEnvironmentVariable(t *testing.T) {
+	// Some users/scripts expect that they can change the current username via environment variables
+	if userBak, ok := os.LookupEnv("USER"); ok {
+		defer os.Setenv("USER", userBak)
+	} else {
+		defer os.Unsetenv("USER")
+	}
+
+	if err := os.Setenv("USER", "__TESTING::USERNAME"); err != nil {
+		t.Skip("Could not set the USER environment variable:", err)
+	}
+	assert.Equal(t, "__TESTING::USERNAME", CurrentUsername())
 }


unknwon:
Thanks for the PR!

However, please see comments on the prior art about why we aren't using stdlib for this: https://github.com/gogs/gogs/pull/5544#issuecomment-445019165

bibo38:
> However, please see comments on the prior art about why we aren't using stdlib for this: #5544 (comment)

First of all: This PR only uses `user.Current()` as an preferred way of getting the username. If this fails, the currently existing code is still invoked. So we just gain the correct username, if it's implemented and keep the current status if it's not implemented.

Furthermore I've looked at the Go sources and it seems, that this missing implementation without cgo isn't even a problem anymore. See [lookup_stubs.go](https://golang.org/src/os/user/lookup_stubs.go#L21) for the non-cgo implementation (and [lookup_unix.go](https://golang.org/src/os/user/lookup_unix.go) for the parsing of the passwd-file in pure Golang). Given that the Copyright comments indicate 2016, i would suggest, that probably most of actively used Golang releases should include this functionality. 

So I've tested this with the following program on an Windows 10 x64 machine with Golang 1.14:
```go
package main

import (
        "fmt"
        "os/user"
)

func main() {
        user, err := user.Current()
        if err != nil {
                fmt.Println("Error getting the current username:", err)
        } else {
                fmt.Println("Your username is:", user.Username)
        }
}
```
I've built it using the following commands:
```cmd
set GOOS=linux
set GOARCH=amd64
go build testuser.go
```
> I've also tested the commands with the flag `CGO_ENABLED=1`, which resulted in an expected compilation error, due to the missing linux headers.

Running this on Linux gives the correct username, even if the `USER` environment variable differs (which is also used by the stdlib, if the `passwd` file cannot be found). Using strace, I could also ensure, that the program gets the UID and then directly opens the `/etc/passwd` file:
```
futex(0xc000100148, FUTEX_WAKE_PRIVATE, 1) = 1
fcntl(1, F_GETFL)                       = 0x2 (flags O_RDWR)
fcntl(2, F_GETFL)                       = 0x2 (flags O_RDWR)
getuid()                                = 1000
openat(AT_FDCWD, "/etc/passwd", O_RDONLY|O_CLOEXEC) = 3
epoll_create1(EPOLL_CLOEXEC)            = 4
pipe2([5, 6], O_NONBLOCK|O_CLOEXEC)     = 0
epoll_ctl(4, EPOLL_CTL_ADD, 5, {EPOLLIN, {u32=5898192, u64=5898192}}) = 0
epoll_ctl(4, EPOLL_CTL_ADD, 3, {EPOLLIN|EPOLLOUT|EPOLLRDHUP|EPOLLET, {u32=703278824, u64=139990867324648}}) = -1 EPERM (Operation not permitted)
epoll_ctl(4, EPOLL_CTL_DEL, 3, 0xc000122bd4) = -1 EPERM (Operation not permitted)
read(3, "root:x:0:0::/root:/bin/bash\nbin:"..., 4096) = 1613
close(3)                                = 0
```

unknwon:
@bibo38: Thanks for the nice write up!

One last thing to is that over the years, users have learned to use `$USER` to fool Gogs about the current user, so the call of `user.Current()` should invoked after the `$USER` env variable is empty. Otherwise, the code might break many existing installations.

bibo38:
> user.Current() should invoked after the $USER env variable is empty.

I've fixed it

internal/osutil/osutil.go
```diff
@@ -33,12 +34,20 @@ func IsExist(path string) bool {
 	return err == nil || os.IsExist(err)
 }
 
-// CurrentUsername returns the current system user via environment variables.
+// CurrentUsername returns the current system user
```
unknwon, line 37:
```suggestion
// CurrentUsername returns the username of the current user.
```

unknwon (approved):
Thanks!

PR #6288, opened 2020-08-28
docker: revert shellcheck fix

### LINKED ISSUE #6287: Docker: Backup Rotator
After pulling the latest docker image the backup rotator scripts creates an error:
find: unrecognized: -mtime +360

I think the quotes added by the the commit 92c5f94fe7fae17620ecb8895e287e9e66fb135c introduced this error:
https://github.com/gogs/gogs/blob/6e1c74a0f25032f43a40c71b78616d7960b3d832/docker/runtime/backup-rotator.sh#L24

The argument "FIND_EXPRESSION" is not correctly passed to the find commando.
Removing the quotes seems to fix this error.

unknwon:
Fixes #6287

### DIFF
--- a/docker/runtime/backup-rotator.sh
+++ b/docker/runtime/backup-rotator.sh
@@ -21,7 +21,8 @@ main() {
 		exit 1
 	fi
 
-	find "${BACKUP_PATH}/" -type f -name "gogs-backup-*.zip" "-${FIND_EXPRESSION}" -print -exec rm "{}" +
+	# shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


xZero707:
I just noticed errors in my container and started investigating. I thought I introduced this in my PR and kinda panicked. 
I know that shellcheck does report this as error, but if you just wrap it as suggested, then bash will not treat is as argument. 

What could work, but I haven't tried:
```
-"${FIND_EXPRESSION}"
```

Anyway, I should have documented this in the code. 

By the way, I can help with Docker/shell related stuff. Just assign appropriate issues to me, no need to ask.

unknwon:
@xZero707 Thanks for catching up! Yeah, I wondered `-"${FIND_EXPRESSION}"` might work too 😄

PR #6176, opened 2020-05-27
repo: support unlisted but publicly accessible repositories

achilleas-k:
This PR adds support for publicly accessible but unlisted (secret) repositories (issue #5733).

## Forms and UI

I'm including screenshots to show how the new checkbox appears in all relevant forms.

### Creation
new-repository
new-fork
new-migration

### Settings
repo-settings

### Settings (forked)
repo-settings-with-fork

### Icon

Note that, as I mentioned in the relevant issue, I included the icon for testing but it should probably be replaced by something more appropriate, like the [eye-closed](https://primer.style/octicons/eye-closed) octicon in newer versions, or anything you might feel is more appropriate..

repo-home
repo-list

---

## Behaviour

Unlisted repositories behave like Private repositories on pages that list repositories.  The difference is that users who know the URL are allowed to view and clone them.  The general idea is that Unlisted repositories should be "secret", in that no user should be able to discover them, but users who know about them can read them.  In the interest of clarity and posterity, below I list how Unlisted repositories behave in all relevant cases.

For the owner of an unlisted repository:
- They appear on the owner's repository list page.
- They appear on the explore page (and search results).

For a collaborator of an unlisted repository:
- They **do not** appear on the owner's repository list page.
- They appear on the explore page (and search results).

For owners of organisations with unlisted repositories:
- They appear on an organisation's repository list page.

For organisation team members with access to unlisted repositories:
- They appear on an organisation's repository list page.

For other users and guest (non-logged-in) users:
- They **do not** appear on the owner's repository list page.
- They **do not** appear on the explore page (and search results).

General:
- They **do not** appear in the owner's public activity (actions).
- Repositories can be created unlisted.
- Migrations can be created unlisted.
- Forking an unlisted repository copies the unlisted property and the property can't be changed in the fork.
- Changing a repository's unlisted status changes the status for all forks of the repository.
- Changing a repository's unlisted status changes the visibility of associated actions.

---

## API

Adding the Unlisted property to the API requires editing the structs in the go-gogs-client, which I'll do later and then reflect the changes here as necessary.

PR #6305, opened 2020-09-03
repo: set default value for `is_unlisted` column

unknwon:
Follow up of #6176, otherwise the existing rows will have NULL and makes the `is_unlisted = false` condition ineffective.

cc @achilleas-k


achilleas-k:
Thanks for the ping.  Sorry I didn't catch this.

unknwon:
@achilleas-k No worries, it is more of an ORM trick...

PR #6349, opened 2020-09-24
Remove branch prefix from SHA URL

zhukovra:
Resolves #6300

PR #6350, opened 2020-09-25
markup: render SHA links without branch prefix

laojianzi:
Refactoring PR #6349

Co-authored-by: Zhukov Roman <zhukov.roman@gmail.com>


internal/db/repo.go
```diff
@@ -429,24 +429,26 @@ func (repo *Repository) UpdateSize() error {
...
-		return nil
-	} else if repo.ExternalMetas == nil {
-		repo.ExternalMetas = map[string]string{
-			"format": repo.ExternalTrackerFormat,
-			"user":   repo.MustOwner().Name,
-			"repo":   repo.Name,
-		}
+	if repo.ExternalMetas == nil {
+		repo.ExternalMetas = make(map[string]string)
+	}
```
unknwon, line 436: We do not need to re-create metas map every time this method is being called:
```suggestion
	if repo.ExternalMetas != nil {
		return repo.ExternalMetas
	}
	
	repo.ExternalMetas = make(map[string]string)
```

internal/markup/markup.go
```diff
@@ -140,12 +140,22 @@ func RenderCrossReferenceIssueIndexPattern(rawBytes []byte, urlPrefix string, me
...
+		if len(metas) > 0 {
+			user := metas["user"]
+			repo := metas["repo"]
+
+			if user != "" && repo != "" {
+				urlPrefix = fmt.Sprintf("/%s/%s", user, repo)
+			}
+		}
+
+		return fmt.Sprintf(`<a href="%s/commit/%s"><code>%s</code></a>`, urlPrefix, m, tool.ShortSHA1(m))
```
unknwon, line 158: Since this link is broken anyway for SHA1 links, I don't think this fall back does any good, thus can remove it.

internal/db/repo.go
```diff
@@ -429,24 +429,26 @@ func (repo *Repository) UpdateSize() error {
...
+
+	repo.ExternalMetas["user"] = repo.MustOwner().Name
+	repo.ExternalMetas["repo"] = repo.Name
```
unknwon, line 439: Use the result of [`repo.Link`](https://github.com/gogs/gogs/blob/211a1394b86dafc90d73d5d622ca85686c2fb5f6/internal/db/repo.go#L544-L546) is better here, that's all we want.
```suggestion
	repo.ExternalMetas["repoLink"] = repo.Link()
```

internal/markup/markup.go
```diff
@@ -140,12 +140,22 @@ func RenderCrossReferenceIssueIndexPattern(rawBytes []byte, urlPrefix string, me
...
-		return fmt.Sprintf(`<a href="%s/commit/%s"><code>%s</code></a>`, urlPrefix, m, tool.ShortSHA1(string(m)))
+
+		if len(metas) > 0 {
+			user := metas["user"]
+			repo := metas["repo"]
+
+			if user != "" && repo != "" {
+				urlPrefix = fmt.Sprintf("/%s/%s", user, repo)
+			}
+		}
```
unknwon, line 156: It is generally safe to assume `metas` won't be empty.

```suggestion
		return fmt.Sprintf(`<a href="%s/commit/%s"><code>%s</code></a>`, metas["repoLink"], m, tool.ShortSHA1(string(m)))
```

internal/markup/markup.go
```diff
@@ -160,7 +170,7 @@ func RenderSpecialLink(rawBytes []byte, urlPrefix string, metas map[string]strin
 
 	rawBytes = RenderIssueIndexPattern(rawBytes, urlPrefix, metas)
 	rawBytes = RenderCrossReferenceIssueIndexPattern(rawBytes, urlPrefix, metas)
-	rawBytes = RenderSha1CurrentPattern(rawBytes, urlPrefix)
+	rawBytes = RenderSha1CurrentPattern(rawBytes, urlPrefix, metas)
```
unknwon, line 173: In fact, I think the only change we need in this file is this:

```suggestion
	rawBytes = RenderSha1CurrentPattern(rawBytes, metas["repoLink"])
```

unknwon (changes requested):
Thanks! Please see inline comments.

laojianzi:
> Thanks! Please see inline comments.

Thanks for your review!

unknwon (approved):
Thank you!

unknwon:
@laojianzi If you're interested, the next step would be refactor the repository metas to be a concrete struct, instead of a `map[string]string`.

laojianzi:
> @laojianzi如果您有兴趣，下一步将是将存储库meta重构为一个具体的结构，而不是一个`map[string]string`。

@unknwon Nice, @me 👍

Issue #3525, opened 2016-08-27
Cookie security


GinoHereIam:
- Gogs version (or commit ref): 0.9.83.0816
- Git version: 2.7.4
- Operating system: CentOS
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [x] MySQL
  - [ ] SQLite
- Can you reproduce the bug at https://try.gogs.io:
  - [ ] Yes (provide example URL)
  - [ ] No
  - [x] Not relevant
- Log gist:

## Description

Cookie security, may it be able that you can set the cookies "httponly secure"? Right now I don't see a possibility to overwrite the cookie attributes.

Thanks
Gino
...

Unknwon:
>set the cookies

Which cookie?

GinoHereIam:
I mean generally all cookies which are set. 
When I tested my website with the new security check tool, it figured out that all header securities were missing. [https://observatory.mozilla.org](https://observatory.mozilla.org)
Though, I didn't see any issue ticket here about cookie headers. That's why I did it put up here.

Unknwon:
Any real example you can attack Gogs site with current cookie?

GinoHereIam:
No, but why should be there possibility to give this option? 
[https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet)

Unknwon:
You can enable this for session data yourself: https://github.com/gogits/gogs/blob/master/conf/app.ini#L243

Unknwon:
OK, found some inconsistency cross the app, will fix all at once but not now.

GinoHereIam:
:( don't see reason why you look for examples not to implement a little security feature. 
Here is a real world example:
[https://blog.codinghorror.com/protecting-your-cookies-httponly/](https://blog.codinghorror.com/protecting-your-cookies-httponly/)

silverwind:
Regarding cookie security attributes, there are these three attributes right now:

- `Secure` - only transmit cookie over HTTPS
- `HttpOnly` - hides the cookie from JavaScript
- `SameSite=strict` - declare the cookie as first-party only

I think both `HttpOnly` and `SameSite=strict` are generally good ideas. `Secure` could be left out except when configured for pure HTTPS.

Unknwon:
[...]

PR #6358, opened 2020-09-27
Add ability to set SameSite attribute of cookie

lfuelling:
Hi,
this PR is somewhat related to #3525 and adds the ability to set the `SameSite` attribute for both the session cookies and the CSRF cookie.

This PR depends on go-macaron/csrf#15 and thus go-macaron/macaron#206 because the version in `go.mod` has to be lifted once those PRs are merged.

This implementation is using the same logic/style as go-macaron/session#38 where `false` is `Lax` and `true` is `Strict`. The default is again `false`.


lfuelling:
I'll take a closer look at go-macaron/macaron#208 and update this PR later this week.

PR #6378, opened 2020-10-08
app.ini: remove default mailer.USER to allow unauth SMTP

mcluseau:
Having a default user in the mailer section prevent from using
unauthenticated SMTP as mailer's Send function expects an empty USER in
this case. Even a explicit empty value in the custom configuration
doesn't erase that default.


unknwon:
Thanks for the PR and sorry about the late catch-up! Unfotunately, we don't accept PRs sending to release branches.

PR #6386, opened 2020-10-14
repo: fix 500 on /watchers & /stars for mssql

### LINKED ISSUE #5482: mssql: Error 500 on /watchers & /stars
- Git version: 2.19.1
- Operating system: Windows
- Database (use `[x]`):
  - [ ] PostgreSQL
  - [ ] MySQL
  - [X] MSSQL
  - [ ] SQLite
  - [X] No
  - [ ] Not relevant

> 2018/10/16 07:14:21 [ERROR] [.../routes/repo/view.go:331 RenderUserCards()] getter: mssql: Cannot call methods on nvarchar.

## Description

This is the same issue as #2176, but with mssql instead of postgresql.

The watchers page runs a query like this with mssql:

```sql
SELECT  TOP 40 * FROM "user" LEFT JOIN "watch" ON user.id=watch.user_id WHERE watch.repo_id=1
```

That results in this error:

> Cannot call methods on nvarchar.

The fix in commit https://github.com/gogs/gogs/commit/91789930bcc9d5796a1b06bdeb1402106c0d18d2 added double quotes to escape the "user" alias just when `setting.UsePostgreSQL` is true.  That same fix should work here if it's updated to use that quoted version when `setting.UseMSSQL` is true.

laojianzi:
Resolves #5482

Ref: 91789930bcc9d5796a1b06bdeb1402106c0d18d2

### DIFF
--- a/internal/db/repo.go
+++ b/internal/db/repo.go
@@ -2285,7 +2285,7 @@ func GetWatchers(repoID int64) ([]*Watch, error) {
 func (repo *Repository) GetWatchers(page int) ([]*User, error) {
 	users := make([]*User, 0, ItemsPerPage)
 	sess := x.Limit(ItemsPerPage, (page-1)*ItemsPerPage).Where("watch.repo_id=?", repo.ID)
-	if conf.UsePostgreSQL {
+	if conf.UsePostgreSQL || conf.UseMSSQL {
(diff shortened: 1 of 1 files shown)


unknwon (approved):
Thank you!

PR #6389, opened 2020-10-19
Update CHANGELOG.md

RDCH106:
Fixed a couple of incorrections detected in `CHANGELOG.md`.

It may be convenient to update [releases notes](https://github.com/gogs/gogs/releases) because in the 0.13 version some configuration options will be deprecated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,6 @@ All notable changes to Gogs are documented in this file.
 - Configuration section `[service]` is no longer used.
 - Configuration option `APP_NAME` is no longer used.
 - Configuration option `[security] REVERSE_PROXY_AUTHENTICATION_USER` is no longer used.
-- Configuration option `[database] PASSWD` is no longer used.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -24,7 +24,6 @@ All notable changes to Gogs are documented in this file.
 - Configuration section `[service]` is no longer used.
 - Configuration option `APP_NAME` is no longer used.
 - Configuration option `[security] REVERSE_PROXY_AUTHENTICATION_USER` is no longer used.
-- Configuration option `[database] PASSWD` is no longer used.
```
unknwon, line 27: I don't think this is a valid deletion?

unknwon (changes requested):
Thanks :)

CHANGELOG.md
RDCH106: Why not?

It seems duplicated:

imagen

I can't see the difference... 😅

unknwon: Oh, OK, didn't see it in diffs 😅

unknwon (approved):

Issue #6413, opened 2020-11-11
SSRF in repository migrate via CRLF injection


unknwon:
This is a placeholder issue for CRLF injection in Git which leads to SSRF using repository migrate feature.

Credit to @stypr

stypr:
Marked CVE-2021-32408

PR #6420, opened 2020-11-16
repo: disallow urlencoded new lines in git protocol paths

stypr:
This will fix the issue of #6413 

Related reference: https://github.com/go-gitea/gitea/pull/13525

### DIFF
--- a/internal/form/repo.go
+++ b/internal/form/repo.go
@@ -72,6 +72,7 @@ func (f MigrateRepo) ParseRemoteAddr(user *db.User) (string, error) {
 		if len(f.AuthUsername)+len(f.AuthPassword) > 0 {
 			u.User = url.UserPassword(f.AuthUsername, f.AuthPassword)
 		}
+		// To prevent CRLF injection in git protocol, see https://github.com/gogs/gogs/issues/6413
 		if u.Scheme == "git" && (strings.Contains(remoteAddr, "%0d") || strings.Contains(remoteAddr, "%0a")) {
 			return "", db.ErrInvalidCloneAddr{IsURLError: true}
 		}
--- a/internal/form/repo.go
+++ b/internal/form/repo.go
@@ -72,6 +72,9 @@ func (f MigrateRepo) ParseRemoteAddr(user *db.User) (string, error) {
 		if len(f.AuthUsername)+len(f.AuthPassword) > 0 {
 			u.User = url.UserPassword(f.AuthUsername, f.AuthPassword)
 		}
+		if u.Scheme == "git" && u.Port() != "" && (strings.Contains(remoteAddr, "%0d") || strings.Contains(remoteAddr, "%0a")) {
+			return "", db.ErrInvalidCloneAddr{IsURLError: true}
+		}
 		remoteAddr = u.String()
 	} else if !user.CanImportLocal() {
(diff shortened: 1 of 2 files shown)


internal/form/repo.go
```diff
@@ -72,6 +72,9 @@ func (f MigrateRepo) ParseRemoteAddr(user *db.User) (string, error) {
 		if len(f.AuthUsername)+len(f.AuthPassword) > 0 {
 			u.User = url.UserPassword(f.AuthUsername, f.AuthPassword)
 		}
+		if u.Scheme == "git" && u.Port() != "" && (strings.Contains(remoteAddr, "%0d") || strings.Contains(remoteAddr, "%0a")) {
```
unknwon, line 75: Why is checking `u.Scheme` and `u.Port()` necessary? Shouldn't we disallow malicious chars in all cases?

stypr: I wouldn't say it is necessary, however other protocols have this filtered properly so I guess it is ok to proceed as is. Do you have any alternative plan? By default, git:// protocol will only use 9418 port for transmitting the data.

unknwon: I see... checking the protocol makes sense now. As for port, can I use urlencoded new lines without specifying a port?

stypr: yes. there seems no problem assuming that nobody uses crlf for a repository name.

unknwon: What I meant was we always need to check _urlencoded new lines_, with or without having a port in the clone URL, the port doesn't play a role here. Does this sound correct to you?

stypr: Yes. That sounds correct. We're on the same page.
If you think there's no need of applying the port number, I think it's fine to delete it in this case.

Initially, this will only cause some impact when the attacker is capable of attacking other ports. 
But as you mentioned, blocking all ports with %0d and %0a also sounds good to me.

stypr: Updated: https://github.com/gogs/gogs/pull/6420/commits/eea3d678561b885cbcce3c87077927d308abe340

internal/form/repo.go
```diff
@@ -72,6 +72,9 @@ func (f MigrateRepo) ParseRemoteAddr(user *db.User) (string, error) {
 		if len(f.AuthUsername)+len(f.AuthPassword) > 0 {
 			u.User = url.UserPassword(f.AuthUsername, f.AuthPassword)
 		}
+		if u.Scheme == "git" && (strings.Contains(remoteAddr, "%0d") || strings.Contains(remoteAddr, "%0a")) {
```
unknwon, line 75:
```suggestion
		// To prevent CRLF injection in git protocol, see https://github.com/gogs/gogs/issues/6413
		if u.Scheme == "git" && (strings.Contains(remoteAddr, "%0d") || strings.Contains(remoteAddr, "%0a")) {
```

unknwon (approved):
Thank you!

unknwon:
Ah, one last thing, would you mind adding a CHANGELOG entry here? https://github.com/gogs/gogs/blob/main/CHANGELOG.md#fixed

stypr:
Does that look good?

unknwon:
Looks great 👍

PR #6421, opened 2020-11-17
Expect personal access token in password field, not username field

### LINKED ISSUE #3866: Use personal access tokens as password
GitHub allows the usage of personal access tokens as password (as per https://help.github.com/articles/which-remote-url-should-i-use/#cloning-with-https-urls-recommended)

It'd be great if Gogs could also allow this, so that I can give the token to an application I want to use, rather than my actual password.

Many thanks for the great software!

bmgxyz:
Fixes #3866

Users with 2FA enabled rely on personal access tokens to authenticate on the command line. Gogs supports this feature, but the implementation expects the user to enter their token in the username field and leave the password field blank. This is poor security because the token is meant to be a secret, but neither Git nor Gogs treat the contents of the username field as a secret.

This small change fixes this behavior by expecting the user to input their username in the username field and their personal access token in the password field. The authentication attempt only succeeds if both fields are correct. Note that this is the behavior on GitHub, as far as I can tell.

I'm quite new to Go, so I welcome any edits or suggestions.

### DIFF
--- a/internal/route/repo/http.go
+++ b/internal/route/repo/http.go
@@ -129,15 +129,15 @@ func HTTPContexter() macaron.Handler {
 			return
 		}
 
-		// If username and password combination failed, try again using username as a token.
+		// If username and password combination failed, try again using password as a token.
 		if authUser == nil {
-			token, err := db.AccessTokens.GetBySHA(authUsername)
+			token, err := db.AccessTokens.GetBySHA(authPassword)
 			if err != nil {
 				if db.IsErrAccessTokenNotExist(err) {
 					askCredentials(c, http.StatusUnauthorized, "")
 				} else {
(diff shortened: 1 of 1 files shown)


PetervdPerk:
Great contribution, was trying configure my gogs server to let it behave like Github 2FA but it seems it's not a configuration issue but more an implementation issue.

Is there any reason why is isn't in upstream yet, who are the maintainers of this project I see @unknwon  is approving pull requests.

unknwon (changes requested):
Thanks for the PR!

This is a great improvement, but also introduces break changes, i.e. people are relying on passing access token via username field is now failing.

I think we should support passing access token in both fields for backward compatibility.

unknwon:
Superseded by https://github.com/gogs/gogs/pull/7198.

Issue #6434, opened 2020-12-01
Allow setting allowed SSH MAC algorithms


eduardok:
golang's crypto module (used by gogs with its default MACs) defines hmac-sha1-96 that is considered weak by security scanning tools such as Tenable (https://www.tenable.com/plugins/nessus/71049), so we need a way to select with algorithms to use.
I already have this code in my fork.

PR #6435, opened 2020-12-01
ssh: allow setting allowed MAC algorithms for built-in server

eduardok:
golang's crypto module (used by gogs with its default MACs) defines hmac-sha1-96 that is considered weak by security scanning tools such as Tenable (https://www.tenable.com/plugins/nessus/71049).
This commit's app.ini drops hmac-sha1-96 given that it is the safer option.
New configuration setting SSH_SERVER_MACS allows one to put hmac-sha1-96 back if necessary (unlikely).

Fixes issue #6434.


internal/ssh/ssh.go
```diff
@@ -162,10 +162,11 @@ func listen(config *ssh.ServerConfig, host string, port int) {
 }
 
 // Listen starts a SSH server listens on given port.
-func Listen(host string, port int, ciphers []string) {
+func Listen(host string, port int, ciphers []string, macs []string) {
```
unknwon, line 165:
```suggestion
func Listen(host string, port int, ciphers, macs []string) {
```

unknwon (approved):
Thank you!

PR #6451, opened 2020-12-13
Use go1.16 official static embed feature instead go-bindata to embed static files

alimy:
* Use go1.16 official static embed feature instead go-bindata to embed static files.
* go1.16 go:embed feature is very nice to use and can try it in go dev tip now.
* ***Need wait go1.16 release***.


alimy:
Optimize implement by  #6628

PR #6463, opened 2021-01-01
Static Resources HTTP Response Header "Expire" Setting

dongtaolqq2020:
add optional HTTP Response Header  Expire Setting.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,24 @@
+name: Publish Docker image
```
unknwon, line 1: Is this file included by mistake?

unknwon (changes requested):

Issue #6470, opened 2021-01-11
API - Add a call to search for a username based on email address


ss23:
The API currently lacks a way to find a username from an email address.

There are two solutions that come to mind:
* Modify `GET /users/search` to allow searching based upon email, not just username
* Add a new API call, `GET /users/search/email` or similar

I am hesitant to suggest a new API endpoint, such as `GET /users/emails/search`, because this is likely to break in the case of an existing user with the username "emails" (or whichever other endpoint that is chosen.

Once an approach is chosen, I am more than happy to implement the required change, as it should be minimal and quick to do.
@unknwon Do you have any preferences?

unknwon:
@ss23 Thanks for filling this issue! I think the option one sounds less work to do and fits good under the purpose of "how to search a user".

>Modify GET /users/search to allow searching based upon email, not just username

PR #6474, opened 2021-01-16
API user search now allows searching based on Email

ss23:
There are several tradeoffs with this code, such as the more generic search call with an allowlist of fields. This could be abused in the future if care is not taken, but allows generic code for now.
There is also the decision to hide emails from users that are not logged in, even though it is possible to search based on them.

This change implements #6470.


ss23:
@unknwon bump :)

PR #6478, opened 2021-01-23
webhook: add missing `%s` URL value for webhook description

rtoma:
"Webhooks Guide" URL is empty. "hooks_desc" has a %s which is not filled in code. With this patch it is.

image

### DIFF
--- a/internal/route/repo/webhook.go
+++ b/internal/route/repo/webhook.go
@@ -81,10 +81,10 @@ func Webhooks(c *context.Context, orCtx *orgRepoContext) {
 	var err error
 	var ws []*db.Webhook
 	if orCtx.RepoID > 0 {
-		c.Data["Description"] = c.Tr("repo.settings.hooks_desc")
+		c.Data["Description"] = c.Tr("repo.settings.hooks_desc", "https://gogs.io/docs/features/webhook.html")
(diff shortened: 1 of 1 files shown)


internal/route/repo/webhook.go
```diff
@@ -81,10 +81,10 @@ func Webhooks(c *context.Context, orCtx *orgRepoContext) {
 	var err error
 	var ws []*db.Webhook
 	if orCtx.RepoID > 0 {
-		c.Data["Description"] = c.Tr("repo.settings.hooks_desc")
+		c.Data["Description"] = c.Tr("repo.settings.hooks_desc", "https://gogs.io/docs/features/webhook.html")
 		ws, err = db.GetWebhooksByRepoID(orCtx.RepoID)
 	} else {
-		c.Data["Description"] = c.Tr("org.settings.hooks_desc")
+		c.Data["Description"] = c.Tr("org.settings.hooks_desc", "https://gogs.io/docs/features/webhook.html")
```
xiaobo9, line 87: 'org.settings.hooks_desc' not need the url

unknwon, line 87:
```suggestion
		c.Data["Description"] = c.Tr("org.settings.hooks_desc")
```

unknwon (approved):
Thank you!

Issue #6141, opened 2020-05-01
404 when comparing tags


RobertKosten:
---
name: 404 when comparing tags
about: The `/compare/`endpoint balks when comparing tags instead of branch names (offered through the UI)

---

**Describe the bug**
When viewing a tag the compare button offers to compare the tag to master, e.g. `https://try.gogs.io/RobertKosten/example-project/compare/master...tag`, which leads to a 404. It works fine when comparing branches, but tags seem to always fail (tried comparing two tags, tags with only `[a-z]+`).

**Gogs version and commit**
Gogs Version: 0.11.91.0811
Go1.12.7

**Git version**
1.9.1

**Operating system**
ubuntu 14.04.1

**Database**
mysql 5.5.62-0ubuntu0.14.04.1

**To Reproduce**
1. Go to Repository page
2. Click on 'Branch:master'
3. Click on 'Tags'
4. Click on any {tagname}
5. Click on green compare button to the left on 'Tree: {tagname}'
6. See 404

**Can you reproduce the bug at https://try.gogs.io?**
No, because I cannot clone and there seems to be no way to create a tag through the UI.

**Expected behavior**
A comparison between tag and branch/tag/etc. is displayed, just like between branches.

**Actual behavior**
404

**Screenshots**
Let's no go overboard, no? ;-)

**Additional context**
`[TRACE] Template: status/404`

unknwon:
Thanks for the feedback!

>When viewing a tag the compare button offers to compare the tag to master,

Which button? Can you provide a screenshot?

>No, because I cannot clone and there seems to be no way to create a tag through the UI.

Why can't? You can create "release" which creates a tag or just `git push --tags` from your local repository.

RobertKosten:
> Thanks for the feedback!

Thanks for gogs ;-)

> > When viewing a tag the compare button offers to compare the tag to master,
> 
> Which button? Can you provide a screenshot?

Sure, here it is on my local installation:
![6141](https://user-images.githubusercontent.com/45453466/80806081-f5f66e00-8bb9-11ea-8931-3b843a20a070.png)


> > No, because I cannot clone and there seems to be no way to create a tag through the UI.
> 
> Why can't? You can create "release" which creates a tag or just `git push --tags` from your local repository.

How do I create a release from the UI?

I can't clone to local (I have added my ssh key, of course):
```
$ git clone git@try.gogs.io:RobertKosten/example-project.git
Cloning into 'example-project'...
[...]

PR #6493, opened 2021-02-04
Fixed comparison of tags & branches

rajat315315:
Fixes: #6141 

Now, we can compare tags with tags, and branches with Tags.


unknwon (approved):
Thank you!

PR #6511, opened 2021-02-28
Make __debug_bin work on Windows

linquize:
Make `__debug_bin` work on Windows when using debugger.

This is because `__debug_bin` does not end with `.exe`

### DIFF
--- a/internal/conf/computed.go
+++ b/internal/conf/computed.go
@@ -7,6 +7,7 @@ package conf
 import (
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/conf/computed.go
```diff
@@ -7,6 +7,7 @@ package conf
 import (
```
captain-hoity-toity, line 7: This code is mostly well-written, but I do have a few suggestions to improve it:

1. **Error Handling:** In the `AppPath()` function, you're using `panic` to handle errors. While this does stop the execution, it's generally not recommended for error handling because it can lead to abrupt termination of the program. Instead, consider returning the error and handling it in the caller function.

2. **Code Duplication:** The `homeDirOnce.Do` block in `HomeDir()` function has repetitive code for getting the home directory. You can create an array of environment variables to check and use a loop to iterate over them. This will make your code DRY (Don't Repeat Yourself) and easier to maintain.

3. **Function Documentation:** While the function names are descriptive, it would be beneficial to add comments above the function declarations to describe what they do. This will make the code easier to understand for other developers.

Here's the revised code snippet:

```go
// AppPath returns the absolute path of the application's binary.
func AppPath() (string, error) {
	appPathOnce.Do(func() {
		var err error
		lookPath := true
		if os.PathSeparator == '\\' {
			appPath = strings.ReplaceAll(os.Args[0], "\\", "/")
			file := path.Base(appPath)
			lookPath = file != "__debug_bin"
		}
		if lookPath {
			appPath, err = exec.LookPath(os.Args[0])
			if err != nil {
				return "", fmt.Errorf("look executable path: %w", err)
			}
		}

		appPath, err = filepath.Abs(appPath)
		if err != nil {
			return "", fmt.Errorf("get absolute executable path: %w", err)
		}
	})

	return appPath, nil
}

// HomeDir returns the home directory by reading environment variables.
func HomeDir() string {
	homeDirOnce.Do(func() {
		envs := []string{"HOME", "USERPROFILE", "HOMEDRIVE" + "HOMEPATH"}
		for _, env := range envs {
			homeDir = os.Getenv(env)
			if homeDir != "" {
				break
			}
		}
	})

	return homeDir
}
``

PR #6526, opened 2021-03-19
fix "fatal: git cat-file: could not get object info" when viewing diff of commit deleting submodule

fphammerle:
fixes bug described in https://github.com/gogs/gogs/issues/6525


Tristaan:
+1 I would love this fix to be included.

PR #6553, opened 2021-05-10
docker: check "/data" mount ownership before forcing it

druppy:
Make sure docker start script only force file ownership on `/data` once.

We use NFS as a data drive and have lots of repositories, and our container times our (in k8s) before it is finish running `chown`.

This is a small change, without any practical implication except much faster container start.

See #6554

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -31,8 +31,12 @@ cleanup() {
 }
 
 create_volume_subfolder() {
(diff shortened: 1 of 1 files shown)


GwynethLlewelyn (approved):
Seems perfect to me!

unknwon (approved):
Thank you!

PR #6557, opened 2021-05-14
docker: reduce additional image layer for the not executable gosu binary

PeterDaveHello:
This is a Dockerfile refactor, which will help reduce the image size and image layers. So I don't think it will need an issue opened first, hopefully it's acceptable, please let me know, if it's not, before directly close this PR, I could open an issue for it if it's required! Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,8 @@ COPY . .
 RUN make build TAGS="cert pam"
 
 FROM alpine:3.11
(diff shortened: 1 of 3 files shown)


unknwon (approved):

unknwon:
Shellcheck problems will be fixed by https://github.com/gogs/gogs/pull/6562 as that is unrelated to this PR.

Issue #6565, opened 2021-05-21
Add %p wildcard for Bind Password in LDAP BindDN to allow for self-binding to LDAP


cstepaniak:
**Describe the feature**
Add the ability to bind to LDAP with user's credentials, for situations where neither anonymous bind allowed nor a dedicated LDAP bind user is available.  

**Describe the solution you'd like**
Add the %p wildcard to "Bind Password" in LDAP BindDN (similar to %s for the username) to pass the user-entered password to bind to LDAP.

**Describe alternatives you've considered**
I've not been able to get Gogs to connect to LDAP for our AD domain.  The admins won't allow anonymous bind nor create a dedicated bind user, so I patched Gogs to pass through the user's password for use in binding to LDAP

**Additional context**
diff of internal/auth/ldap/config.go below:

    111c111
    < func (c *Config) findUserDN(l *ldap.Conn, name string) (string, bool) {
    ---
    > func (c *Config) findUserDN(l *ldap.Conn, name string, passwd string) (string, bool) {
    116c116,119
    <               err := l.Bind(bindDN, c.BindPassword)
    ---
    >               bindPW := strings.Replace(c.BindPassword, "%p", passwd, -1)
    >               err := l.Bind(bindDN, bindPW)
    220c223
    <               userDN, found = c.findUserDN(l, name)
    ---
    >               userDN, found = c.findUserDN(l, name, passwd)

PR #6566, opened 2021-05-24
Enable use of %p wildcard in "Bind Password"

cstepaniak:
Enable use of %p wildcard in "Bind Password" field in BindDN to allow user to bind to LDAP with user's credentials, for situations where neither anonymous bind allowed nor a dedicated LDAP bind user is available.


cstepaniak:
This is in regards to Issue #6565

PR #6593, opened 2021-07-27
Example ports change to 9022 and 9080

brandon-foster:
Chrome blocks traffic through port 10080 as of Chrome 91. See the release notes for Chrome 91: https://support.google.com/chrome/a/answer/7679408#91&zippy=%2Cchrome

By using port 9080, for example, instead of port 10080 as the container's exposed HTTP port used in this README, Chrome (as well as Firefox) will not block traffic when the user tries to reach the container's exposed port 9080.

Also suggested is changing the container's port 10022 to port 9022, simply to be consistent with the container's exposed 9080 port.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -18,13 +18,13 @@ $ docker pull gogs/gogs
 $ mkdir -p /var/gogs
 
 # Use `docker run` for the first time.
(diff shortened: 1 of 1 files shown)


unknwon:
Thanks for the PR!

The example Docker port has been changed to 10880 by the merge of https://github.com/gogs/gogs/pull/6597.

Issue #6572, opened 2021-06-05
Chrome now blocks port 10080.


lmpater:
Check this article:
https://www.bleepingcomputer.com/news/security/google-chrome-blocks-port-10080-to-stop-nat-slipstreaming-attacks/

It may make sense to update the documentation that now suggests to map the port 10080 of the host to the 3000 of the container. I just spent about 1hr debugging the issue and, when I was close to give up, I tried to open Gogs on Safari and it worked - so I was able to narrow down the problem to Chrome and found out the information on the port restriction.

Btw, it seems Firefox has been blocking this port for a while now.

djdomi:
Update: all newer Browsers blocks

- ports 
 - 69, 
 - 137, 
 - 161, 
 - 554, 
 - 1719, 
 - 1720, 
 - 1723, 
 - 5060, 
 - 5061, 
 - 6566.

To use them without a Paint use either a other port or a reverse-proxy like nginx

kylegoetz:
FWIW I am still perfectly able to connect to 10080 with FF and Safari as of yesterday.

lmpater:
Hi,

Yeah… The problem seems to be exclusive to Chrome. I solved it (before filing the bug ;-) by simply changing the published port when launching the container.

The reason I created the bug was to share the information. I was concerned that new users may experience this issue and simply give up on Gogs because the example/tutorial suggests a port that doesn’t work for the most common browser. That would be a shame because Gogs rocks - thank you so much btw!

So take the bug as a “documentation bug”, not a software one ;-)

Cheers,
Marcelo

djdomi:
I just wanted to fix that in general and due the fix was easy I wrote a patch ;)

unknwon:
Thanks for the feedback! The doc is updated with https://github.com/gogs/gogs/pull/6597.

PR #6597, opened 2021-07-30
docker: update default port to be exposed on the host

djdomi:
As Described in #6572 Port 10080 has being blocked default by modern Browsers, i suggest changing the default port.


Khairil82:
Selamat [Khairil](khairilisharimysewan@khairilishari.com)

unknwon (approved):
Thank you!

djdomi:
I am glad that you agree to this patch ;)

PR #6628, opened 2021-10-15
use official go:embed instead go-bindate to embed static assets

alimy:
* Use go1.16+ official  go:embed instead go-bindata to embed static files.
* Require go version go1.16+
* Optimize implement of #6451


unknwon:
Hey @alimy, I appreciate the effort, but this is the type of PR that is impossible to review.

If we really want to move forward with the idea, we would first need a plan to be able to iteratively migrate off go-bindata to go:embed. WDYT?

alimy:
yes, we now support  three big version of go like now is  `go-version: [ 1.15.x, 1.16.x, 1.17.x ]`, so maybe start add go:embed support when  go1.18 released.

unknwon:
> yes, we now support three big version of go like now is `go-version: [ 1.15.x, 1.16.x, 1.17.x ]`, so maybe start add go:embed support when go1.18 released.

Yep, my thinking would be once Go 1.18 is out, we could drop 1.15 so all three major versions we support would have go:embed and we don't need to do any crazy magic for backward compatibility.

But again, we need a plan :) that iteratively approaching the goal with small sets of changes.

PR #6630, opened 2021-10-22
support mermaid uml

free1139:
Trying to support mermaid issue https://github.com/gogs/gogs/issues/4229 .

I found that need to ran `make` not `go build` to took effect.

Test the mermaid's diagrams were passed.
~~~
```mermaid
graph TD;
    A-->B;
    A-->C;
    B-->D;
    C-->D;
```
~~~

~~~
```mermaid
sequenceDiagram
    participant Alice
    participant Bob
    Alice->>John: Hello John, how are you?
    loop Healthcheck
        John->>John: Fight against hypochondria
    end
    Note right of John: Rational thoughts <br/>prevail!
    John-->>Alice: Great!
    John->>Bob: How about you?
    Bob-->>John: Jolly good!
```
~~~

The renderings
image


free1139:
> Thanks @free1139!
> 
> Few notes:
> 
> 1. Release branch only get cherry-picked commits for patch releases, so you should target `main` at all time.
> 2. Could you include a copy of pinned version for the JS file into https://github.com/gogs/gogs/tree/main/public/plugins, and link from it. We don't use CDN because this could be ran in a restricted network environment.

Yeah, it's much better after removing CDN. I used it for a while and found that CDN is not very easy to download.

I have made corrections. If you think it is appropriate, I will submit this PR again based on the main branch. At present, there are too many conflicts with the main branch, and I need a stable branch to be used in the production environment.

free1139:
This PR move to 
https://github.com/gogs/gogs/pull/6776

PR #6812, opened 2022-03-06
security: fix SSRF in repository migration

michaellrowley:
https://huntr.dev/bounties/327797d7-ae41-498f-9bff-cc0bf98cf531/ / [CVE-2022-0870](https://nvd.nist.gov/vuln/detail/CVE-2022-0870)

### Describe the pull request

This pull requests contains a proposed patch for the vulnerability noted in the above links.

### Checklist

- [X] I have read and acknowledge the [Contributing guide](https://github.com/gogs/gogs/blob/main/.github/contributing.md).
- [X] I have added test cases to cover the new code.

### DIFF
--- a/internal/form/repo.go
+++ b/internal/form/repo.go
@@ -13,6 +13,7 @@ import (
 	"gopkg.in/macaron.v1"
 
 	"gogs.io/gogs/internal/db"
+	"gogs.io/gogs/internal/netutil"
 )
 
 // _______________________________________    _________.______________________ _______________.___.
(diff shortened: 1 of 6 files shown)


unknwon:
Thanks for the patch! Looks like the CI is failling.

michaellrowley:
I think that the issues have been fixed.

unknwon:
I think this PR is only fixing part of problem after a closer look, which is the webhook... not the repository migration, because the [`isLocalHostname` is only used by webhook](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/gogs/gogs%24+isLocalHostname&patternType=literal) related code logic.

I'm going to make few direct pushes.

unknwon:
@michaellrowley If you wanna look at the steps I made, you can go through each commit.

internal/form/repo.go
```diff
@@ -69,6 +70,11 @@ func (f MigrateRepo) ParseRemoteAddr(user *db.User) (string, error) {
 		if err != nil {
 			return "", db.ErrInvalidCloneAddr{IsURLError: true}
 		}
+
+		if netutil.IsLocalHostname(u.Hostname()) {
+			return "", db.ErrInvalidCloneAddr{IsURLError: true}
+		}
```
unknwon, line 76: @michaellrowley This is where the real fix to the report as far as I understand.

unknwon (approved):

internal/form/repo.go
michaellrowley: Yeah, that seems to be the case (although without the rewritten ``IsLocalHostname`` it would still have been possible to bypass this call and the one in ``validateWebhook``)

PR #6969, opened 2022-05-25
Prevent XSS from _csrf cookie by sanitizing, stripping invalid chars

tsimmons:
### Prevent XSS from _csrf cookie by sanitizing user cookie

To prevent a Cross-Site Scripting (XSS) vulnerability, we should sanitize the user-supplied cookie header. Strip all but a-z, A-Z and 0-9 from the _csrf cookie.

[Link to the issue](https://github.com/gogs/gogs/issues/6953)

### Checklist

- [X] I have read and acknowledge the [Contributing guide](https://github.com/gogs/gogs/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"regexp"
 
 	"github.com/go-macaron/cache"
 	"github.com/go-macaron/csrf"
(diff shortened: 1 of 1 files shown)


unknwon (approved):
Thanks for the PR!

After a closer look, I found a better way of handling it, see https://github.com/gogs/gogs/pull/6969/commits/fc15ac4bfb35304b99bb44ae22f134c18efa1672

After sanitization:

CleanShot 2022-05-26 at 17 52 28@2x

Let me know what you think!

tsimmons:
Looking at your screen shot, it seems the markup.Sanitize is not stripping all brackets (or sanitizing them) ... "<div>" should not be part of the variable name, no angle brackets should break through. In addition, it looks like you would permit other markup that should not be allowed for the _csrf variable looking at NewSanitizer() in internal/markup/sanitizer.go. I'd have to vote no on this proposal for compliance.

tsimmons:
Looking at the source for go-macaron csrf, only alphanumeric characters are allowed for the token, if you can create a new Sanitizer that strips all but a-z, A-Z, 0-9 that would be better. 

[See go-macaron csrf code](https://github.com/go-macaron/csrf/blob/master/csrf.go#L148).

unknwon:
OK, makes sense! Pushed another version in https://github.com/gogs/gogs/pull/6969/commits/bc8545868a4d49b14c80324ab332daa62c1f7d0e

tsimmons:
Looks good!

PR #6970, opened 2022-05-26
repo_editor: prohibits uploading files to `.git.` directory 

1135:
### Describe the pull request

I try to fix it, but it's not necessarily the best solution.
This commit is for informational purposes only. No test cases have been added yet.

### Checklist

- [x] I have added test cases to cover the new code.

### DIFF
--- a/internal/db/repo_editor.go
+++ b/internal/db/repo_editor.go
@@ -460,7 +460,7 @@ type UploadRepoFileOptions struct {
 
 // isRepositoryGitPath returns true if given path is or resides inside ".git" path of the repository.
 func isRepositoryGitPath(path string) bool {
-	return strings.HasSuffix(path, ".git") || strings.Contains(path, ".git"+string(os.PathSeparator) || strings.Contains(path, string(os.PathSeparator)+".git." )
+	return strings.HasSuffix(path, ".git") || strings.Contains(path, ".git"+string(os.PathSeparator) || strings.Contains(path, string(os.PathSeparator)+".git.")
 }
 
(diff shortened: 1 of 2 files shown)


internal/db/repo_editor.go
```diff
@@ -460,7 +460,7 @@ type UploadRepoFileOptions struct {
 
 // isRepositoryGitPath returns true if given path is or resides inside ".git" path of the repository.
 func isRepositoryGitPath(path string) bool {
-	return strings.HasSuffix(path, ".git") || strings.Contains(path, ".git"+string(os.PathSeparator))
+	return strings.HasSuffix(path, ".git") || strings.Contains(path, ".git"+string(os.PathSeparator) || strings.Contains(path, string(os.PathSeparator)+".git.")
```
unknwon, line 463:
```suggestion
	return strings.HasSuffix(path, ".git") || strings.Contains(path, ".git"+string(os.PathSeparator) || strings.Contains(path, ".git."+string(os.PathSeparator))
```

Shouldn't the `string(os.PathSeparator)` being concatenated in this way?

1135: I recommend this modification for the following reasons.

```suggestion
	return strings.HasSuffix(path, ".git") || strings.HasSuffix(path, ".git.") || strings.Contains(path, ".git"+string(os.PathSeparator) || strings.Contains(path, ".git."+string(os.PathSeparator)
```

Whether it contains a os separator or not, the value of `tree_path` will be treated as a folder when `path.Join`.

ref: /gogs/internal/db/repo_editor.go `targetPath := path.Join(dirPath, upload.Name)`

unknwon (approved):

PR #6003, opened 2020-03-22
cmd: backup support exclude mirror repositories

fishbullet:
Fixes #6000

Abilitty for backup cli command exclude mirror repositories.

PR #7060, opened 2022-06-15
docker: support adding extra options to backup command

liuxhit:
### Describe the pull request

PR-[#6003](https://codecov.io/gh/gogs/gogs/pull/6003?src=pr&el=desc) added support for exclude mirror repositories when backup, but we do not have a convinent way for docker use. 

This PR:

1. add aupport for set custom backup args via docker environment; 
```
  environment:
      - TZ=Asia/Shanghai
      - RUN_CROND=true
      - BACKUP_INTERVAL=3d
      - BACKUP_RETENTION=7d
      - BACKUP_EXTRA_ARGS=--verbose --exclude-mirror-repos
````
2.fix bug that failing to backup when there is multipul parameters, otherwise we will see errors in the log:
```
2022/06/15 10:31:00 [ INFO] Packing backup files to: /backup --verbose --exclude-mirror-repos/gogs-backup-20220615103100.zip
2022/06/15 10:31:00 [FATAL] [...s.io/gogs/internal/cmd/backup.go:90 runBackup()] Failed to create backup archive '/backup --verbose --exclude-mirror-repos/gogs-backup-20220615103100.zip': open /backup --verbose --exclude-mirror-repos/gogs-backup-20220615103100.zip: no such file or directory
Error: Backup job returned non-successful code.
```

Link to the issue: n/a

### Checklist

### CODE REVIEW

**Review by unknwon — APPROVED** [review]
Thank you!

File: docker/runtime/backup-job.sh [hunk 1]
```diff
@@ -16,7 +17,11 @@ execute_backup_job() {
 		BACKUP_ARGS="${BACKUP_ARGS} --exclude-repos=${BACKUP_ARG_EXCLUDE_REPOS}"
 	fi
 
-	./gogs backup "${BACKUP_ARGS}" || echo "Error: Backup job returned non-successful code." && exit 1
+	if [ -n "${BACKUP_EXTRA_ARGS}" ]; then
+		BACKUP_ARGS="${BACKUP_ARGS} ${BACKUP_EXTRA_ARGS}"
+	fi
+
+	./gogs backup ${BACKUP_ARGS} || echo "Error: Backup job returned non-successful code." && exit 1
```
unknwon, line 24:
```suggestion
	./gogs backup "${BACKUP_ARGS}" || echo "Error: Backup job returned non-successful code." && exit 1
```

Is removing the double quota here on purpose (i.e. fix the bug you mentioned)? Because shell lint is complaining about it.

liuxhit: Yes, thanks for reminding me and sorry for the late reply. 
I removed the double quotes here for bug-fix. It broken the shell lint rule, but we acturlly need to pass multi arguments in string to the backup command(rather than pass the hole string as one argument).

liuxhit: typo: "the hole string" -> "the whole string"

unknwon, line 24:
```suggestion
	# NOTE: We actually need word splitting to be able to pass multiple arguments.
	# shellcheck disable=SC2086
	./gogs backup ${BACKUP_ARGS} || echo "Error: Backup job returned non-successful code." && exit 1
```

unknwon:
Thank you!

PR #7102, opened 2022-07-07
chore: add Simona Iacob as translator for Romanian language

ecx0d:
Added Simona Iacob as Translator for Romanian Language


unknwon (approved):
Thank you!

PR #7192, opened 2022-10-17
ui: remove useless "Cancel" buttons

yardenshoham:
Port of go-gitea/gitea#21381

### Describe the pull request

There's no need for a cancel button, just leave the page.

### Checklist

- [X] I have read and acknowledged the [Contributing guide](https://github.com/gogs/gogs/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/templates/org/create.tmpl
+++ b/templates/org/create.tmpl
@@ -20,7 +20,6 @@
 						<button class="ui green button">
 							{{.i18n.Tr "org.create_org"}}
 						</button>
-						<a class="ui button" href="{{AppSubURL}}/">{{.i18n.Tr "cancel"}}</a>
 					</div>
 				</div>
 			</form>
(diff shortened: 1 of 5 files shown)


unknwon:
Thanks for the PR, could you explain why it is "useless"?

If "cancel" buttons are not worth it (not yet convinced), we should remove all of them (at least three including this one, "New Repository", "New Migration" and "New Organization"), not just this single one.

Aside from that, I like the "make it better" approach more that is described in https://github.com/go-gitea/gitea/issues/21131, i.e. takes user back to the previous page.

yardenshoham:
It's useless because there is no reason to use it. If I want to go to the home page I'll just go back

unknwon:
Can you also remove other "cancel" buttons from "New Migration" and "New Organization" pages?

yardenshoham:
> Can you also remove other "cancel" buttons from "New Migration" and "New Organization" pages?

Done

unknwon (approved):
Thanks for the followup!

PR #7198, opened 2022-10-22
auth: enable authentication by token from password

### LINKED ISSUE #3866: Use personal access tokens as password

ly4096x:
### Describe the pull request

As #6421 is not getting update, I created this PR to close #3866 . This PR tries to authenticate using password as token at first, together with verifying username, and if that fails, it will try to fallback to try authenticate using username as token.

### Checklist

- [x] I have added test cases to cover the new code. ~~**Having trouble locating test cases, please help point out where to add tests.**~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to Gogs are documented in this file.
 
 ### Added
 
+- Support using personal access token in the password field. [#3866](https://github.com/gogs/gogs/issues/3866)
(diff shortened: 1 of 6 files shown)


unknwon (approved):
Thanks for the PR!

I pushed a revised version to make this PR actually works :)

PR #7291, opened 2022-12-24
conf: add new config option for default branch name

### LINKED ISSUE #6296: Recognize `main` as default branch if `master` is not available
~~New repository should start using `main` as the default branch, fall back to `master` when `main` is not available when the default branch is not set.~~

When detecting the default branch, try `main` if `master` is not available.

vamc19:
This PR adds a new configuration parameter `DEFAULT_BRANCH` which can be used to set the name of the default branch globally instead of changing the it for every repository. It defaults to "master" in order to remain backwards compatible. For new installs, it can be configured on the install page.

When creating a new repo, the bare repository will be initialized with the global default branch name. The commands on
the bare repo page will display the correct branch name.

Relates to/Fixes #6296 

### Checklist


unknwon (approved):
Thank you!

PR #7296, opened 2022-12-31
English, German, French and Romanian corrected and completed

ecxod:
nothing special, but some translations. English, German, French and Romanian corrected and completed


soloturn:
i stopped to review the translations. too much work :)

PR #7710, opened 2024-03-31
docker: allow running Gogs with a different user

feed3r:
At the moment the user running Gogs is git, although this is valid for the vast majority of the cases, there are situations where a different user needs to be configured.

For example: we need that a repository in Gogs is synchronized with an existing git repository in which user git is already used and allow only the commits coming from Gogs repository. In this case we need a specific user name different from the standard one.

PLEASE NOTE:

It's very important to notice that the user written in the app.ini configuration file (created during the gogs installation) must be consistent with the user set in the GOGS_USER environment variable.
Actually, this is completely left to the user, who must be aware to set the very same user.

As a future fix, it would be good to prefill the user input box in the installation screen with the value of the passed GOGS_USER.

Link to the issue: "n/a"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,7 @@ RUN apk --no-cache --no-progress add \
   tzdata \
   rsync
 
+ARG USER_GOGS
 ENV GOGS_CUSTOM /data/gogs
 
 # Configure LibC Name Service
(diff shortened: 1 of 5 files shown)


unknwon:
Hi @feed3r, thanks for your effort putting together this PR ❤️ Unfortunately we cannot accept this PR after our internal security review as this change would open up too many risky points that might be exploited for arbitrary command execution via the user name, i.e. user name, home directory and the UID/GID are now become non-deterministic.

Given this is special case for your special need, we recommend you keep and build from your Dockerfile instead.

PR #7807, opened 2024-09-12
install: fix SMTP password config name to save

mobilemindtec:
fix SMTP password config name from PASSWD to PASSWORD on install.

## Describe the pull request

After installation, I found that sending emails was not working. Then I noticed that the email password was not being loaded because the initial installation configured the password in a field called PASSWD but apparently this field was renamed to PASSWORD. I tested using the new name and the email was sent successfully.


unknwon (approved):
Thank you!

unknwon:
The 0.13.1 has been released that includes the patch of the reported issue.

PR #7981, opened 2025-06-08
web_editor: prohibit CRUD to symbolic files

unknwon:
Fixes [GHSA-wj44-9vcg-wjq7](https://github.com/gogs/gogs/security/advisories/GHSA-wj44-9vcg-wjq7)


unknwon:
The 0.13.3 has been released that includes this patch.

PR #8078, opened 2026-01-05
security: prevent symlink traversal in repository API

ravisastryk:
## Summary

This PR fixes **CVE-2025-8110**, a critical symlink bypass vulnerability (CVSS 8.7) that allows authenticated users to achieve Remote Code Execution(RCE) by exploiting the `PutContents` API.

**The Vulnerability:**
The API validates file paths for directory traversal (`../`) but fails to validate symlink destinations. An attacker can:
1. Create a repository containing a symlink pointing to `/.git/config`
2. Use `PUT /repos/:owner/:repo/contents/:path` to write to the symlink
3. The system follows the symlink and overwrites `.git/config`
4. By injecting `sshCommand`, the attacker achieves RCE on next git operation

**The Fix:**
Adds symlink-aware path validation at all file write entry points:
- New `internal/pathutil` package with `ValidateRepoPath()` function
- Resolves symlinks using `filepath.EvalSymlinks()` and validates destinations
- Blocks writes if resolved path escapes repository directory
- Handles chained symlinks and relative symlink escapes

**References:**
- https://www.wiz.io/blog/wiz-research-gogs-cve-2025-8110-rce-exploit
- 700+ instances compromised in the wild as of Dec 2025

Link to the GHSA: https://github.com/gogs/gogs/security/advisories/GHSA-gg64-xxr9-qhjp

## Checklist

## Test plan

**1. Run unit tests:**
```bash
go test ./internal/pathutil/... -v
```
Expected: All tests pass including symlink traversal detection tests.

**2. Verify symlink bypass is blocked (requires local Gogs instance):**
```bash
# Setup: Create a test repository
mkdir /tmp/test-repo && cd /tmp/test-repo
git init

# Create malicious symlink pointing outside repo
ln -s /etc/passwd evil_link
git add evil_link && git commit -m "add symlink"

# Attempt to write via API (replace with your Gogs URL and token)
curl -X PUT "http://localhost:3000/api/v1/repos/testuser/test-repo/contents/evil_link" \
  -H "Authorization: token YOUR_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"message":"test","content":"dGVzdA=="}'
```

**Before fix:** Returns 200, file outside repo is overwritten  
**After fix:** Returns 400 with `"message": "Invalid path: symlink traversal detected"`

**3. Verify legitimate symlinks still work:**
```bash
# Create symlink pointing within repo
ln -s README.md docs_link
git add docs_link && git commit -m "add valid symlink"

# Write via API
curl -X PUT "http://localhost:3000/api/v1/repos/testuser/test-repo/contents/docs_link" \
  -H "Authorization: token YOUR_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"message":"update","content":"dGVzdA=="}'
```

**Expected:** Returns 200, symlink target within repo is updated successfully.

**4. Verify web editor is also protected:**
- Navigate to repository file editor
- Attempt to edit a file that is a symlink pointing outside the repository
- Expected: Error message "Invalid path: symlink traversal detected"


internal/database/repo_editor.go
```diff
@@ -119,23 +119,32 @@ type UpdateRepoFileOptions struct {
 
 // UpdateRepoFile adds or updates a file in repository.
 func (r *Repository) UpdateRepoFile(doer *User, opts UpdateRepoFileOptions) (err error) {
-	// 🚨 SECURITY: Prevent uploading files into the ".git" directory.
-	if isRepositoryGitPath(opts.NewTreeName) {
-		return errors.Errorf("bad tree path %q", opts.NewTreeName)
-	}
-
+	// CVE-2025-8110: Validate treePath for symlink traversal
```
unknwon: I think this code comment is out of place and can actually be removed.

internal/database/repo_editor.go
```diff
@@ -119,23 +119,32 @@ type UpdateRepoFileOptions struct {
...
+	if err := r.DiscardLocalRepoBranchChanges(opts.OldBranch); err != nil {
+		return fmt.Errorf("discard local repo branch[%s] changes: %v", opts.OldBranch, err)
 	} else if err = r.UpdateLocalCopyBranch(opts.OldBranch); err != nil {
 		return fmt.Errorf("update local copy branch[%s]: %v", opts.OldBranch, err)
 	}
 
 	repoPath := r.RepoPath()
 	localPath := r.LocalCopyPath()
 
+	// CVE-2025-8110: Validate path before file operations
```
unknwon: Instead of mentioning CVE-2025-8110, which is pretty abstract in this case, it should use the code comment pattern // 🚨 SECURITY: ... where it states what exactly it is protected against. e.g.,

```go
// 🚨 SECURITY: Ensure the path resolves within the repository directory by following through symlink(s) (if any).
```


Same for other mentions of CVE-2025-8110 in this PR.

internal/database/repo_editor.go
```diff
@@ -119,23 +119,32 @@ type UpdateRepoFileOptions struct {
...
+		return fmt.Errorf("security: %w", err)
+	}
+	if err := pathutil.ValidateRepoPath(repoPath, opts.NewTreeName); err != nil {
+		return fmt.Errorf("security: %w", err)
+	}
+
+	// 🚨 SECURITY: Prevent uploading files into the ".git" directory.
+	if isRepositoryGitPath(opts.NewTreeName) {
+		return errors.Errorf("bad tree path %q", opts.NewTreeName)
+	}
```
unknwon: 1. Validation of the input options should happen as the first thing at the top of the function body.
2. The composition of error messages can be improved as `errors.Wrap(err, "validate old tree path")`, etc.
3. `pathutil` aims to be "path"-specific package, which has no knowledge about what a "repo" is, if we need a place for such a helper, we can put it into `repoutil`, and name the helper like `repoutil.ValidatePathWithin`.

internal/pathutil/symlink.go
unknwon, line 1: This file can be named as simply as `symlink`.

internal/pathutil/symlink_security.go
```diff
@@ -0,0 +1,226 @@
+// Copyright 2025 The Gogs Authors. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+// Security fix for CVE-2025-8110: Symlink bypass leading to RCE
```
unknwon:
```suggestion
```

internal/pathutil/symlink_security.go
```diff
@@ -0,0 +1,226 @@
...
+package pathutil
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// ErrSymlinkTraversal is returned when a symlink points outside the allowed directory
```
unknwon:
```suggestion
// ErrSymlinkTraversal is returned when the final path resolves outside the allowed directory by following through symlink(s).
```

internal/route/api/v1/repo/contents.go
```diff
@@ -181,6 +181,22 @@ type PutContentsRequest struct {
 
 // PUT /repos/:username/:reponame/contents/*
 func PutContents(c *context.APIContext, r PutContentsRequest) {
+	// CVE-2025-8110: Validate path for symlink traversal before any file operations
+	treePath := c.Repo.TreePath
+	repoPath := repoutil.RepositoryPath(c.Params(":username"), c.Params(":reponame"))
+
+	if err := pathutil.ValidateRepoPath(repoPath, treePath); err != nil {
+		if pathutil.IsErrSymlinkTraversal(err) {
+			c.JSON(400, map[string]string{
```
unknwon:
```suggestion
			c.JSON(http.StatusBadRequest, map[string]string{
```

internal/route/api/v1/repo/contents.go
```diff
@@ -210,7 +226,7 @@ func PutContents(c *context.APIContext, r PutContentsRequest) {
 		return
 	}
 
-	repoPath := repoutil.RepositoryPath(c.Params(":username"), c.Params(":reponame"))
+	repoPath = repoutil.RepositoryPath(c.Params(":username"), c.Params(":reponame"))
```
unknwon: Why do we need to grab the `repoPath` again that we already did in line 186? i.e., this line should be deleted instead.

internal/route/api/v1/repo/contents.go
```diff
@@ -192,7 +208,7 @@ func PutContents(c *context.APIContext, r PutContentsRequest) {
 	}
 
 	// 🚨 SECURITY: Prevent path traversal.
-	treePath := pathutil.Clean(c.Params("*"))
+	treePath = pathutil.Clean(c.Params("*"))
```
unknwon: Shouldn't we clean the path on the line 185 directly?

unknwon (changes requested):
Thanks for the PR! Submitting first round of feedback. Let me know if you got any questions.

internal/database/repo_editor.go
ravisastryk: Thank you @unknwon

internal/database/repo_editor.go
ravisastryk: That's a great suggestion so that we can maintain in one `repoutil` package

ravisastryk:
Thank you @unknwon for the feedback. I have applied the suggested changes. When you get a chance, please take another look, resolve comments so that we can close out this critical symlink bypass issue.

internal/database/repo_editor.go
```diff
@@ -119,6 +120,15 @@ type UpdateRepoFileOptions struct {
 
 // UpdateRepoFile adds or updates a file in repository.
 func (r *Repository) UpdateRepoFile(doer *User, opts UpdateRepoFileOptions) (err error) {
```
unknwon: Since you're touching this method, would you mind taking out the named return value, it looks like we don't use it for special capture. And fix any compile errors.
```suggestion
func (r *Repository) UpdateRepoFile(doer *User, opts UpdateRepoFileOptions) error {
```

internal/repoutil/repoutil.go
```diff
@@ -73,3 +74,13 @@ func RepositoryLocalPath(repoID int64) string {
 func RepositoryLocalWikiPath(repoID int64) string {
 	return filepath.Join(conf.Server.AppDataPath, "tmp", "local-wiki", strconv.FormatInt(repoID, 10))
 }
+
+// ValidatePathWithin validates a repository tree path against the repository
+// root path. It ensures the tree path is cleaned and validated to be within
+// the repository directory, following through any symlinks.
+//
+// 🚨 SECURITY: Ensure the path resolves within the repository directory by following through symlink(s) (if any).
```
unknwon:
```suggestion
//
// 🚨 SECURITY: This function MUST be called to validate any user-input tree path.
```

internal/route/api/v1/repo/contents_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"gogs.io/gogs/internal/conf"
+	"gogs.io/gogs/internal/pathutil"
+	"gogs.io/gogs/internal/repoutil"
+)
+
+// TestValidateRepoPathSymlink ensures that repository path validation
+// detects symlink traversal. This validation is used by PutContents.
+func TestValidateRepoPathSymlink(t *testing.T) {
```
unknwon: For what this test is doing, it should be placed alongside the where the `repoutil.ValidatePathWithin` lives. If we want to be extra sure that the `PutContents` is calling the correct security helper, we should mock it as if it was an actual HTTP request to the `PUT /repos/:username/:reponame/contents/*` and expects a 400 response code.

internal/route/api/v1/repo/contents_test.go
```diff
@@ -0,0 +1,53 @@
...
+	if os.Getenv("GO_WANT_DEBUGGER") == "1" {
+		t.Skip("skipping under debugger")
+	}
```
unknwon: This seems rather unusual to have in this codebase, we should remove it unless there is a very good reason.

internal/route/repo/editor.go
```diff
@@ -430,12 +442,13 @@ func UploadFile(c *context.Context) {
 }
 
 func UploadFilePost(c *context.Context, f form.UploadRepoFile) {
+	// UploadFilePost handles uploading files to repository via editor UI.
+	// The function performs validation and then delegates to `UploadRepoFiles`.
```
unknwon: This block of docstring is out of place? i.e., should be moved before line 444.

internal/route/repo/editor.go
```diff
@@ -444,10 +457,8 @@ func UploadFilePost(c *context.Context, f form.UploadRepoFile) {
 	f.TreePath = pathutil.Clean(f.TreePath)
 	treeNames, treePaths := getParentTreeFields(f.TreePath)
 	if len(treeNames) == 0 {
-		// We must at least have one element for user to input.
```
unknwon, line 447: Why do we delete this comment?

internal/route/repo/editor.go
```diff
@@ -457,46 +468,38 @@ func UploadFilePost(c *context.Context, f form.UploadRepoFile) {
 	c.Data["commit_choice"] = f.CommitChoice
 	c.Data["new_branch_name"] = branchName
 
+	// Validate upload path and branch early.
```
unknwon:
```suggestion
// 🚨 SECURITY: Ensure the path resolves within the repository directory by following through symlink(s) (if any).
```

internal/route/repo/editor.go
```diff
@@ -457,46 +468,38 @@ func UploadFilePost(c *context.Context, f form.UploadRepoFile) {
...
-		// User can only upload files to a directory.
-		if !entry.IsTree() {
-			c.FormErr("TreePath")
-			c.RenderWithErr(c.Tr("repo.editor.directory_is_a_file", part), tmplEditorUpload, &f)
-			return
-		}
+	if err := ensureUploadTargetIsDirectory(c, treeNames); err != nil {
+		c.FormErr("TreePath")
+		c.RenderWithErr(err.Error(), tmplEditorUpload, &f)
+		return
```
unknwon, line 496: I don't feel like the refactoring here is strictly necessary for this PR. For security patches, we should aim minimal possible changes because backporting changes to release branch could also issues with refactoring.

internal/pathutil/symlink.go
```diff
@@ -0,0 +1,215 @@
...
+func IsSymlink(path string) (bool, error) {
+	info, err := os.Lstat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return info.Mode()&os.ModeSymlink != 0, nil
+}
```
unknwon: An identical helper `osutil.IsSymlink` already exists.

internal/database/repo_editor.go
ravisastryk: Thank you, this is a good go practice too. Removed named return value as it's not used for special capture in this function and made adjustments with declaring `err` locally with no side-effects.

internal/route/api/v1/repo/contents_test.go
ravisastryk: Moved unit test to repoutil package where ValidatePathWithin lives.

internal/route/api/v1/repo/contents_test.go
ravisastryk: Yes, agree. It is not a standard pattern in this codebase. Removed.

internal/route/repo/editor.go
ravisastryk: Moved comments.

internal/route/repo/editor.go
ravisastryk: Yes, it was inadvertent. Put back the comments.

internal/route/repo/editor.go
ravisastryk: Updated to use standard security comment pattern.

internal/route/repo/editor.go
ravisastryk: Initially tried to revert the changes with minimal refactoring, but complained about cyclomatic complexity errors. For better code maintenance I refactored further. I would appreciate it if you could take another look and make any updates directly.

internal/pathutil/symlink.go
ravisastryk: Good to know. Removed duplicate `IsSymlink` function.

ravisastryk:
Thank you @unknwon for taking the time to re-review. Please feel free to make any minor updates directly on this branch so that we can expedite it. I appreciate your time once again.

unknwon:
Thanks again for the followup, I am going to checkout locally to make and test the change, expect some direct pushes.

unknwon:
Hey @rbukovansky, after trialing multiple approaches, I ended up going with the simplest route that rejects any repo updates related to symlink in https://github.com/gogs/gogs/pull/8082. Dealing with all the troubles with symlink do not feel worth it at this point.

PR #8090, opened 2026-01-17
Check URI Scheme validity according to RFC3986

### LINKED ISSUE #6732: Add XMPP URI support (RFC5122)
### Describe the feature

Dear @gogs team,

In first, I wish you a Happy New Year 2022!

Can you see for add XMPP URI support?
XMPP URI support must be added because it does not work.

Examples:
- xmpp:username@domain.tld
- xmpp:username@domain.tld?message
- xmpp:mucroom@domain.tld?join

It is a standard:
- RFC5122: https://tools.ietf.org/html/rfc5122

Can you update it and solves this big bug?

-----------------------------

1/ In file like README

2/ In issue/pr/ticket/comment

3/ In user/organization and repository description too

-------------------------------

Thanks in advance.

### Describe the solution you'd like

That it works:
- xmpp:username@domain.tld
- xmpp:username@domain.tld?message
- xmpp:mucroom@domain.tld?join

### Describe alternatives you've considered

None

### Additional context

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

GwynethLlewelyn:
## Issue description

XMPP URI Schemes were defined and standardised by [RFC5122](https://datatracker.ietf.org/doc/html/rfc5122#section-3), duly registered with [IANA](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml#:~:text=%5BRFC3529%5D-,xmpp,-Extensible%20Messaging%20and) since 2008 (with permanent status), and fully conforming with the generic URI Scheme as defined by [RFC3986, Section 3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3:~:text=scheme%20%20%20%20%20%20%3D%20ALPHA%20*%28%20ALPHA%20/%20DIGIT%20/%20%22%2B%22%20/%20%22%2D%22%20/%20%22.%22%20%29).

However, [the current code base](https://github.com/gogs/gogs/blob/dfe8fbde8449bc709e5d0bf6657fd356ebbf1fd9/internal/markup/markdown.go#L38) only checks for a subset of possible valid URIs using the regular expression:

```regex
^[a-z][\w-]+://|^mailto:
```

This _ad hoc_ implementation unfortunately does not conform with the URI Schemes specification.

Most notably, it explicitly excludes all URIs without an explicit authority element (i.e., what follows `//`). But such schemes are perfectly valid; in fact, the above regular expression requires an additional element to be checked, namely `mailto:`, which, by convention, skips the authority element, like many others, including XMPP.

Furthermore, by using `\w`, the regex as implemented allows not only mixing lowercase and uppercase letters (URI Schemes are not case sensitive, although all RFCs **strongly recommend** that canonicalised URIs should be lowercase **only**), but also allows the underscore character `_`, which is, however, **not allowed** by the RFCs specifying the URI Schemes. This means that (imaginary) URI Schemes as `fake_scheme://this.blows.up.everything` are accepted with the regular expression in the code, although such schemes are invalid according to the RFCs; and, on the other hand, perfectly valid URI schemes such as `tel:+1-555-1234` or `xmpp:my.name@example.com`  are _rejected_ (both of which, incidentally, have been fully registered with IANA for a long time).

As such, this simple PR proposes to correct the regular expression so that it conforms to the actual rules specified by the RFCs, accepting all IANA-registered schemes, while at the same time guaranteeing that invalid schemes are properly rejected.

This PR closes #6732.

(Special thanks to @Neustradamus for first reporting this issue exactly four years ago and having patiently waited for a solution since then!)

## Checklist

- [ ] I have added an entry to [CHANGELOG](https://github.com/gogs/gogs/blob/main/CHANGELOG.md). (if applicable)

## Proposed solution and test plan

The [ABNF](https://datatracker.ietf.org/doc/html/rfc5234 "Augmented Backus-Naur Form") specifying all valid URI Schemes is: 

```abnf
scheme        = alpha *( alpha | digit | "+" | "-" | "." )
```

References:  
- [RFC2396, Section 3.1](
https://datatracker.ietf.org/doc/html/rfc2396#section-3:~:text=well%20as%20%22http%22%29.-,scheme%20%20%20%20%20%20%20%20%3D%20alpha%20*%28%20alpha%20%7C%20digit%20%7C%20%22%2B%22%20%7C%20%22%2D%22%20%7C%20%22.%22%20%29,-Relative%20URI%20references)
- [RFC3986, Section 3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3:~:text=scheme%20%20%20%20%20%20%3D%20ALPHA%20*%28%20ALPHA%20/%20DIGIT%20/%20%22%2B%22%20/%20%22%2D%22%20/%20%22.%22%20%29) (supersedes RFC2396)

This corresponds to the regular expression[^1]:

```regex
^[a-z][a-z\-\+.]+:(?:\/\/)? 
```

which was tested on Regex101: <https://regex101.com/r/HEpzIM/1>  
and the Go Playground: <https://go.dev/play/p/vtYEugsNAfo>

No further changes beyond changing the existing regular expression in `markdown.go` are necessary to fully validate URI Schemes according to RFC3986.

[^1]: Note that, although the specifications allow for a mix of lowercase and uppercase US-ASCII letters, it also strongly recommends that URIs are canonicalised in lowercase, which is assumed here to be the case.


unknwon:
Hey @GwynethLlewelyn, thanks for the PR!

Could you:
1. Add test cases to sanity check the regex is actually matching all of them. And some counter-cases as well, make sure it doesn't over matching obviously non-URL strings.
2. I think this PR worth as CHANGELOG entry.

unknwon (changes requested):
See above comment.

Issue #8085, opened 2026-01-09
Adding repository which exists in repo ROOT but not in DB removes its files


gerab:
### Gogs version

0.13.3

### Git version

- Server: 2.47.3
- Client:  2.52.0.windows.1

### Operating system

Debian 13

### Database

SQLite 3

### Describe the bug

I'm recovering my instance after a server failure with a slightly outdated DB file. Some repositories that physically exist on the disk are not shown in Gogs, which is expected. I've just tried to add (recover) a "new" repository, which already exists by adding a new repo with the same name that it had before the failure. 

Achieved result:
- Gogs alerts about the path already exists and wipes the repository files completely, and they cannot be recovered anymore

### To reproduce

1. Put the git repo files in your repository ROOT/{user} catalog
2. Try to add this repo using UI
3. Observe the files were removed

### Expected behavior

- Gogs alerts about the path already exists and does nothing more
OR
- Gogs merges existing repo files with the "new" repo (less expected for security reasons)

### Additional context

UI log message:
An error has occurred : initRepository: initRepository: path already exists: /mnt/Repositories/{user}/{repoName}.git

Test log message:
[ERROR] [...gogs/internal/route/repo/repo.go:100 handleCreateError()] CreatePost: initRepository: initRepository: path already exists: /mnt/Repositories/{user}/{repoName}.git

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #8091, opened 2026-01-18
repo: prevent data loss by checking directory existence before creati…

maishivamhoo123:
## Describe the pull request
This PR fixes a data loss vulnerability where Gogs would delete an existing directory on the filesystem if a user tried to create a repository with the same name.
#8085 

By moving the path existence check to the beginning of `CreateRepository`, the process exits before the initialization phase. This prevents the execution of `RemoveAllWithNotice` during a rollback, ensuring existing data remains untouched.

Link to the issue: closes https://github.com/gogs/gogs/issues/8085
gogs

## Checklist
- [x] I have added an entry to [CHANGELOG](https://github.com/gogs/gogs/blob/main/CHANGELOG.md).

## Test plan
1. **Manual Test:** - Created a folder `secret-project.git` with a file `secret.txt` inside the repository root via terminal.
   - Attempted to create a repo named `secret-project` in the UI.
   - **Result:** Received "directory already exists" error; `secret.txt` survived.
2. **Automated Test:** - Added and ran `Test_CreateRepository_PreventDeletion` in `internal/database/repo_test.go`.
   - **Result:** `PASS`

### DIFF
--- a/internal/database/repo_test.go
+++ b/internal/database/repo_test.go
@@ -58,17 +58,15 @@ func TestRepository_ComposeMetas(t *testing.T) {
 }
 
 func Test_CreateRepository_PreventDeletion(t *testing.T) {
-	owner := &User{Name: "testuser", IsAdmin: true}
+	owner := &User{Name: "testuser"}
 	opts := CreateRepoOptionsLegacy{Name: "safety-test"}
 	repoPath := RepoPath(owner.Name, opts.Name)
-
-	assert.NoError(t, os.MkdirAll(repoPath, os.ModePerm))
+	require.NoError(t, os.MkdirAll(repoPath, os.ModePerm))
 
 	canary := filepath.Join(repoPath, "canary.txt")
-	assert.NoError(t, os.WriteFile(canary, []byte("should survive"), 0644))
+	require.NoError(t, os.WriteFile(canary, []byte("should survive"), 0644))
 
 	_, err := CreateRepository(owner, owner, opts)
-
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "repository directory already exists")
 	assert.True(t, osutil.IsExist(canary))


internal/database/repo.go
```diff
@@ -1207,7 +1203,10 @@ func CreateRepository(doer, owner *User, opts CreateRepoOptionsLegacy) (_ *Repos
 	if !owner.canCreateRepo() {
 		return nil, ErrReachLimitOfRepo{Limit: owner.maxNumRepos()}
 	}
-
+	repoPath := RepoPath(owner.Name, opts.Name)
+	if osutil.IsExist(repoPath) {
+		return nil, fmt.Errorf("repository directory already exists: %s", repoPath)
```
unknwon: Since we are touching this line, let's use `github.com/pkg/errors` package:
```suggestion
		return nil, errors.Errorf("repository directory already exists: %s", repoPath)
```

internal/database/repo.go
```diff
@@ -1207,7 +1203,10 @@ func CreateRepository(doer, owner *User, opts CreateRepoOptionsLegacy) (_ *Repos
...
+	}
```
unknwon, line 1206:
```suggestion
	}

```

internal/database/repo_test.go
```diff
@@ -52,3 +55,28 @@ func TestRepository_ComposeMetas(t *testing.T) {
 		assert.Equal(t, "https://someurl.com/{user}/{repo}/{issue}", metas["format"])
 	})
 }
+func Test_CreateRepository_PreventDeletion(t *testing.T) {
```
unknwon, line 60:
```suggestion
}

func Test_CreateRepository_PreventDeletion(t *testing.T) {
```

internal/database/repo_test.go
```diff
@@ -52,3 +55,28 @@ func TestRepository_ComposeMetas(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	_, err = CreateRepository(owner, owner, opts)
+	if err == nil {
+		t.Fatal("Expected error when directory exists, but got nil")
+	}
+
+	if !osutil.IsExist(canary) {
+		t.Error("CRITICAL: The existing directory was deleted during CreateRepository failure!")
+	}
```
unknwon: 1. The existing code comments provide no value because they're literally describing what's obvious about what the code is doing, i.e., should be removed.
2. Why are we using two different test utilities? i.e., we do not use `t.Fatal` and `t.Error`, etc.
3. `os.WriteFile` exists, should be used for creation of `canary`.
4. We should check for the specific error returned by `CreateRepository`, not just any random error that we think the test would pass.

maishivamhoo123:
"Hi @unknwon Sir ,
 I have applied the suggested changes (using pkg/errors and cleaning up the test formatting).

I discovered that the tests were failing in CI because the mock user was hitting the repository creation limit before reaching the path existence check. I've updated the test case to use an Admin user to bypass those initial constraints. The tests should pass now. Could you please take another look?

unknwon:
@maishivamhoo123 the test is failing because you're mocking a user entry without setting its `MaxRepoCreation`, if you set it to `-1`, should mean unlimited.

maishivamhoo123:
@unknwon can please review it? all the test cases are running properly and made all the requested changes.

internal/database/repo.go
unknwon: @maishivamhoo123 this comment is resolved without applying the suggestion diff?

internal/database/repo_test.go
```diff
@@ -52,3 +55,20 @@ func TestRepository_ComposeMetas(t *testing.T) {
...
+	repoPath := RepoPath(owner.Name, opts.Name)
+
+	assert.NoError(t, os.MkdirAll(repoPath, os.ModePerm))
+
+	canary := filepath.Join(repoPath, "canary.txt")
+	assert.NoError(t, os.WriteFile(canary, []byte("should survive"), 0644))
+
+	_, err := CreateRepository(owner, owner, opts)
+
+	assert.Error(t, err)
```
unknwon: This line should use require, assert won't stop code and it will then panic if err is nil

PR #8164, opened 2026-02-08
email: replace gomail with go-mail

unknwon:
Replace `gopkg.in/gomail.v2` with `github.com/wneessen/go-mail` for both message composition and SMTP sending.

go-mail handles RFC 5322 compliance (header folding, encoding), MIME structure, quoted-printable, STARTTLS, implicit TLS, and auth mechanism auto-discovery — eliminating the custom message builder and SMTP sender code entirely.

Key changes:
- Replace gomail.Message with go-mail's Msg type for message composition.
- Replace the hand-rolled smtpSender (LoginAuth, STARTTLS, SMTPS, client certs) with go-mail's Client.
- Remove the `[email] DISABLE_HELO` configuration option (HELO/EHLO is now always sent).
- Remove `gopkg.in/gomail.v2` dependency.

Closes https://github.com/gogs/gogs/issues/6307

### DIFF
--- a/internal/email/email.go
+++ b/internal/email/email.go
@@ -74,7 +74,7 @@ func render(tpl string, data map[string]any) (string, error) {
 func SendTestMail(email string) error {
 	msg, err := newMessage([]string{email}, "Gogs Test Email", "Hello 👋, greeting from Gogs!")
 	if err != nil {
-		return err
+		return errors.Wrap(err, "new message")
 	}
 	return sendMessage(msg)
 }
@@ -116,7 +116,7 @@ func SendUserMail(_ *macaron.Context, u User, tpl, code, subject, info string) e
 
 	msg, err := newMessage([]string{u.Email()}, subject, body)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "new message")
 	}
 	msg.info = fmt.Sprintf("UID: %d, %s", u.ID(), info)
 
@@ -146,7 +146,7 @@ func SendActivateEmailMail(c *macaron.Context, u User, email string) error {
 
 	msg, err := newMessage([]string{email}, c.Tr("mail.activate_email"), body)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "new message")
 	}
 	msg.info = fmt.Sprintf("UID: %d, activate email", u.ID())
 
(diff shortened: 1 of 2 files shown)


internal/email/message.go
```diff
@@ -2,214 +2,133 @@ package email
...
-	*gomail.Message
+type message struct {
+	info        string
+	msg         *gomail.Msg
 	confirmChan chan struct{}
 }
 
-// NewMessageFrom creates new mail message object with custom From header.
-func NewMessageFrom(to []string, from, subject, htmlBody string) *Message {
+func newMessageFrom(to []string, from, subject, htmlBody string) *message {
```
unknwon: this function should wrap and return errors only, `internal/email` is a helper package, only the call site of this package should determine how to handle the error.

internal/email/email.go
```diff
@@ -200,10 +220,13 @@ func composeIssueMessage(issue Issue, repo Repository, doer User, tplName string
 	if err != nil {
 		log.Error("HTMLString (%s): %v", tplName, err)
```
unknwon: why is this still printing not return error?

internal/email/email.go
unknwon, line 1: all functions here should always return error let call sites to print.

internal/email/email.go
```diff
@@ -107,26 +111,28 @@ func SendUserMail(_ *macaron.Context, u User, tpl, code, subject, info string) {
...
-		log.Error("render: %v", err)
-		return
+		return errors.Wrap(err, "render")
 	}
 
-	msg := NewMessage([]string{u.Email()}, subject, body)
-	msg.Info = fmt.Sprintf("UID: %d, %s", u.ID(), info)
+	msg, err := newMessage([]string{u.Email()}, subject, body)
+	if err != nil {
+		return err
```
unknwon: where is the error wrapping? don't be lazyass. same for other places in this package.

Neustradamus:
Dear @unknwon,

After several years of battle, the fix has been done about the unsecure gomail package.
This PR fixes it.

Can you reopen my advisory and validate it?
- https://github.com/gogs/gogs/security/advisories/GHSA-9xhf-v677-vp99

Thanks in advance.

Linked to:
- https://github.com/gogs/gogs/issues/7917
- https://github.com/gogs/gogs/issues/6307

PR #8166, opened 2026-02-08
lfs: verify content hash and prevent object overwrite

unknwon:
## Summary

- Prevent cross-repository LFS object overwrite by using `O_CREATE|O_EXCL` to atomically refuse writing to an existing file path. If the file already exists (same OID uploaded by another repo), the existing size is returned so the caller can still create the DB record linking the new repo to that OID.
- Compute SHA-256 of uploaded content during write and reject the upload if the hash does not match the claimed OID, preventing an attacker from writing arbitrary content to a known OID path.

## Test plan

- [x] Existing `TestLocalStorage_Upload` updated with matching OID/content pair
- [x] New test: OID mismatch rejects upload and cleans up file
- [x] New test: duplicate upload returns existing size and preserves original content
- [x] `task lint` passes
- [x] `task test` passes (lfsutil)

References: https://github.com/gogs/gogs/security/advisories/GHSA-cj4v-437j-jq4c

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/lfsutil/storage.go
+++ b/internal/lfsutil/storage.go
@@ -107,7 +107,7 @@ func (s *LocalStorage) Upload(oid OID, rc io.ReadCloser) (int64, error) {
 		return 0, ErrOIDMismatch
 	}
 
-	if err := os.Rename(tmpPath, fpath); err != nil {
+	if err := os.Rename(tmpPath, fpath); err != nil && !os.IsExist(err) {
 		return 0, errors.Wrap(err, "publish object file")
 	}
 	return written, nil


internal/lfsutil/storage.go
```diff
@@ -10,7 +12,10 @@ import (
 	"gogs.io/gogs/internal/osutil"
 )
 
-var ErrObjectNotExist = errors.New("Object does not exist")
+var (
+	ErrObjectNotExist = errors.New("Object does not exist")
```
unknwon:
```suggestion
	ErrObjectNotExist = errors.New("object does not exist")
```

internal/lfsutil/storage.go
```diff
@@ -58,29 +63,47 @@ func (s *LocalStorage) Upload(oid OID, rc io.ReadCloser) (int64, error) {
...
+	// existing file's size.
+	if fi, err := os.Stat(fpath); err == nil {
+		_, _ = io.Copy(io.Discard, rc)
+		return fi.Size(), nil
+	}
+
+	// Write to a temp file and verify the content hash before publishing.
+	// This ensures the final path always contains a complete, hash-verified
+	// file, even when concurrent uploads of the same OID race.
+	tmp, err := os.CreateTemp(dir, ".lfs-upload-*")
```
unknwon: put tmp files in a different dir alongside the actual dir, mixing them within the same dir is messy.

internal/lfsutil/storage_test.go
```diff
@@ -47,49 +50,53 @@ func TestLocalStorage_storagePath(t *testing.T) {
 
 func TestLocalStorage_Upload(t *testing.T) {
 	s := &LocalStorage{
-		Root: filepath.Join(os.TempDir(), "lfs-objects"),
+		Root: t.TempDir(),
```
unknwon: where is the subdir ????

internal/lfsutil/storage_test.go
```diff
@@ -47,49 +50,53 @@ func TestLocalStorage_storagePath(t *testing.T) {
...
 	}
-	t.Cleanup(func() {
-		_ = os.RemoveAll(s.Root)
+
+	// SHA-256 of "Hello world!" is c0535e...
+	const helloWorldOID = OID("c0535e4be2b79ffd93291305436bf889314e4a3faec05ecffcbb7df31ad9e51a")
```
unknwon:
```suggestion
	const helloWorldOID = OID("c0535e4be2b79ffd93291305436bf889314e4a3faec05ecffcbb7df31ad9e51a") // "Hello world!"
```

internal/lfsutil/storage_test.go
```diff
@@ -47,49 +50,53 @@ func TestLocalStorage_storagePath(t *testing.T) {
...
-	}
-	for _, test := range tests {
-		t.Run(test.name, func(t *testing.T) {
-			written, err := s.Upload(test.oid, io.NopCloser(strings.NewReader(test.content)))
-			assert.Equal(t, test.expWritten, written)
-			assert.Equal(t, test.expErr, err)
-		})
-	}
+	t.Run("oid mismatch", func(t *testing.T) {
+		// Use a valid OID format that doesn't match the content.
```
unknwon:
```suggestion
	t.Run("valid OIC but wrong content", func(t *testing.T) {
```

internal/lfsutil/storage_test.go
```diff
@@ -47,49 +50,53 @@ func TestLocalStorage_storagePath(t *testing.T) {
...
-		Root: filepath.Join(os.TempDir(), "lfs-objects"),
+		Root: t.TempDir(),
 	}
-	t.Cleanup(func() {
-		_ = os.RemoveAll(s.Root)
+
+	// SHA-256 of "Hello world!" is c0535e...
+	const helloWorldOID = OID("c0535e4be2b79ffd93291305436bf889314e4a3faec05ecffcbb7df31ad9e51a")
+
+	t.Run("invalid oid", func(t *testing.T) {
```
unknwon:
```suggestion
	t.Run("invalid OID", func(t *testing.T) {
```

internal/lfsutil/storage_test.go
```diff
@@ -47,49 +50,53 @@ func TestLocalStorage_storagePath(t *testing.T) {
...
-		content    string
-		expWritten int64
-		expErr     error
-	}{
-		{
-			name:   "invalid oid",
-			oid:    "bad_oid",
-			expErr: ErrInvalidOID,
-		},
+	t.Run("valid oid", func(t *testing.T) {
```
unknwon:
```suggestion
	t.Run("valid OID", func(t *testing.T) {
```

internal/lfsutil/storage_test.go
```diff
@@ -47,49 +50,53 @@ func TestLocalStorage_storagePath(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.Equal(t, "Hello world!", buf.String())
+	})
 }
 
 func TestLocalStorage_Download(t *testing.T) {
 	oid := OID("ef797c8118f02dfb649607dd5d3f8c7623048c9c063d532cc95c5ed7a898a64f")
 	s := &LocalStorage{
-		Root: filepath.Join(os.TempDir(), "lfs-objects"),
+		Root: t.TempDir(),
```
unknwon: Ditto

internal/lfsutil/storage.go
```diff
@@ -58,29 +63,51 @@ func (s *LocalStorage) Upload(oid OID, rc io.ReadCloser) (int64, error) {
...
+	// existing file's size.
+	if fi, err := os.Stat(fpath); err == nil {
+		_, _ = io.Copy(io.Discard, rc)
+		return fi.Size(), nil
+	}
+
+	// Write to a temp file and verify the content hash before publishing.
+	// This ensures the final path always contains a complete, hash-verified
+	// file, even when concurrent uploads of the same OID race.
+	tmpDir := filepath.Join(s.Root, ".tmp")
```
unknwon: This is very lazyass way ok? Backup would include all partial garbage.

We already have pattern about the use of temporary uploads for repo files, i.e., https://github.com/gogs/gogs/blob/400ae7bd28eb80c824022d0d81929f5d8304ecad/conf/app.ini#L134C1-L135

So, add a new config option under `lfs`, similar to repo uploads and use that instead.

conf/app.ini
```diff
@@ -279,6 +279,8 @@ ACCESS_CONTROL_ALLOW_ORIGIN =
 STORAGE = local
 ; The root path to store LFS objects on local file system.
 OBJECTS_PATH = data/lfs-objects
+; The path to temporarily store LFS objects during upload verification.
+TEMP_PATH = data/tmp/lfs
```
unknwon:
```suggestion
OBJECTS_TEMP_PATH = data/tmp/lfs-objects
```

internal/lfsutil/storage.go
```diff
@@ -58,29 +65,50 @@ func (s *LocalStorage) Upload(oid OID, rc io.ReadCloser) (int64, error) {
...
+		return fi.Size(), nil
+	}
+
+	// Write to a temp file and verify the content hash before publishing.
+	// This ensures the final path always contains a complete, hash-verified
+	// file, even when concurrent uploads of the same OID race.
+	if err := os.MkdirAll(s.TempDir, os.ModePerm); err != nil {
+		return 0, errors.Wrap(err, "create temp directory")
+	}
+	tmp, err := os.CreateTemp(s.TempDir, ".lfs-upload-*")
```
unknwon:
```suggestion
	tmp, err := os.CreateTemp(s.TempDir, "upload-*")
```

internal/lfsutil/storage.go
```diff
@@ -58,29 +65,50 @@ func (s *LocalStorage) Upload(oid OID, rc io.ReadCloser) (int64, error) {
...
-		return 0, errors.Wrap(err, "copy file")
+		return 0, errors.Wrap(err, "write object file")
+	}
+
+	if computed := hex.EncodeToString(hash.Sum(nil)); computed != string(oid) {
+		return 0, ErrOIDMismatch
+	}
+
+	if err := os.Rename(tmpPath, fpath); err != nil {
+		return 0, errors.Wrap(err, "publish object file")
```
unknwon: Is there a specific error type we can assert so if the error is about file already exist, we don't need to fail the client request.

PR #8260, opened 2026-05-14
docs: document MariaDB support

robertsilen:
## Describe the pull request

Document MariaDB (same as MySQL: TYPE = mysql). README + installation + local dev docs. Docs only.
≥ 10.3: matches scripts/mysql.sql (legacy InnoDB branch only below 10.3).


unknwon (approved):
Thanks :)

PR #8261, opened 2026-05-16
Show merge commit label in commit view

### LINKED ISSUE #2696: Show if a commit is a merge in the commit view

gshaowei6:
Closes #2696

The commits list already marks merge commits based on `ParentsCount > 1`, but the individual commit page only showed the parent SHAs. This applies the same signal on the commit page so merge commits are easier to spot there too.

Changes:
- set `IsMergeCommit` when the viewed commit has more than one parent
- render a compact `Merge commit` label next to the commit metadata
- cover the parent-count helper with a focused test

IssueHunt: https://issuehunt.io/repos/16752620/issues/2696

Checks:
- `git diff --check`
- DeepSource on this PR: Docker, Go, and Shell passed

### DIFF
--- a/conf/locale/locale_en-US.ini
+++ b/conf/locale/locale_en-US.ini
@@ -879,6 +879,7 @@ settings.description_desc = Description of repository. Maximum 512 characters le
 settings.description_length = Available characters
 
 diff.browse_source = Browse Source
+diff.merge_commit = Merge commit
 diff.parent = parent
 diff.commit = commit
 diff.data_not_available = Diff Data Not Available.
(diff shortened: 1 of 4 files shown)


gshaowei6:
Closing this after a local rebase audit against current `upstream/main`.

I attempted to rebase the branch and found that the conflict is not just nearby context drift: the old single-commit `Diff` / `RawDiff` handler that this PR changed is no longer present or routed in current `main`. Reapplying the old patch would reintroduce stale handler code and would not be a safe, reviewable update.

Given the current code shape, this should not be force-rebased as-is. If the merge-commit label is still useful, it needs to be re-evaluated against the current route/template flow and submitted as a fresh focused change with tests for the active path.

PR #8265, opened 2026-05-18
security: rate-limit two-factor verification attempts

unknwon:
## Summary

Cap failed TOTP and recovery code submissions to a small budget per user to prevent brute-forcing the second factor when primary credentials are already known. The counter is keyed by user ID in the existing cache and shared across both verification handlers, so alternating endpoints does not bypass the limit. The counter is cleared on successful verification so legitimate users are not penalized for prior typos.

Addresses [GHSA-5mgh-chvj-hqf5](https://github.com/gogs/gogs/security/advisories/GHSA-5mgh-chvj-hqf5).

## Test plan

- [x] `task lint` passes.
- [x] `go test ./internal/userx/...` passes.
- [ ] Manually verify lockout after the configured threshold of failed attempts.
- [ ] Manually verify the counter resets on successful TOTP / recovery code submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to Gogs are documented in this file.
 
 ## 0.15.0+dev (`main`)
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


unknwon:
Reopening from a branch in gogs/gogs directly.

PR #8267, opened 2026-05-18
fix: store remember-me state in server-side session

unknwon:
## Summary

Fixes [GHSA-4pph-25p3-pw73](https://github.com/gogs/gogs/security/advisories/GHSA-4pph-25p3-pw73) (CVE-2026-24886).

The legacy "remember me" cookie was derived from `user.rands + user.passwd` — both columns of the `user` table — so an attacker with a database dump could forge a valid cookie for any user and trigger `AutoLogin()` to mint a session without knowing the password.

This PR moves remember-me state into the server-side session. The session ID is already an unguessable, server-issued random value, so there is nothing on the client that can be derived from leaked database contents.

## Approach

- On login with remember-me checked, set `rememberMe = true` in the session and overwrite the session cookie with `Max-Age = LoginRememberDays * 86400`.
- On each authenticated request, if `rememberMe` is set, re-emit the cookie to slide its expiration forward.
- Delete `AutoLogin()` and the `gogs_incredible` / `gogs_awesome` cookie set/read paths.
- Remove the now-unused `[security] COOKIE_REMEMBER_NAME` and `[security] COOKIE_USERNAME` config options.

Existing forged or legitimate `gogs_incredible` cookies in users' browsers become inert — they are no longer read by the server.

## Test plan

- [x] `task lint` — clean
- [x] `task build` — succeeds
- [x] `go test ./...` — passes (pre-existing `TestSSHParsePublicKey/dsa-1024` failure is unrelated to this change and depends on the local `ssh-keygen` build)
- [ ] Manual: log in with "remember me" checked, confirm session cookie has long Max-Age and survives browser restart
- [ ] Manual: log in without "remember me", confirm session cookie is a session cookie (no Max-Age)
- [ ] Manual: sign out clears the session cookie


unknwon:
Reopening from a branch in gogs/gogs directly.

PR #8384, opened 2026-07-12
web: fix frontend assets behind reverse-proxy subpaths

### LINKED ISSUE #8385: Frontend assets fail when Gogs is served from a reverse-proxy subpath
### Gogs version

latest-commit-build (2a0092707145)

### Git version

- Server: Not relevant
- Client: Not relevant

### Operating system

Not relevant

### Database

Not relevant

### Describe the bug

Hi! I found this while testing the current `main` branch for a production deployment under `/gogs`.

Gogs does not load its production frontend correctly when the reverse proxy strips the configured subpath.

The HTML entrypoints use `/gogs/assets/...`, but the proxy forwards those requests to Gogs as `/assets/...`. The production static handler expects the `/gogs` prefix, so the request may fall through to the SPA handler and return HTML instead of JavaScript or CSS.

Some nested bundle assets also use root-relative `/assets/...` URLs. These bypass the `/gogs/*` proxy route and return a 404.

I linked draft PR for the proposed fix 

### To reproduce

1. Build Gogs from commit `2a0092707145` with the production frontend.
2. Configure Gogs with:

   ```ini
   [server]
   EXTERNAL_URL = https://example.com/gogs/
   ```
3. Configure Caddy to strip the external prefix:
   ```
   redir /gogs /gogs/
   handle_path /gogs/* {
       reverse_proxy 127.0.0.1:3000
   }
   ```
4. Start Gogs and open https://example.com/gogs/.
5. Inspect the browser network panel.
   Requests below /gogs/assets/ may return HTML instead of the expected asset. Nested requests to /assets/ return 404 because they are outside the /gogs/* route.

### Expected behavior

All frontend assets should load through `/gogs/assets/` externally and resolve through `/assets/` after the proxy strips the prefix.

JavaScript, CSS, fonts, and worker files should return successful responses with their expected content types.

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

duytbp:
## Describe the pull request

Fixes #8385.

This change makes production frontend assets work behind a prefix-stripping reverse proxy and updates the Caddy subpath example.

## Checklist

- [x] I have added test cases to cover the new code and provided the test plan below.
- [ ] I have added an entry to [CHANGELOG](https://github.com/gogs/gogs/blob/main/CHANGELOG.md).

## Test plan

1. Serve Gogs under `/gogs` using the Caddy configuration documented in this PR.
2. Confirm JavaScript, CSS, fonts, and workers load successfully from `/gogs/assets/`.
3. Run `moon run gogs:test`.
4. Run `go test -tags=prod ./cmd/gogs/internal/web`.
5. Run `moon run web:lint` and `moon run web:build`.


cmd/gogs/internal/web/web.go
```diff
@@ -863,8 +863,11 @@ func renderIndex(index []byte, wc context.WebContext) ([]byte, error) {
 		"{{.WebContext}}", script,
 	}
 	if wc.SubURL != "" {
-		// Prefix Vite's absolute root paths with the subpath for non-root mounts.
+		// Prefix entrypoint paths with the subpath for non-root mounts. Other
+		// bundled assets stay relative to the entrypoint that references them.
 		pairs = append(pairs,
+			`src="./assets/`, `src="`+wc.SubURL+`/assets/`,
+			`href="./assets/`, `href="`+wc.SubURL+`/assets/`,
```
unknwon, line 870: Hmm, why do we need these change?

docs/fine-tuning/reverse-proxy.mdx
unknwon: You're using Caddy 2 syntax, so the heading should be updated as well

cmd/gogs/internal/web/webapp_prod_test.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		contents, err := public.WebAssets.ReadFile(path.Join("dist/assets", entry.Name()))
+		require.NoError(t, err)
+		require.NotContains(t, string(contents), "url(/assets/", entry.Name())
+		require.NotContains(t, string(contents), "/assets/worker-", entry.Name())
+	}
+}
+
+func TestRenderIndex_PrefixesRelativeEntrypoints(t *testing.T) {
```
unknwon: this test does not belong here, `renderIndex` is not in `webapp_prod.go`.

Then the rest of the file doesn't seem to have value to keep, not sure what they are testing for and what values those tests provide.

unknwon (changes requested):
Thanks for the PR!

However it seems too many related changes bundled together. If you wanna land this PR, I think only the changes in `cmd/gogs/internal/web/webapp_prod.go` is needed?

docs/fine-tuning/reverse-proxy.mdx
duytbp: Updated the heading to `Caddy 2`.

cmd/gogs/internal/web/webapp_prod_test.go
duytbp: I moved the `renderIndex` test to `web_test.go`, next to the code it covers.

The two remaining production tests depend on the embedded bundle. One checks that `/assets/...` returns the JavaScript file instead of falling through to the SPA HTML handler. The other checks that the generated CSS and JavaScript do not contain root-relative font or worker URLs.

cmd/gogs/internal/web/web.go
duytbp: Setting `base` to `./` in `vite.config.ts` makes the entrypoints in `index.html` relative:

```html
<script src="./assets/app.js">
```

This works at `/gogs/`, but on a direct visit to `/gogs/user/sign-in`, the browser resolves it as `/gogs/user/assets/app.js` instead of `/gogs/assets/app.js`.

This change rewrites the entrypoints to the stable, subpath-aware URL.

duytbp (commented):
Thanks for taking a look. I updated the Caddy heading and reorganized the tests.

I tested limiting the fix to `webapp_prod.go`, but that only fixes the backend side. The complete flow has three parts:

- `vite.config.ts` makes URLs inside the generated CSS and JavaScript relative. Without this, fonts and workers still request `/assets/...` and bypass `/gogs/*`.
- `web.go` gives the main JavaScript and CSS entrypoints a stable `/gogs/assets/...` URL. Without this, opening a deeper page directly can resolve them under the wrong path.
- `webapp_prod.go` serves those requests after the proxy strips `/gogs`, leaving Gogs with `/assets/...`.

Each change handles a different stage of the same asset request, so removing any one leaves part of the frontend broken.

unknwon:
> * `vite.config.ts` makes URLs inside the generated CSS and JavaScript relative. Without this, fonts and workers still request `/assets/...` and bypass `/gogs/*`.

I am not sure I follow, aren't `/assets/` already replaced with subpath right below your new additions?

cmd/gogs/internal/web/webapp_prod_test.go
unknwon, line 1: What's the value of keeping these tests?

duytbp:
Those replacements only run against `index.html`, so they cover its `<script>` and `<link>` tags. They do not rewrite URLs inside the generated CSS or JavaScript.

Without `base: "./"`, font and worker URLs inside those bundles still use `/assets/...`, which bypasses `/gogs/`.

cmd/gogs/internal/web/webapp_prod_test.go
duytbp: One test checks that assets are still served after the proxy strips `/gogs`. The other checks that fonts and workers do not go back to using `/assets/...`.

I added them to cover both parts of the fix, but I’m happy to remove them if you think they’re unnecessary.

unknwon:
> Those replacements only run against `index.html`, so they cover its `<script>` and `<link>` tags. They do not rewrite URLs inside the generated CSS or JavaScript.
> 
> Without `base: "./"`, font and worker URLs inside those bundles still use `/assets/...`, which bypasses `/gogs/`.

Ok that makes sense. Now if we use `./assets/`, the lines to replace `/assets/` can be removed?

cmd/gogs/internal/web/webapp_prod_test.go
unknwon: I'm incline to remove them because it looks like they're merely testing what the they mocked up.

duytbp:
Yes, I just verified that the build only uses `./assets/...`, so those replacements can be removed.

cmd/gogs/internal/web/webapp_prod_test.go
duytbp: Removed

unknwon (approved):
Thank you!

PR #8387, opened 2026-07-13
tool: avoid panic in BasicAuthDecode on a value with no colon

arpitjain099:
## Describe the pull request

`BasicAuthDecode` splits the decoded credentials with `strings.SplitN(s, ":", 2)` and then reads `auth[1]`. When the decoded value has no colon, `SplitN` returns a one-element slice and that access panics with an index-out-of-range.

It is reachable from any `Authorization: Basic <base64>` header whose payload decodes to a string with no colon (for example `base64("foobar")`), on both the web/API auth path (`internal/context/auth.go`) and the git-over-HTTP handler, since both discard the returned error. The panic is recovered per request so it becomes a logged 500 rather than a crash, but it is still an easy way for an unauthenticated request to spam errors. The fix returns an empty password when there is no colon, matching what the sibling `internal/authx.DecodeBasic` already does.

Link to the issue: n/a

## Checklist

- [ ] I have added an entry to [CHANGELOG](https://github.com/gogs/gogs/blob/main/CHANGELOG.md). (if applicable)

## Test plan

Run `go test ./internal/tool/`. The added `TestBasicAuthDecode` covers the no-colon input (which panics without the guard and passes with it) alongside the normal and colon-in-password cases.

To reproduce the original panic manually, send a request carrying `Authorization: Basic Zm9vYmFy` (that is `base64("foobar")`, which has no colon) to any Basic-auth endpoint.


unknwon:
Thanks for the PR, please update the description using PR template and I can re-open.

arpitjain099:
Thanks @unknwon, I have updated the description to follow the PR template. Could you re-open when you get a chance? Happy to add a CHANGELOG entry too if you would like one for this.

internal/tool/tool.go
```diff
@@ -55,6 +55,11 @@ func BasicAuthDecode(encoded string) (string, string, error) {
 	}
 
 	auth := strings.SplitN(string(s), ":", 2)
+	// A value with no colon splits into a single element, so guard the second
+	// index to avoid an out-of-range panic on malformed input.
```
unknwon:
```suggestion
```

unknwon (approved):

unknwon:
Thank you!

PR #8400, opened 2026-07-30
markup: support alerts in Markdown render

Nahid-NHB:
## Describe the pull request

Block quotes that start with `[!NOTE]`, `[!TIP]`, `[!IMPORTANT]`, `[!WARNING]`, or `[!CAUTION]` on their own line now render as callouts with a colored border, an icon, and a title, the way GitHub renders them. Everything else keeps rendering as a plain block quote.

Link to the issue: https://github.com/gogs/gogs/issues/8388

### Root cause

Markdown goes through blackfriday v1, which has no GFM alerts extension, and `MarkdownRenderer` never overrode `BlockQuote`. So `> [!NOTE]` fell through to the stock renderer and came out as a normal block quote with the marker still sitting there as text.

### What changed

`MarkdownRenderer.BlockQuote` checks whether a paragraph of the block quote starts with an alert marker on its own line, and emits an alert box if it does. Anything that isn't an alert still goes through the stock renderer, so plain block quotes are byte for byte what they were before.

One wrinkle worth knowing about: blackfriday merges block quotes that are only separated by a blank line into a single one, so the five alerts in the issue arrive as one block quote with five paragraphs. The renderer therefore looks at every paragraph instead of just the first, and content that follows a marker (lists, code blocks, more paragraphs) stays inside that alert.

The marker has to take up the whole first line, same as GitHub, so `> [!NOTE] text on the same line` stays a plain quote. Matching is case insensitive and works with `ENABLE_HARD_LINE_BREAK` either on or off.

bluemonday strips `class` by default, so the sanitizer needed an allowlist for the alert classes. The patterns only accept the five alert types and the five octicons used here, nothing broader.

Styling lives in `_markdown.less` with the matching rules added to `gogs.min.css`. Colors are GitHub's, all at least 4.5:1 against white. Icons come from the octicons font that head.tmpl already loads, and they carry `aria-hidden` since the title text is what conveys the meaning. Octicons 4.3.0 has no `report` icon, so `[!IMPORTANT]` uses `megaphone`.

### Notes for reviewers

A plain `>` quote written right after an alert with only a blank line between them ends up inside that alert. blackfriday has already merged the two into one block quote by the time the renderer runs, so there is nothing left to tell them apart. Same root cause as the alert merging above, and it only really goes away with a different Markdown library.

Also, I see the contributing guide asks for a proposal in Discussions before new features. Happy to move this over there if you'd rather discuss the approach first, or to close it if the direction isn't what you want.

## Checklist

- [x] I have added an entry to [CHANGELOG](https://github.com/gogs/gogs/blob/main/CHANGELOG.md). (if applicable)

## Test plan

Automated:

```
go test -race -count=1 ./internal/markup/...
```

`Test_MarkdownAlert` covers all five types, a lowercase marker, consecutive alerts, block content inside an alert, a marker with no body, a plain quote followed by an alert, a marker that isn't on its own line, an unknown type, and a plain block quote. `Test_Sanitizer` gained cases for the accepted and the rejected classes. Both fail on `main` and pass with this change.

Manual:

1. Create a repository and put this in its `README.md`:

   ```
   > [!NOTE]
   > General information or additional context.

   > [!TIP]
   > A helpful suggestion or best practice.

   > [!IMPORTANT]
   > Key information that shouldn't be missed.

   > [!WARNING]
   > Critical information that highlights a potential risk.

   > [!CAUTION]
   > Information about potential issues that require caution.
   ```

2. Open the repository home page. Each block renders as a callout: blue Note, green Tip, purple Important, amber Warning, red Caution, each with its octicon and a colored left border.
3. Add `> A regular quote.` somewhere separated by other content and confirm it still renders as a gray block quote.


tats-u:
https://github.com/orgs/community/discussions/204814

FYI, only GitHub has chosen to make Caution's color red instead of yellow. (All other similar decisions have been made under the influence of GitHub.)

PR #8413, opened 2026-09-04
context: drop invalid charset parameter from JSON Content-Type

HsukqiLee:
## Describe the pull request

JSON responses currently use `application/json; charset=UTF-8` as the `Content-Type` because the macaron render middleware always appends a charset parameter. The `charset` parameter is not defined for the `application/json` media type, so including one makes the media type invalid ([RFC 8259](https://www.rfc-editor.org/info/rfc8259/#section-8.1), [IANA Media Types registry](https://www.iana.org/assignments/media-types/application/json)).

This change overrides the JSON rendering in the request context to set the `Content-Type` to exactly `application/json`, keeping the existing indentation behavior (indented outside production, compact in production).

Link to the issue: https://github.com/gogs/gogs/issues/8409

## Checklist

- [x] I have added an entry to [CHANGELOG](https://github.com/gogs/gogs/blob/main/CHANGELOG.md). (if applicable)

## Test plan

1. `go test ./internal/context/...` — new `TestContext_JSON` covers the `Content-Type`, status code, JSON body, and the environment-dependent indentation.
2. Run a local instance and create an access token:

```bash
curl --location-trusted \
    --fail \
    --header "Authorization: Basic $(printf '%s' <username>:<password> | base64)" \
    --header 'Content-Type: application/json' \
    --data-binary '{"name": "ExampleToken"}' \
    --write-out '%{content_type}\n' \
    --output /dev/null \
    http://localhost:3000/api/v1/users/<username>/tokens
```

The printed content type is `application/json` instead of `application/json; charset=UTF-8`.


unknwon:
Thanks for the PR, but I don't think it's a useful change, see https://github.com/gogs/gogs/issues/8409#issuecomment-5552877266

